PR #30, opened 2020-06-15
CI Build improvements

marcperez:
- Adds a unique build ID
- Allows downloading artifacts after publishing workflow

### DIFF
--- a/src/core/core.ts
+++ b/src/core/core.ts
@@ -19,7 +19,8 @@ class Core {
     public static readonly version = {
         version: process.env.VERSION,
         revision: process.env.COMMIT_HASH,
-        branch: process.env.COMMIT_BRANCH
+        branch: process.env.COMMIT_BRANCH,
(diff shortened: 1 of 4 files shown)


src/core/core.ts
```diff
@@ -19,7 +19,8 @@ class Core {
     public static readonly version = {
         version: process.env.VERSION,
         revision: process.env.COMMIT_HASH,
-        branch: process.env.COMMIT_BRANCH
+        branch: process.env.COMMIT_BRANCH,
+        buildId: process.env.ADYEN_BUILD_ID
```
pabloai, line 23: Shall we also add the name of the library here so it can be easily found?

marcperez: Good you mentioned @pabloai! That's been included in the `buildId` (e.g. `@adyen/adyen-web-f59434c6-1af3-...`)

PR #32, opened 2020-06-15
MBWay shopper data - expect it to be inputted the same way it is outputted

sponglord:
**Description**
The shopper data that can be sent into the MBWay component should have the same fields as the data that is returned by the component i.e. `shopperEmail` & `telephoneNumber`.
Currently we are expecting the _internal_ names: `email` & `phoneNumber` from the data object used to configure the component

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/src/components/MBWay/MBWay.tsx
+++ b/src/components/MBWay/MBWay.tsx
@@ -10,8 +10,8 @@ export class MBWayElement extends UIElement {
 
     formatProps(props: UIElementProps): UIElementProps {
         if (props.data) {
-            props.data.email = props.data.shopperEmail || '';
-            props.data.phoneNumber = props.data.telephoneNumber || '';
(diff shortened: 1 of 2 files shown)


src/components/MBWay/MBWay.tsx
```diff
@@ -9,6 +9,10 @@ export class MBWayElement extends UIElement {
     private static type = 'mbway';
 
     formatProps(props: UIElementProps): UIElementProps {
+        if (props.data) {
+            props.data.email = props.data.shopperEmail || props.data.email;
+            props.data.phoneNumber = props.data.telephoneNumber || props.data.telephoneNumber;
```
marcperez, line 14:
```suggestion
            props.data.phoneNumber = props.data.telephoneNumber || props.data.phoneNumber;
```

Was this the intended change?

PR #102, opened 2020-07-10
Improve types

marcperez:
## Summary
- Add more property types to Core options
- Extract Card types

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -1,6 +1,6 @@
 import { CustomTranslations, Locales } from '../language/types';
 import { PaymentAmount, PaymentMethods, PaymentMethodOptions } from '../types';
-import { AnalyticsOptions } from './Analytics/Analytics';
+import { AnalyticsOptions } from './Analytics/types';
 import { PaymentMethodsResponseObject } from './ProcessResponse/PaymentMethodsResponse/types';
 import { RiskModuleOptions } from './RiskModule/RiskModule';
 
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -1,36 +1,71 @@
 import { CustomTranslations, Locales } from '../language/types';
+import { PaymentAmount, PaymentMethods, PaymentMethodOptions } from '../types';
+import { AnalyticsOptions } from './Analytics/Analytics';
+import { PaymentMethodsResponseObject } from './ProcessResponse/PaymentMethodsResponse/types';
 
 export interface CoreOptions {
     /**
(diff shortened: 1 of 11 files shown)


sponglord:
Is the pattern we're going for a separate `types` file? So should the `Analytics.js` also use a `types.js`?

src/core/types.ts
```diff
@@ -1,36 +1,74 @@
...
+import { AnalyticsOptions } from './Analytics/Analytics';
+import { PaymentMethodsResponseObject } from './ProcessResponse/PaymentMethodsResponse/types';
+import { RiskModuleOptions } from './RiskModule/RiskModule';
 
 export interface CoreOptions {
     /**
-     * Use test. When you're ready to accept live payments, change the value to one of our {@link https://docs.adyen.com/checkout/drop-in-web#testing-your-integration }live environments}.
+     * Use test. When you're ready to accept live payments, change the value to one of our {@link https://docs.adyen.com/checkout/drop-in-web#testing-your-integration | live environments}.
      */
     environment?: 'test' | 'live' | 'live-us' | 'live-au' | string;
```
emidev98, line 11: Shouldn't it be an enum if the values are fixed?

src/core/types.ts
```diff
@@ -1,36 +1,74 @@
...
 
     /**
-     * A client-side key linked to your website, used to validate Adyen’s Web component library. Use the {@link https://docs.adyen.com/api-explorer/#/CheckoutUtility/v1/originKeys /originKeys} endpoint to generate one.
-     * For more information, refer to {@link https://docs.adyen.com/user-management/how-to-get-an-origin-key How to get an origin key}.
+     * A client-side key linked to your website, used to validate Adyen’s Web component library. Use the {@link https://docs.adyen.com/api-explorer/#/CheckoutUtility/v1/originKeys | /originKeys} endpoint to generate one.
+     * For more information, refer to {@link https://docs.adyen.com/user-management/how-to-get-an-origin-key | How to get an origin key}.
      */
     originKey?: string;
```
emidev98, line 17: As far as I understand the originKey is required to instance the Web Component.
Is there any reason for being optional parameter? - because I don't see any @defaultValue

marcperez: Hey @emidev98, thanks for your comment! The originKey is mandatory for most (but not all) our components, that's why it's optional in the core. Cheers!

src/core/types.ts
marcperez: That'd make a lot of sense indeed, although at this point environment can be more things besides those values. We'll keep it in mind for a future PR, thanks! :)

PR #126, opened 2020-07-16
Feature/automatically render kcp fields

sponglord:
## Summary
Based on a combination of the config options `koreanAuthenticationRequired` & `countryCode` as well as the `issuingCountryCode` returned by the `/binLookup` endpoint - we hide/show the extra KCP fields

## Tested scenarios
- never show fields when `koreanAuthenticationRequired` is `false`
- when `koreanAuthenticationRequired` is `true`:
    - show fields at startup when `countryCode` is `"KR"`
    - don't show fields at startup if `countryCode` is not `"KR"`
    - regardless of `countryCode` show fields if `issuingCountryCode` is `"KR"`
- see all tested scenarios in `CardInput.test.tsx`

**Fixed issue**:  COWEB-782

### DIFF
--- a/playground/pages/Cards/Cards.js
+++ b/playground/pages/Cards/Cards.js
@@ -69,7 +69,7 @@ getOriginKey()
             .create('bcmc', {
                 type: 'bcmc',
                 hasHolderName: true,
-                // holderNameRequired: true,
+                //                holderNameRequired: true,
                 enableStoreDetails: false
             })
             .mount('.bancontact-field');
@@ -112,6 +112,7 @@ getOriginKey()
                 type: 'scheme',
                 brands: ['mc', 'visa', 'amex', 'bcmc', 'maestro'],
                 koreanAuthenticationRequired: true
+                //                countryCode: 'KR'
             })
             .mount('.card-kcp-field');
     });
--- a/playground/pages/Cards/Cards.js
+++ b/playground/pages/Cards/Cards.js
@@ -69,7 +69,7 @@ getOriginKey()
             .create('bcmc', {
                 type: 'bcmc',
                 hasHolderName: true,
-                //                holderNameRequired: true,
+                // holderNameRequired: true,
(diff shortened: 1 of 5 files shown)


src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -189,12 +199,23 @@ class CardInput extends Component<CardInputProps, CardInputState> {
...
         let isOneClick: boolean = this.props.storedPaymentMethodId ? true : false;
         if (this.props.oneClick === true) isOneClick = true; // In the Drop-in the oneClick status may already have been decided, so give that priority
 
+        // If the merchant defined countryCode is 'KR'
+        let isKorea = countryCode?.toLowerCase() === 'kr';
+
+        // Override the value if issuingCountryCode is set
+        if (issuingCountryCode !== null) {
+            isKorea = issuingCountryCode.toLowerCase() === 'kr';
+        }
```
pabloai, line 217: I'd enforce the format of the `issuingCountryCode` and the `countryCode` in the `formatProps` in the Card element and then just check the value in one line.
```suggestion
        const isKorea = issuingCountryCode === 'KR' || countryCode === 'KR';
```

sponglord: That doesn't cover all scenarios and the tests fails

pabloai: No worries, we can always merge and review it later.

playground/pages/Cards/Cards.js
```diff
@@ -69,7 +69,7 @@ getOriginKey()
             .create('bcmc', {
                 type: 'bcmc',
                 hasHolderName: true,
-                // holderNameRequired: true,
+                //                holderNameRequired: true,
```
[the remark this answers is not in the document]
sponglord: That's just my linter doing that

PR #195, opened 2020-08-17
Feature/reading pm configuration data values

sponglord:
## Summary
PaymentMethod objects from the `/paymentMethods` response can now contain a configuration object. This object contains merchant defined values pulled from the CA.
This means that the merchant doesn't need to individually configure components when initialising Checkout - so PMs like GooglePay, PayPal & Korean cards can be given all the information they need to run in PayByLInk.
If the merchant does configure a component when initialising Checkout then these values will overwrite those from the `/paymentMethods` response

## Tested scenarios
Tested GooglePay, PayPal & Korean cards to ensure that configuration values from the PM object from the `/paymentMethods` response are read.
Tested that these values are overwritten by any locally defined merchant configuration

**Relates to issue**:  FOC-29226, FOC-32104 & DEV-29467


src/components/PayPal/types.ts
```diff
@@ -12,7 +12,7 @@ declare global {
  * The intent for the transaction. This determines whether the funds are captured immediately, or later.
  * @see {@link https://developer.paypal.com/docs/checkout/reference/customize-sdk/#intent}
  */
-type Intent = 'sale' | 'capture' | 'authorize' | 'order';
+type Intent = 'sale' | 'capture' | 'authorize' | 'order' | '';
```
pabloai, line 15: I'd undo this change in order to make the `Intent` type more precise.
I know we have `intent: ''` in a couple of places, but we could replace the empty string with `null`.

sponglord: done

src/components/GooglePay/types.ts
```diff
@@ -24,6 +28,7 @@ export interface GooglePayPropsConfiguration {
 export interface GooglePayProps extends UIElementProps {
     environment?: google.payments.api.Environment | string;
     configuration?: GooglePayPropsConfiguration;
+    paymentMethods?: PaymentMethod[];
```
marcperez, line 31: Is there any reason why we added `paymentMethods` into the `GooglePayProps` types?

config/jest.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
         '\\.scss$': '<rootDir>/config/testMocks/styleMock.js',
         '~(.*)$': '<rootDir>/src$1'
     },
-    verbose: true,
+    verbose: false,
```
marcperez, line 14: Is this intentional? What is the motivation?

src/components/PayPal/Paypal.tsx
```diff
@@ -25,6 +25,20 @@ class PaypalElement extends UIElement<PayPalElementProps> {
...
+    // Required for transition period (until configuration object becomes the norm)
+    // - if merchant has defined properties directly in props, use these instead
+    protected formatProps(props) {
+        const { configuration, intent, merchantId } = props;
+        return {
+            ...props,
+            configuration: {
+                ...configuration,
+                ...(intent && { intent }),
+                ...(merchantId && { merchantId })
```
pabloai, line 37: Shall we do a similar thing for cards? Since `koreanAuthenticationRequired` can be both inside and outside the `configuration` object.

config/jest.config.js
sponglord: If you want verbose testing turn it on - otherwise stop cluttering up the terminal

src/components/PayPal/Paypal.tsx
sponglord: I already do:
```
configuration: {
                ...props.configuration,
                ...(props.koreanAuthenticationRequired !== undefined && { koreanAuthenticationRequired: props.koreanAuthenticationRequired })
            }
```

src/components/GooglePay/types.ts
sponglord: This can go - they were propagated in an earlier commit that was since revised

pabloai (approved):

PR #202, opened 2020-08-19
Feature: Validation Message with complex Validations

DarDevDuck:
## Summary
Complex Validations should show validation error messages to help user identify issue. For instance, issue FOC-32362 which requires that `econtext` `telephoneNumber` fields must be 10 or 11 digits.

Also enhanced `dateOfBirth` validations to provide error messages to users that they `Must be at least 18 years old`

## Affected Components
`Doku`, `Econtext`, `OpenInvoice`

## Tested scenarios
- Tested `dateOfBirth` validation message for `OpenInvoice` when age is less than 18 years old.
- Tested `telephoneNumber` validations for `eContext` flows

**Relates to issue**: FOC-32362


pabloai:
Converting to draft to avoid merging it accidentally.

src/components/Econtext/components/EcontextInput/EcontextInput.tsx
```diff
@@ -1,21 +1,24 @@
-import { h } from 'preact';
-import { useEffect, useRef, useState } from 'preact/hooks';
+import {h} from 'preact';
```
pabloai, line 1: Mmm it seems your Prettier plugin is not picking up the `prettier.config.js` file.

src/components/internal/PersonalDetails/types.ts
```diff
@@ -9,14 +10,15 @@ export interface PersonalDetailsProps {
...
 }
 
 export interface PersonalDetailsStateError {
     firstName?: boolean;
     lastName?: boolean;
     gender?: boolean;
-    dateOfBirth?: boolean;
-    telephoneNumber?: boolean;
+    dateOfBirth?: string;
+    telephoneNumber?: string;
```
pabloai, line 21: This could be `string | bolean` for cases where we don't want to show an error message.

src/utils/Validator/ValidatorWithMessage.ts
```diff
@@ -0,0 +1,39 @@
...
+const defaultRules = {
+    input: {
+        default: [new ValidationRule(()=> true, '')]
+    },
+    blur: {
+        default: [new ValidationRule(()=> true, '')]
+    }
+}
+
+class ValidatorWithMessage implements ValidatorI {
```
pabloai, line 14: Having both the `Validator` class and the `ValidatorWithMessage` class is good to comparing and testing different scenarios but before merging it would be better to remove one and stay with a class which could handle both scenarios.

pabloai (changes requested):
Adding some comments and proposals.

src/components/internal/PersonalDetails/validate.ts
```diff
@@ -1,17 +1,20 @@
-import { email, telephoneNumber } from '../../../utils/regex';
-import { unformatDate } from '../FormFields/InputDate/utils';
+import {email, telephoneNumber} from '../../../utils/regex';
+import {unformatDate} from '../FormFields/InputDate/utils';
+import ValidationRule from "../../../utils/Validator/ValidationRule";
+import Language from "../../../language/Language";
 
+const i18n = new Language();
```
sponglord, line 6: In practice - we would also need to access the existing Language instance, rather than creating a new one.
Creating a new one like this will default to `en-US` and so only give keys in that language. 
In the future we may want the merchant to pass in the translations object to use - then we would _definitely_ have to make sure to access the existing Language instance.

src/components/Econtext/components/EcontextInput/EcontextInput.tsx
DarDevDuck: Hmm. Will check it out :)

src/utils/Validator/ValidatorWithMessage.ts
DarDevDuck: Hi Pablo,

The `Validator` class is still in use. Am not too sure if you are referring to merging the classes that use `Validator` to use `ValidatorWithMessage`?

src/components/Econtext/components/EcontextInput/EcontextInput.tsx
DarDevDuck: Fixed

src/components/internal/PersonalDetails/types.ts
DarDevDuck: Fixed

src/components/internal/PersonalDetails/validate.ts
DarDevDuck: Fixed. Thanks for that. Did not know.

pabloai (approved):

pabloai:
Merging PR. Thanks for your contribution @DarDevDuck!

PR #221, opened 2020-08-26
Feature/preconfiguring aria config objects

sponglord:
## Summary
In light of recent requests around a11y: 
- a configurable `title` attribute in the <head> of the securedField iframe
- that the contents of the `aria-error` field in the iframe match that of the error displayed by the component

The `SecuredField.ts` now takes responsibility for ensuring all securedFields have a default `ariaConfig` object (with `iframeTitle` and `error`_*_ properties) if the merchant hasn't configured one themselves.
Likewise `SecuredField.ts` also takes responsibility for ensuring a default `placeholders` config object exists for each securedField if the merchant hasn't defined one.
Previously these responsibilities, to the extent that they were fulfilled, fell to the `SecuredFieldsProvider`
_* this error property is given a value pulled from the relevant translation file_

## Tested scenarios
- that pre-existing, merchant defined, aria config objects are not overwritten
- that, where an `iframeTitle` or `error` property was missing from this aria config object, they are added
- that if no aria config object exists then one is created with the default `iframeTitle` and `error` properties
- that if no placeholder config object exists one is created with a default, translated, value
- that these config objects are transmitted, received and used by the securedFields

**Fixed issue**:  Addresses some of the points in FOC-33752 (namely point _1_ and point _2h_)


src/components/internal/SecuredFields/lib/core/utils/init/processAriaConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+        };
+
+        // Add error translation
+        ariaFieldConfigWithTranslation = addErrorTranslationToObject(newAriaFieldConfigObj, fieldType, i18n, CSF_FIELDS_ARRAY);
+    } else {
+        // Create a new object with the default title
+        newAriaFieldConfigObj = { iframeTitle };
+        // Add error translation
+        ariaFieldConfigWithTranslation = addErrorTranslationToObject(newAriaFieldConfigObj, fieldType, i18n, CSF_FIELDS_ARRAY);
+    }
```
pabloai, line 33: Small suggestion: Since lines 27 and 32 are the same, you could delete line 14 and declare `ariaFieldConfigWithTranslation` as a `const` after the if/else.

sponglord: ok

pabloai (approved):

playground/config/commonConfig.js
```diff
@@ -1,8 +1,8 @@
 import getCurrency from './getCurrency';
 import { getSearchParameters } from '../utils';
 
-const DEFAULT_LOCALE = 'en-US';
-const DEFAULT_COUNTRY = 'US';
+const DEFAULT_LOCALE = 'nl-NL';
+const DEFAULT_COUNTRY = 'NL';
```
marcperez, line 5: Can we revert this?

sponglord: whoops. definitely!

PR #249, opened 2020-09-08
Feature/improve error handling for secured fields

sponglord:
## Summary
Currently for securedField based components we have a "one error fits all" policy. 
What is displayed to the shopper is a message along the lines of "Invalid {insert field name}", which we translate.

However we have a much more meaningful set of errors generated by securedFields ("card too old", "date too far in the future", "unsupported card entered" etc).

This PR takes these errors, which now have an associated code, and stores them using this code. The relevant components (CardNumber, ExpirationDate) can then translate them and display them to the shopper.

Meanwhile a securedField, when initialised, is sent an error config object containing the possible errors stored in a _code:translation_ key-value pair; which means, from an a11y perspective, that we can set the same, more meaningful, translated errors into the `aria-describedby` (error) element so they can be picked up by the screenreader.

For the custom card component we also translates these errors and add them to the error object sent to the merchant defined, `onError`, callback, along with a "developer-readable" version of the error.

## Tested scenarios
That errors referred to by code can be translated
That errors can be retrieved by either code or human-readable form
That translated, more meaningful errors are both displayed to the shopper and set for the screenreader to pick up.
That everything continues to work as expected re. setting and clearing errors

**Links to issues**:  COWEB-261, COWEB-558


src/components/internal/SecuredFields/SecuredFieldsProviderHandlers.ts
```diff
@@ -121,12 +123,23 @@ function handleOnError(cbObj: CbObjOnError, hasUnsupportedCard: boolean = null):
     if (this.state.hasUnsupportedCard && cbObj.fieldType === ENCRYPTED_CARD_NUMBER && hasUnsupportedCard === null) {
         return false;
     }
+    // console.log('### SecuredFieldsProviderHandlers::handleOnError:: cbObj.error', cbObj.error);
```
marcperez, line 126: Can we remove this commented console.logs?

src/components/internal/SecuredFields/utils.js
```diff
@@ -75,25 +75,30 @@ export const getErrorReducer = (numDateFields, state) => (acc, field) => {
...
  */
-export const getTranslatedErrors = (i18n = {}) => ({
-    [ENCRYPTED_CARD_NUMBER]: i18n.get && i18n.get('creditCard.numberField.invalid'),
-    [ENCRYPTED_EXPIRY_DATE]: i18n.get && i18n.get('creditCard.expiryDateField.invalid'),
-    [ENCRYPTED_EXPIRY_MONTH]: i18n.get && i18n.get('creditCard.expiryDateField.invalid'),
-    [ENCRYPTED_EXPIRY_YEAR]: i18n.get && i18n.get('creditCard.expiryDateField.invalid'),
-    [ENCRYPTED_SECURITY_CODE]: i18n.get && i18n.get('creditCard.oneClickVerification.invalidInput.title'),
-    defaultError: 'error.title'
-});
+//export const getTranslatedErrors = (i18n = {}) => ({
```
marcperez, line 94: It looks like this is not necessary anymore, shall we remove it?

src/components/internal/SecuredFields/SecuredFieldsProviderHandlers.ts
sponglord: yup

src/components/internal/SecuredFields/utils.js
sponglord: I thought it was gone already... looks like I didn't commit it

src/components/internal/SecuredFields/SecuredFieldsProviderHandlers.ts
sponglord: done

src/components/internal/SecuredFields/utils.js
sponglord: done

marcperez (approved):

PR #250, opened 2020-09-08
Split codebase into packages (workspaces)

marcperez:
## Summary
This PR splits the codebase into 3 packages: `adyen-web`, `playground` and `server` by leveraging Yarn Workspaces.


packages/playground/src/pages/Dropin/Dropin.js
```diff
@@ -1,4 +1,5 @@
-import AdyenCheckout from '../../../src';
+import AdyenCheckout from '@adyen/adyen-web';
+import '@adyen/adyen-web/dist/adyen.css';
```
pabloai, line 2: Why do we only load this CSS here and not in the other pages?

packages/playground/src/style/border-radius.scss
```diff
@@ -0,0 +1,4 @@
+$border-radius-small: 3px;
```
pabloai, line 1: I wouldn't duplicate all these files. If we cannot access them from the playground, that's fine, let's just remove them and not use these variables.
That being said, feel free to just merge this PR and then we can remove them in another PR.

packages/playground/src/pages/Dropin/Dropin.js
marcperez: That's fixed now 👍

packages/playground/src/style/border-radius.scss
marcperez: Fair point! I think it'd be better to refactor this in a separate PR, not to add even more changes to this one :)

packages/playground/src/style.scss
```diff
@@ -1,4 +1,4 @@
-@import '../src/style';
+@import './style/index';
```
pabloai, line 1:
```suggestion
@import './style';
```

pabloai (approved):

PR #253, opened 2020-09-08
Migrate 3DS2 components to TypeScript.

pabloai:
## Summary
Migrate 3DS2 components to TypeScript.

## Tested scenarios
No new tests were added.


src/components/ThreeDS2/ThreeDS2Challenge.tsx
```diff
@@ -2,18 +2,22 @@ import { h } from 'preact';
...
+    type?: string;
+}
+
+class ThreeDS2ChallengeElement extends UIElement<ThreeDS2ChallengeElementProps> {
     public static type = 'threeDS2Challenge';
 
     public static defaultProps = {
         dataKey: 'threeds2.challengeResult',
-        challengeContainer: null,
         size: '01',
```
sponglord, line 20: Why have you removed the `null` default props?

src/components/ThreeDS2/components/Challenge/Do3DS2Challenge.tsx
```diff
@@ -7,15 +7,20 @@ import getProcessMessageHandler from '../../../../utils/get-process-message-hand
...
+    private readonly iframeCallback;
+    private processMessageHandler;
+    private challengePromise: { cancel: () => void; promise: Promise<any> };
+
     constructor(props) {
         super(props);
 
         this.iframeCallback = () => {
-            this.setState({ status: 'iframeLoaded' });
+            this.state = { status: 'iframeLoaded' };
```
sponglord, line 23: Is it ok to set state directly?

src/components/ThreeDS2/components/utils.ts
```diff
@@ -1,11 +1,26 @@
...
+interface ErrorCode {
+    errorCode?: string;
+    message?: string;
+}
+
+export const decodeAndParseToken = (token: string): ChallengeToken => {
     const decodedToken = base64.decode(token);
     try {
-        return JSON.parse(decodedToken);
+        if (decodedToken) return JSON.parse(decodedToken);
```
sponglord, line 23: Is the `if` needed? That's why this is in a `try...catch`

src/components/ThreeDS2/components/Challenge/Do3DS2Challenge.tsx
marcperez: I think this function should be moved out of the constructor

src/components/ThreeDS2/ThreeDS2Challenge.tsx
pabloai: Well, does it add any value to set it to `null` instead of just being undefined?
I might be missing something.

src/components/ThreeDS2/components/Challenge/Do3DS2Challenge.tsx
pabloai: Yes, that seems like the best solution here.

src/components/ThreeDS2/ThreeDS2Challenge.tsx
sponglord: Just that it adds a clear indication of what all the props will be, once they are defined.
It became quite complicated keeping track of what particular 3DS2 processes were expecting to be sent so this presented a clear list, for reference.

pabloai: Well, we now have `ThreeDS2ChallengeElementProps` to list all the props, right?

sponglord: Ok. But it does miss `challengeContainer`

src/components/ThreeDS2/components/Challenge/Do3DS2Challenge.tsx
sponglord: Mine was also a question for my own understanding. `setState` or `this.state`?

src/components/ThreeDS2/ThreeDS2Challenge.tsx
pabloai: Ah, I see. I'll check this, thanks for pointing it out!

src/components/ThreeDS2/components/Challenge/Do3DS2Challenge.tsx
pabloai: The rule of thumb is that you use `this.state` in the constructor and `this.setState` everywhere else.

src/components/ThreeDS2/components/utils.ts
pabloai: Yes, but just because `base64.decode` can return either a string or `false`, while JSON.parse expects a string, so TS was complaining about that.

src/components/ThreeDS2/ThreeDS2Challenge.tsx
pabloai: It seems `challengeContainer` is not used anymore, so it's fine to remove it.

src/components/ThreeDS2/components/utils.ts
sponglord: Fair enough!

src/components/ThreeDS2/components/Challenge/Do3DS2Challenge.tsx
sponglord: ok, thanks

sponglord (approved):

PR #273, opened 2020-09-11
Feature/PayPal commit property

dkatashev:
## Summary
The main purpose of this Pull Request is to manage commit status of the transaction - https://developer.paypal.com/docs/checkout/reference/customize-sdk/#commit.

This will expand the possibilities and make possible the next feature - https://developer.paypal.com/docs/checkout/integration-features/confirmation-page/.

### DIFF
--- a/packages/lib/src/components/PayPal/Paypal.tsx
+++ b/packages/lib/src/components/PayPal/Paypal.tsx
@@ -28,13 +28,12 @@ class PaypalElement extends UIElement<PayPalElementProps> {
     // Required for transition period (until configuration object becomes the norm)
     // - if merchant has defined properties directly in props, use these instead
     protected formatProps(props) {
-        const { configuration, intent, commit, merchantId } = props;
+        const { configuration, intent, merchantId } = props;
         return {
             ...props,
             configuration: {
                 ...configuration,
                 ...(intent && { intent }),
-                ...{ commit },
                 ...(merchantId && { merchantId })
             }
         };
--- a/packages/lib/src/components/PayPal/Paypal.tsx
+++ b/packages/lib/src/components/PayPal/Paypal.tsx
@@ -28,12 +28,13 @@ class PaypalElement extends UIElement<PayPalElementProps> {
     // Required for transition period (until configuration object becomes the norm)
     // - if merchant has defined properties directly in props, use these instead
     protected formatProps(props) {
-        const { configuration, intent, merchantId } = props;
+        const { configuration, intent, commit, merchantId } = props;
         return {
             ...props,
             configuration: {
                 ...configuration,
                 ...(intent && { intent }),
+                ...{ commit },
                 ...(merchantId && { merchantId })
             }
         };
--- a/packages/lib/src/components/PayPal/utils.ts
+++ b/packages/lib/src/components/PayPal/utils.ts
@@ -28,7 +28,7 @@ const getPaypalSettings = ({ amount, countryCode, debug, environment = '', local
         ...(debug && isTestEnvironment && { debug }),
         ...(currency && { currency }),
         ...(intent && { intent }),
-        ...{ commit },
+        commit,
         'client-id': clientId,
         'integration-date': INTEGRATION_DATE,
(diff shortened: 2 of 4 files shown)


packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -28,12 +28,13 @@ class PaypalElement extends UIElement<PayPalElementProps> {
...
     // - if merchant has defined properties directly in props, use these instead
     protected formatProps(props) {
-        const { configuration, intent, merchantId } = props;
+        const { configuration, intent, commit, merchantId } = props;
         return {
             ...props,
             configuration: {
                 ...configuration,
                 ...(intent && { intent }),
+                ...{ commit },
```
pabloai, line 37: No need to add this here. Actually there's no need to change this file at all, as the `commit` property is already included in the props.

packages/lib/src/components/PayPal/utils.ts
```diff
@@ -19,7 +19,7 @@ const getPaypalSettings = ({ amount, countryCode, debug, environment = '', local
     const isTestEnvironment: boolean = environment.toLowerCase() === 'test';
     const clientId: string = isTestEnvironment ? ADYEN_CLIENTID_TEST : ADYEN_CLIENTID_LIVE;
 
-    const { merchantId, intent } = configuration;
+    const { merchantId, intent, commit } = configuration;
```
pabloai, line 22: The `commit` property will arribe inside the `props`, not this `configuration` object.

pabloai (changes requested):
Thanks for the contribution!
I added some comments to make it ready to be merged. Cheers.

dkatashev:
Hi @pabloai,
I did the same as the **intent** attribute.
Maybe I didn't fully understand the idea. According to the code - PayPal **configuration** this is the place where such information should be stored, it is also present in unit tests.

Tell me in this case, where is it better to store this attribute?
Thanks in advance

pabloai:
Hi again @dkatashev,

The `intent` property is a special case because we receive it from the backend in the response from the `paymentMethods` endpoint. On this case it should get to the component in the `props` so there's no need of any special handling.

That being said, if I'm not mistaken this `commit` property is used in the Express Checkout flow. Is that what you're aiming to achieve? Would you do the call to the `/payments/details` endpoint later in the flow?

Just trying to understand the use case. Thanks!

dkatashev:
Hi @pabloai,
Thank you for your detailed explanation.
Yes, you're right. I want to make possible PayPal Express Checkout Flow.
I guess I'll be able to initiate a component with that option first. Then on the server get information from PayPal about the customer (payer) based on PayPal Order ID. Then, do the call to the `/payments/details` to confirm the order.
Do you see any problems to make it possible?
I'll be very grateful for your feedback.

packages/lib/src/components/PayPal/utils.ts
```diff
@@ -28,6 +28,7 @@ const getPaypalSettings = ({ amount, countryCode, debug, environment = '', local
         ...(debug && isTestEnvironment && { debug }),
         ...(currency && { currency }),
         ...(intent && { intent }),
+        ...{ commit },
```
pabloai, line 31: There's no need to use the spread operator here. Just using the shorthand is enough.
```suggestion
        commit,
```

pabloai (commented):
The PayPal Express Checkout is a flow we are not focusing on (hence the lack of this `commit` prop). I guess it could be done if you store the data instead of doing the `payments/details` call when the `onAdditionalDetails` is called and then do the call once the shopper confirms the order. We can merge this PR so you can test this flow and see if it works for you. I just added a small comment, but it's almost ready to be merged.

packages/lib/src/components/PayPal/utils.ts
dkatashev: Updated

pabloai:
Merging this PR.
Thanks for your collaboration @dkatashev !

pabloai (approved):

PR #288, opened 2020-09-15
Fix/Delete unused file

pabloai:
## Summary
Delete unused CSS file.


packages/lib/src/language/locales/en-US.json
```diff
@@ -72,9 +72,6 @@
     "issuerList.wallet.placeholder": "Select your wallet",
     "privacyPolicy": "Privacy policy",
     "afterPay.agreement": "I agree with the %@ of AfterPay",
-    "companyDetails": "Company details",
```
marcperez, line 75: Does this change belong here?

pabloai: No, thanks for pointing it out.

marcperez (approved):

PR #330, opened 2020-09-22
Feature/japanese installments

sponglord:
## Summary
For Japanese installments, an extra field in the form of a "plan type" is required. This plan can be set to either "regular", where a number of installments are selected, or "revolving", where no number is selected.
In order to set this plan type a new UI is required showing 3 radio buttons as well as the dropdown (see image in COWEB-785)

## Tested scenarios
- Radio button UI renders when a `plans` property exists in the `installmentOptions` for a particular card brand
- This UI is removed for 'regular' installments i.e. the usual 'installments' UI is rendered (dropdown only)
- The object created to be submitted in the call to the `/payments` end point (under the key `installments`) contains an extra `plan` property, set to `revolving` when the relevant radio button is selected
- For all other use-cases only the expected `value` property is set in the `installments` object

**Fixed issue**:  COWEB-785


packages/lib/src/components/Card/components/CardInput/CardInput.scss
```diff
@@ -108,3 +108,23 @@
...
+        display: flex;
+        flex-direction: column;
+    }
+}
+
+.adyen-checkout__fieldset--revolving-plan{
+    .adyen-checkout__radio_group__input-wrapper{
+        margin-top: 20px
+    }
+}
```
marcperez, line 123:
```suggestion
.adyen-checkout__fieldset--revolving-plan {
    .adyen-checkout__radio_group {
        display: flex;
        flex-direction: column;
    }

    .adyen-checkout__radio_group__input-wrapper {
        margin-top: 20px
    }
}
```

There's a duplicate CSS selector here

packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -156,6 +157,8 @@ class CardInput extends Component<CardInputProps, CardInputState> {
         // If issuingCountryCode is set or the merchant defined countryCode is 'KR'
         const isKorea = issuingCountryCode ? issuingCountryCode === 'kr' : countryCode === 'kr';
 
+        const showAmountsInInstallments = existy(showInstallmentAmounts) ? showInstallmentAmounts : true;
```
pabloai, line 160:
```suggestion
        const showAmountsInInstallments = showInstallmentAmounts ?? true;
```

pabloai (approved):

PR #332, opened 2020-09-22
DropinComponent - add public method 'closeActivePaymentMethod'

mrkosima:
DropIn component has a few methods and props to control and handle selection, like `openFirstPaymentMethod` or `onSelect()`.

But in some cases I need to reset the selected payment method in an already mounted component.
At the moment there's no public API for that.

So, this PR adds `closeActivePaymentMethod ` to reset selected payment method.

### DIFF
--- a/packages/lib/src/components/Dropin/Dropin.test.ts
+++ b/packages/lib/src/components/Dropin/Dropin.test.ts
@@ -24,4 +24,15 @@ describe('Dropin', () => {
             expect(() => dropin.submit()).toThrow();
         });
     });
+
+    describe('closeActivePaymentMethod', () => {
+        test('should close active payment method', () => {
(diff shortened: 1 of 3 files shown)


pabloai:
Hi @mrkosima, and thanks for collaborating!

I think this is a good idea, but personally I feel the name of the method can be misleading, as calling the `resetActivePaymentMethod` would not reset the Drop-in to its initial state (which could be with the first payment method being open).

Maybe it could be renamed to `closeActivePaymentMethod`.

Let me know what you think. Cheers!

mrkosima:
Hi @pabloai , thank you for the quick response!

Agree with you, please check out latest changes in the PR!

pabloai (approved):

pabloai:
Merging. Thanks again!

PR #338, opened 2020-09-24
Fix/unsupported card error not clearing

sponglord:
## Summary
With `/binLookup` in operation, when entering an unsupported card, an error is generated. 
This error wasn't clearing if you then _pasted_ a full, supported card number into the field and no payment could be made.
This was happening because the encrypted blob for the new, supported, card number was being ignored because the "unsupported card" error was still in place. 
The binLookup for this new, supported, card number happens after this event and clears the error - but this is too late.

The solution is to inform the card number securedField when it is in an "unsupported card" error state (previously we we keeping this level of knowledge within Components) - and to inform it again when this error is cleared. In response to this second message the card number securedField sends it's encryptedBlob a second time

## Tested scenarios
Tested above scenarios.
Now when pasting in a full, supported, card number the error clears and the payment can proceed.
Adjusted tests accordingly


packages/lib/src/components/internal/SecuredFields/SecuredFieldsProviderHandlers.ts
```diff
@@ -119,9 +119,15 @@ function handleOnBrand(cardInfo: CbObjOnBrand): void {
  * Handles validation errors
  */
 function handleOnError(cbObj: CbObjOnError, hasUnsupportedCard: boolean = null): boolean {
+    console.log('\n### SecuredFieldsProviderHandlers::handleOnError:: this.state.hasUnsupportedCard', this.state.hasUnsupportedCard);
```
marcperez, line 122: Is this intentional? Shall we remove them or comment them with `process.env` conitional?

sponglord: No, it's not. I thought I'd removed all logs

sponglord: fixed

packages/lib/src/components/internal/SecuredFields/SecuredFieldsProvider.ts
```diff
@@ -204,6 +209,21 @@ class SecuredFieldsProvider extends Component<SFPProps, SFPState> {
     }
 
     public processBinLookupResponse(binValueObject: BinLookupObject): void {
+        console.log('### SecuredFieldsProvider::processBinLookupResponse:: binValueObject', binValueObject);
+
+        console.log('### SecuredFieldsProvider::processBinLookupResponse:: is currently in unsupportedCard state=', this.state.hasUnsupportedCard);
+        console.log('### SecuredFieldsProvider::processBinLookupResponse:: state.errors=', this.state.errors);
```
marcperez, line 215: Some more logs

packages/lib/src/components/internal/SecuredFields/SecuredFieldsProvider.ts
```diff
@@ -170,8 +171,12 @@ class SecuredFieldsProvider extends Component<SFPProps, SFPState> {
     }
 
     public handleUnsupportedCard(errObj: CbObjOnError): boolean {
+        console.log('\n### SecuredFieldsProvider::handleUnsupportedCard:: errObj', errObj);
         const hasUnsupportedCard = !!errObj.error;
         this.handleOnError(errObj, hasUnsupportedCard);
+        // Inform CSF that the number field has an unsupportedCard error
+        console.log('### SecuredFieldsProvider::handleUnsupportedCard:: this.csf=', this.csf);
```
marcperez, line 178: And some more logs, otherwise looks good to me! We can merge after this 👍

packages/lib/src/components/internal/SecuredFields/lib/core/CSF.ts
```diff
@@ -162,6 +162,31 @@ class CSF extends AbstractCSF {
...
             },
+            hasUnsupportedCard: (pFieldType: string, code: string): CSFReturnObject => {
+                if (this.state.isConfigured) {
+                    if (Object.prototype.hasOwnProperty.call(this.state.securedFields, pFieldType)) {
+                        //
+                        this.state.securedFields[pFieldType].hasError = !!code;
+                        this.state.securedFields[pFieldType].errorType = code;
+
+                        console.log('### CSF::hasUnsupportedCard:: hasError', !!code);
+                        console.log('### CSF::hasUnsupportedCard:: errorType', code);
```
marcperez, line 173: Some logs

packages/lib/src/components/internal/SecuredFields/SecuredFieldsProvider.ts
sponglord: wtf! I removed all these! I actively went in and removed them because I know I'm prone to leaving log statements in the code!!!??? 
I'll fix it!

marcperez (approved):

PR #339, opened 2020-09-25
Update MBWay component

pabloai:
## Summary
- Remove unneeded email field.
- Extract type definition to their own file.
- Align code style to other similar components.
- Delete unneeded CSS module file.

**Fixed issue**: COWEB-811

### DIFF
--- a/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
+++ b/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
@@ -2,53 +2,22 @@ import { h } from 'preact';
 import { useState, useEffect } from 'preact/hooks';
 import classNames from 'classnames';
 import useCoreContext from '../../../../core/Context/useCoreContext';
-import styles from './MBWayInput.module.scss';
-import './MBWayInput.scss';
 import Field from '../../../internal/FormFields/Field';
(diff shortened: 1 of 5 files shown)


packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -57,90 +26,49 @@ function MBWayInput(props: MBWayInputProps) {
...
+        const { value, isValid }: ValidationObj = validator.validate('telephoneNumber', mode)(val);
 
-        const { value, isValid }: ValidationObj = validator.validate(key, type)(val);
-
-        if (type === 'input') {
-            setData({ ...data, [key]: value });
-            setErrors({ ...errors, [key]: false });
-        }
-
-        if (type === 'blur') {
```
sponglord, line 75: By removing this the following scenario is again possible:
- the phone number has less than 7 digits
- it shows an error
- you delete all the digits and click away from the field
- it continues to show the error until 7 digits have been entered
Is this deliberate? We don't normally consider an empty field to be in error until the pay button is pressed

pabloai: That's a good point. I went and tried some components and we don't have a consistent behaviour for this.

sponglord (approved):

PR #362, opened 2020-10-01
Feature/COWEB-816 Automatically create Apple Pay session

pabloai:
## Summary
Add `validateMerchant` method to automatically create an Apple Pay session.


packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -91,6 +95,28 @@ class ApplePayElement extends UIElement<ApplePayElementProps> {
...
 
+    private validateMerchant(resolve, reject) {
+        const { hostname: domainName } = window.location;
+        const { clientKey, configuration, loadingContext, initiative } = this.props;
+        const { merchantName: displayName, merchantIdentifier } = configuration;
+        const path = `${APPLEPAY_SESSION_ENDPOINT}?token=${clientKey}`;
+        const options = { loadingContext, path, method: 'post' };
+        const request: ApplePaySessionRequest = { displayName, domainName, initiative, merchantIdentifier };
+
+        return fetchJsonData(options, request)
```
marcperez, line 106: Maybe, since we already use a try-catch block we could use async/await here.

marcperez (approved):

packages/lib/src/components/ApplePay/ApplePay.tsx
pabloai: Good idea 👍

PR #369, opened 2020-10-05
New gift card flow

marcperez:
## Summary
- Gift cards will now implement the new `onBalanceCheck` event, which gives you a chance to call the `/paymentMethods/balance` endpoint on our API.
- If the balance is not enough to complete the transaction, the `onOrderRequest` event will be fired. This is your chance to create an order in the `/orders` endpoint.
- Drop-in supports a new `order` prop, which will fetch the order information and display a summary of the payments related to that order.
- Once an order is created, you should fetch `/paymentMethods` to retrieve the payment methods available to complete that order.
- The main instance supports a new `.update()` method to, for example, update the `paymentMethodsResponse` parameter after creating a new order.


packages/lib/src/components/Dropin/components/PaymentMethod/OrderPaymentMethods.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Remaining amount warning message
+.adyen-checkout__order-remaining-amount {
+    display: block;
+    width: 100%;
+    padding: 8px 16px;
+    color: #7F4A00;
+    background: #FFEACC;
+    border-radius: 6px;
```
pabloai, line 71:
```suggestion
    border-radius: $border-radius-medium;
```

packages/lib/src/components/Dropin/components/PaymentMethod/OrderPaymentMethods.scss
```diff
@@ -0,0 +1,78 @@
...
+.adyen-checkout__order-payment-methods-list {
+    list-style: none;
+    margin: 0 auto 16px;
+    padding: 0;
+}
+
+.adyen-checkout__order-payment-method {
+    position: relative;
+    background: #fff;
+    border: 1px solid #e6e9eb;
```
pabloai, line 12:
```suggestion
    border: 1px solid $color-gray-light;
```

packages/lib/src/components/Dropin/components/PaymentMethod/OrderPaymentMethods.scss
```diff
@@ -0,0 +1,78 @@
...
+
+.adyen-checkout__order-payment-methods-list {
+    list-style: none;
+    margin: 0 auto 16px;
+    padding: 0;
+}
+
+.adyen-checkout__order-payment-method {
+    position: relative;
+    background: #fff;
```
pabloai, line 11:
```suggestion
    background: $color-white;
```

packages/lib/src/components/Dropin/components/PaymentMethod/OrderPaymentMethods.scss
```diff
@@ -0,0 +1,78 @@
...
+    font-weight: 500;
+}
+
+// Remaining amount warning message
+.adyen-checkout__order-remaining-amount {
+    display: block;
+    width: 100%;
+    padding: 8px 16px;
+    color: #7F4A00;
+    background: #FFEACC;
```
pabloai, line 70: Maybe we can add these two to our colors SASS file?

marcperez: Good catch @pabloai! I'll add them 👍

pabloai (approved):

Issue #341, opened 2020-09-25
webdropIn PayPal Button is shown with showPayButton : false configuration in since version 3.7.0


AlexanderSeel:
**Describe the bug**
We used version 3.6.3 and in configuration showPayButton : false, because we use our own button.
When we try to update to a version above 3.6.3 it shows always at the paypal payment the "PayPal" button and payment with our own button, results in js exception.

**To Reproduce**
 Use Version webdrop above 3.6.3
set const configuration = {{
            showPayButton:false,

**Expected behavior**
Hide the paypal button and process payment like in version 3.6.3

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser chrome

pabloai:
Hi @AlexanderSeel, and thanks for reaching out.

The PayPal flow cannot be fire triggered by calling some method (as discussed in paypal/paypal-checkout-components#522) I'm not sure how useful it could be to hide the button completely.

This was working before because it was a completely different integration which relied in a redirection instead of opening the PayPal pop up.

Could you describe you use case? Would you use the PayPal component in some other part of the page?
Cheers!

pabloai:
Hi @AlexanderSeel,
We've added this in version 3.16.0. Please have in consideration what I mention in the message above.

Feel free to reach out if you have more questions or ideas, cheers!

AlexanderSeel:
Hi, 

sorry for the late response.
We integrated the flow with redirect, so with our own button we trigger the payment in the web dropin, send the response to our backend call your c# lib method, take the response and give it back to the web drop in.
The question is: Is this still possible with paypal or not?
Kind regards

PR #392, opened 2020-10-12
Feature/Update PayPal component to accept the vault and the showPayButton props.

pabloai:
## Summary
Update PayPal component to accept the vault and the `showPayButton` props.
When `showPayButton` is set to `false`, the PayPal buttons won't render.
Keep in mind the payment flow cannot be triggered programatically so it would require to mount the PayPal component somewhere else on the page.

**Fixed issue**:  #341

### DIFF
--- a/packages/lib/src/components/PayPal/Paypal.tsx
+++ b/packages/lib/src/components/PayPal/Paypal.tsx
@@ -104,6 +104,8 @@ class PaypalElement extends UIElement<PayPalElementProps> {
     }
 
     render() {
+        if (!this.props.showPayButton) return null;
+
         return (
             <PaypalComponent
                 ref={ref => {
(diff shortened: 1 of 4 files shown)


marcperez (approved):

frenic:
Is it possible to display the Paypal button some how even when `showPayButton` is set to `false`? I would rather disable my custom submit button when Paypal is chosen to make it work.

pabloai:
Hi @frenic, thanks for reaching out.

PayPal doesn't allow a custom button to be used. What you could do is to mount the PayPal component where you would place your custom button. You can find more information about on Adyen/adyen-web#879.
Feel free to add your questions there.

frenic:
Thank you for your response 👏

PR #397, opened 2020-10-13
revolving installments alignment issue

sponglord:
## Summary
Fixed possibility that the installments dropdown could overlap the text by removing fixed position and adding margin-left

## Tested scenarios
As the label text changes (due to translations) - the dropdown stays the required distance to the right

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.scss
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.scss
@@ -110,7 +110,7 @@
 }
 
 .adyen-checkout__fieldset--revolving-plan {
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/Card/components/CardInput/CardInput.scss
```diff
@@ -110,6 +110,10 @@
 }
 
 .adyen-checkout__fieldset--revolving-plan {
+    .adyen-checkout__fieldset__fields{
```
marcperez, line 113:
```suggestion
    .adyen-checkout__fieldset__fields {
```

marcperez (approved):

PR #412, opened 2020-10-17
New 3DS2 flow

sponglord:
## Summary
To reduce complexity for the merchant, and to align with other components, we have improved the 3DS2 integration flow.
Now, a card payment that leads to a 3DS2 fingerprint and challenge, will follow the same pattern as other "action" payment methods: 
- a call to the `/payments` endpoint, 
- a response containing an `action`
- a final `/details` call to finalise the payment.

For 3DS2 this means we have replaced the additional call to the `/details` endpoint that was involved in fingerprinting with a call to a new endpoint. This is handled internally within the component and requires no additional integration work by the merchant.

To have this fully work the merchant needs to bump their API version so that the `/payments` call returns an `action` with the new `type:'threeDS2'`.

## Tested scenarios
- For standalone Card Component:
    - Frictionless flow (fingerprint only, no challenge required) using both `handleAction` and `checkout.createFromAction`
    - Full flow (fingerprint & challenge) using both `handleAction` and `checkout.createFromAction`
    - Challenge only flow (no fingerprint) using both `handleAction` and `checkout.createFromAction`
    - Redirect flow using both `handleAction` and `checkout.createFromAction`
- For Dropin:
    - the 4 scenarios above via `handleAction`
- Tested the "old" flow: handling actions by creating a new 3DS2 component - `checkout.create('threeDS2DeviceFingerprint')` and `checkout.create('threeDS2Challenge')`
- Tested that even if the merchant doesn't bump their API version the combination of "old" action type and calling of new endpoint can still work and leads to this new optimised flow (without the need for an additional `/details` call)

**Fixed issue**:  DEV-31189, COWEB-834


packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -82,14 +82,27 @@ class DropinElement extends UIElement<DropinElementProps> {
...
+
+        // action.type === 'threeDS2Fingerprint' creates backwards compatibility with old API versions
+        const threeDS2Options =
+            action.type === 'threeDS2' || action.type === 'threeDS2Fingerprint'
+                ? {
+                      elementRef: this.elementRef,
+                      ...(this.props.challengeWindowSize && { challengeWindowSize: this.props.challengeWindowSize }),
+                      ...(pmConfig.challengeWindowSize && { challengeWindowSize: pmConfig.challengeWindowSize })
+                  }
+                : null;
```
marcperez, line 98: Could this be moved into `src/core/ProcessResponse/PaymentAction/actionTypes.ts`?

sponglord: How? - these are all props & config specific to the Dropin element that need to be passed on in order to configure the action component

marcperez: On both `actionTypes.ts` and the ThreeDS2 component itself you should be able to access a prop named `isDropin` that will be injected when the component has been created on a Drop-in context. 
Using that, you could set those specific options in those places, instead of the more generic Dropin class.

sponglord: But `isDropin` is just a boolean. 
So it doesn't contain the `this` that is required to get the `elementRef` nor does it contain the `pmConfig` necessary to asses which level of `challengeWindowSize` (and further down in the code, `onError`) we need for the action. 
The problem here is less what I'm trying to do for 3DS2 and more a deeper one of which (of the 3 possible levels) of config properties we take to pass onto the action. 
And this `handleAction` function is surely where we configure and collect those properties prior to handing them on to the action.
So yes, I could find a way to do all this logic in the `actionTypes.ts` (always pass `elementRef` and `pmConfig` to the action so _it_ can do the sorting out of what it needs...) but is that really the place to do it?
Is the problem here that you feel that having the code here is _dirtying_ the `Dropin.tsx` component?

sponglord: We could just do this:
```
handleAction(action: PaymentAction) {
        if (!action || !action.type) throw new Error('Invalid Action');

        if (this.activePaymentMethod.updateWithAction) {
            return this.activePaymentMethod.updateWithAction(action);
        }

        const pmType = action.paymentMethodType || 'scheme';

        // Extract desired props that we need to pass on from the pmConfiguration for this particular PM
        const pmConfig = getComponentConfiguration(pmType, this.props.paymentMethodsConfiguration);

        const paymentAction: UIElement = this.props.createFromAction(action, {
            isDropin: true,
            onAdditionalDetails: state => this.props.onAdditionalDetails(state, this),
            onError: this.props.onError, // Add ref to onError in case the merchant has defined one in the component options
            ...(pmConfig?.onError && { onError: pmConfig.onError }), // Overwrite ref to onError in case the merchant has defined one in the pmConfig options
            elementRef: this.elementRef,
            ...(this.props.challengeWindowSize && { challengeWindowSize: this.props.challengeWindowSize }),
            ...(pmConfig.challengeWindowSize && { challengeWindowSize: pmConfig.challengeWindowSize })
        });

        if (paymentAction) {
            return this.setStatus(paymentAction.props.statusType, { component: paymentAction });
        }
        return null;
    }
```

marcperez: I think it makes sense to send down `elementRef` and `this.props.paymentMethodsConfiguration`, and then do the actual mapping in either the actionType function or the component itself.

packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -11,6 +11,7 @@ class DropinElement extends UIElement<DropinElementProps> {
     public static type = 'dropin';
     protected static defaultProps = defaultProps;
     public dropinRef = null;
+    private componentFromAction = null;
```
pabloai, line 14: Is this `componentFromAction` used somewhere other than on the tests?

packages/lib/src/components/Dropin/Dropin.tsx
sponglord: Now send `elementRef` and spread all `props` and `paymentMethodsConfiguration` into the options object sent to the action

packages/lib/src/components/Dropin/Dropin.tsx
sponglord: No - but I don't know how else to make the `paymentAction` available to the tests since `handleAction` in the Dropin, unlike in UIElement, doesn't return the paymentAction

sponglord: Have made the creation of this prop conditional on `process.env.NODE_ENV==="test"`

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (data.action?.type === 'authenticationFinished') {
+            const detailsObj = {
+                data: {
+                    // Sending no details property should work - BUT currently doesn't in the new flow
+                    details: { 'threeds2.challengeResult': btoa('{"transStatus":"Y"}') },
+                    paymentData: data.action.paymentData
+                }
+            };
+
+            return UIElement.prototype.onComplete.call(this, detailsObj);
```
marcperez, line 41: Since this method is bound to the `ThreeDS2DeviceFingerprint`, is there any reason not to call `this.onComplete(details)`?

packages/lib/src/components/UIElement.tsx
```diff
@@ -112,9 +113,19 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> {
...
-            onError: this.props.onError // Add ref to onError in case the merchant has defined one in the component options
+        // const threeDS2Options =
+        //     action.type === 'threeDS2' ? { ...(this.props.challengeWindowSize && { challengeWindowSize: this.props.challengeWindowSize }) } : null;
+
+        // const paymentAction: UIElement = this.props.createFromAction(action, {
+        //     onAdditionalDetails: state => this.props.onAdditionalDetails(state, this.elementRef),
+        //     // Add ref to onError in case the merchant has defined one in the component options
+        //     onError: this.props.onError,
+        //     ...threeDS2Options
+        // });
```
marcperez, line 124: Can this go?

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
```diff
@@ -0,0 +1,59 @@
...
+         */
+        if (data.action?.type === 'threeDS2') {
+            return actionHandler.handleAction(data.action);
+        }
+
+        /**
+         * Redirect (usecase: we thought we could do 3DS2 but it turns out we can't)
+         */
+        if (data.action?.type === 'redirect') {
+            data.action.paymentMethodType = 'scheme';
```
marcperez, line 55: Why do we need to override the `paymentMethodType`?

marcperez (changes requested):

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
sponglord: Because it's a method on the super, `UIElement` class, not the `ThreeDS2DeviceFingerprint` component. Calling `this.onComplete` would trigger an infinite loop

packages/lib/src/components/UIElement.tsx
sponglord: yes

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
sponglord: Because the response for the `redirect` action is not returning a `paymentMethodType` - so the `handleAction` for "redirect" will fail

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
sponglord: yup - you're right. This is a throwback to when I was overwriting the onComplete in `ThreeDS2DeviceFingerprint`

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
sponglord: It's not clear yet how easy it is for the endpoint to set this value in the response

sponglord: This has been changed in the endpoint - we no longer need to set paymentMethodType on "redirect" action

marcperez (approved):

PR #417, opened 2020-10-19
Amazon Pay component

pabloai:
## Summary
* Add component for Amazon Pay, which loads the Amazon Pay script and handles the full payment flow.
* Add Amazon Pay to the Wallets section in the playground.


packages/lib/src/components/AmazonPay/AmazonPay.scss
```diff
@@ -0,0 +1,11 @@
+@import '../../style';
+
+.adyen-checkout__amazonpay__button {
+    margin: auto;
+}
+
+.adyen-checkout__button--ghost {
```
marcperez, line 7: Should this belong to `Button.scss`? Or is really specifc to AmazonPay?

packages/lib/src/components/AmazonPay/AmazonPay.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export class AmazonPayElement extends UIElement<AmazonPayElementProps> {
+    public static type = 'amazonpay';
+    protected static defaultProps = defaultProps;
+
+    formatProps(props) {
+        return {
+            ...props,
+            environment: props.environment.toUpperCase(),
+            locale: props.locale.replace('-', '_'),
```
marcperez, line 18: Have we considered using `this.props.i18n.locale` instead? (The resolved locale, for example: `es-AR` => `es-ES`) Or are do we really want the locale as specified by the merchant?

packages/playground/src/pages/Wallets/Wallets.js
```diff
@@ -21,17 +21,74 @@ getOriginKey()
...
+        const amazonCheckoutSessionId = urlSearchParams.get('amazonCheckoutSessionId');
+        const step = urlSearchParams.get('step');
+
+        // Initial state
+        if (!amazonCheckoutSessionId) {
+            window.amazonpay = checkout
+                .create('amazonpay', {
+                    currency: 'GBP',
+                    environment: 'test',
+                    merchantId: 'A3SKIS53IXYBBU',
```
marcperez, line 42: Should this go under a `configuration` object?

packages/lib/src/components/AmazonPay/AmazonPay.scss
pabloai: These styles are only for Amazon Pay. I'll add a selector to limit the scope of the style. Good catch, cheers!

packages/lib/src/components/AmazonPay/AmazonPay.tsx
pabloai: Both the Amazon Pay SDK and the PayPal SDK accept a different subset of locales (see `packages/lib/src/components/AmazonPay/config.ts`).
We could use `this.props.i18n.locale` and transform `en_US` to `en_GB` for the `EU` region but I don't know if it would unclear for merchants. What do you think?

packages/playground/src/pages/Wallets/Wallets.js
pabloai: Well on one hand that would align with other components, but on another we're not planning to use the `/paymentMethods` response so we wouldn't be able to get these fields from the API, so we cannot completely align it.

pabloai: I just checked and we only store the `publicKeyId` in the CA.

marcperez (approved):

pabloai:
After some discussions we decided we'll add some extra functionalities to the component before releasing it.
Changing this to [Draft] to not merging it by mistake.

packages/playground/src/config/commonConfig.js
```diff
@@ -2,7 +2,7 @@ import getCurrency from './getCurrency';
 import { getSearchParameters } from '../utils';
 
 const DEFAULT_LOCALE = 'en-US';
-const DEFAULT_COUNTRY = 'US';
+const DEFAULT_COUNTRY = 'GB';
```
sponglord, line 5: Is this deliberate?

pabloai: Not really, I'll roll it back. Thanks for checking!

marcperez (approved):

PR #430, opened 2020-10-21
Improve build process

marcperez:
## Summary
- Migrate from Webpack to Rollup
- Export multiple formats (UMD, CJS, ESM)
- Add language code splitting to ESM build


packages/lib/config/rollup.config.js
```diff
@@ -0,0 +1,86 @@
...
+const currentVersion = require('./version')();
+
+if (process.env.CI !== 'true') {
+    console.warn(
+        '\x1b[33m%s\x1b[0m',
+        'Warning: Building custom bundle. We recommend using one of the official builds served by our servers or NPM. Check https://docs.adyen.com/checkout for more information.'
+    );
+}
+
+let cache;
```
pabloai, line 17: Any reason for this to be a `let` instead of a `const`?

pabloai (approved):

packages/lib/config/rollup.config.js
marcperez: Not anymore! Good catch 👍

PR #431, opened 2020-10-22
Remove `details` dependency

marcperez:
## Summary
- Remove `/paymentMethods` response `details` property dependencies
- IssuerList components (iDEAL, DotPay, EPS...) use the new `issuers` property
- We assume every payment method not listed in the `componentsMap` follows a redirect flow.

### DIFF
--- a/packages/lib/src/components/Dragonpay/Dragonpay.tsx
+++ b/packages/lib/src/components/Dragonpay/Dragonpay.tsx
@@ -25,10 +25,6 @@ export class DragonpayElement extends UIElement {
         };
     }
 
-    formatProps(props) {
-        return props;
-    }
-
     render() {
         return (
             <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext}>
--- a/packages/lib/src/components/Dragonpay/Dragonpay.tsx
+++ b/packages/lib/src/components/Dragonpay/Dragonpay.tsx
@@ -26,10 +26,7 @@ export class DragonpayElement extends UIElement {
     }
 
     formatProps(props) {
-        return {
-            ...props,
-            items: props.details && props.details.length ? (props.details.find(d => d.key === 'issuer') || {}).items : props.items
-        };
+        return props;
     }
 
     render() {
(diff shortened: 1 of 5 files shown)


packages/lib/src/components/Dragonpay/Dragonpay.tsx
```diff
@@ -26,10 +26,7 @@ export class DragonpayElement extends UIElement {
     }
 
     formatProps(props) {
-        return {
-            ...props,
-            items: props.details && props.details.length ? (props.details.find(d => d.key === 'issuer') || {}).items : props.items
-        };
+        return props;
     }
```
pabloai, line 30: We can delete this `formatProps`, right?

packages/lib/src/components/Dropin/elements/createElements.ts
```diff
@@ -20,8 +20,8 @@ const createElements = (components: PaymentMethod[] = [], props, componentsConfi
 
         let componentInstance = getComponent(component.type, componentProps);
 
-        // Fallback to redirect if payment method not available and no details are required
-        if (!componentInstance && !component.details) {
+        // Fallback to redirect if payment method not available
+        if (!componentInstance) {
             componentInstance = getComponent(FALLBACK_COMPONENT, componentProps);
         }
```
pabloai, line 26: Maybe we could update `getComponent` to return the fallback when no component is found instead of calling it twice.

packages/lib/src/components/helpers/IssuerListContainer.tsx
```diff
@@ -44,18 +43,12 @@ class IssuerListContainer extends UIElement<IssuerListProps> {
...
-     * Formats props on construction time
-     */
-    formatProps(props) {
-        return {
-            ...props,
-            items: props.details && props.details.length ? (props.details.find(d => d.key === 'issuer') || {}).items : props.items
-        };
+    protected formatProps(props) {
+        return props;
     }
```
pabloai, line 52: Same with this one, we could just delete it.

pabloai (approved):

PR #454, opened 2020-10-28
Change Apple Pay DOM element to button, add lang parameter

marcperez:
## Summary
- Apple Pay element changed from a `<div>` element to a `<button>` element
- Language parameter added (See [Localizing Apple Pay Buttons](https://developer.apple.com/documentation/apple_pay_on_the_web/localizing_apple_pay_buttons))

### DIFF
--- a/packages/lib/src/components/ApplePay/components/ApplePayButton.tsx
+++ b/packages/lib/src/components/ApplePay/components/ApplePayButton.tsx
@@ -22,6 +22,8 @@ class ApplePayButton extends Component<ApplePayButtonProps> {
         /* eslint-disable jsx-a11y/no-static-element-interactions */
         return (
             <button
+                type="button"
+                aria-label={this.props.i18n.get('payButton')}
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/ApplePay/components/ApplePayButton.tsx
```diff
@@ -19,7 +21,8 @@ class ApplePayButton extends Component<ApplePayButtonProps> {
     render({ buttonColor, buttonType }) {
         /* eslint-disable jsx-a11y/no-static-element-interactions */
         return (
-            <div
+            <button
```
pabloai, line 24: It might be a good idea to add a `type="button"` to this and maybe an `aria-label` as well.

marcperez: Good idea, I added those two attributes 👍

pabloai (approved):

PR #478, opened 2020-11-05
Refactor clientKey usage and services functionality

pabloai:
## Summary
Update how we submit the `clientKey` as a search parameter.
Refactor services.

## Tested scenarios
Updated existing tests.


packages/lib/src/core/Services/fetch-json-data.ts
```diff
@@ -0,0 +1,66 @@
...
+            ...(clientKey && { 'X-Client-Key': clientKey }),
+            ...options.headers
+        },
+        redirect: 'follow',
+        referrerPolicy: 'no-referrer-when-downgrade',
+        ...(data && { body: JSON.stringify(data) })
+    } as RequestInit;
+
+    // TODO: Delete
+    // const url = `${loadingContext}${path}${clientKey ? `?token=${clientKey}` : ''}`;
```
marcperez, line 37: Is this intentionally left here?

pabloai: Yes, the idea is that these changes can be tested before the `x-api-key` header is supported.
I'll delete it before merging the branch.

marcperez (approved):

PR #483, opened 2020-11-06
Props order and drop-in refactor

marcperez:
## Summary
* Drop-in uses the new internal `_id` property internally.
* Drop-in elements are created using the global `core.create()` method.
* Allow only a set of global props that will be passed down to created components.
* Circular dependencies have been removed.


packages/lib/src/components/Dropin/Dropin.test.ts
```diff
@@ -1,5 +1,6 @@
 import { mount } from 'enzyme';
 import Dropin from './Dropin';
+import AdyenCheckout from '../../core';
```
pabloai, line 3: Do we need this import?

marcperez: Not at all, good catch!

pabloai (approved):

PR #489, opened 2020-11-09
Google pay onClick event

marcperez:
## Summary
Add an `onClick` event to the Google Pay component. It will be triggered when the button is clicked, and gives the merchant an opportunity to perform a validation.

```
onClick: (resolve, reject) => {
  // validate some external form
  resolve(); // or reject();
}

### DIFF
--- a/packages/lib/src/components/GooglePay/GooglePay.tsx
+++ b/packages/lib/src/components/GooglePay/GooglePay.tsx
@@ -43,7 +43,7 @@ class GooglePay extends UIElement<GooglePayProps> {
     public loadPayment = () => {
         const { onSubmit = () => {}, onAuthorized = () => {} } = this.props;
 
-        return new Promise((resolve, reject) => (this.props.onClick ? this.props?.onClick(resolve, reject) : resolve()))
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: packages/lib/src/components/GooglePay/GooglePay.tsx [hunk 1]
```diff
@@ -43,8 +43,8 @@ class GooglePay extends UIElement<GooglePayProps> {
     public loadPayment = () => {
         const { onSubmit = () => {}, onAuthorized = () => {} } = this.props;
 
-        return this.googlePay
-            .initiatePayment(this.props)
+        return new Promise((resolve, reject) => (this.props.onClick ? this.props?.onClick(resolve, reject) : resolve()))
+            .then(() => this.googlePay.initiatePayment(this.props))
```
pabloai, line 47: Another possibility would be to add the `onClick` to the default props making it just call the resolve function. Then can just call it without checking if it's defined or not.

marcperez: 👍

pabloai (approved):

PR #507, opened 2020-11-12
Improve Gift card error messages

marcperez:
## Summary
Handle more error types on the balance check step in the new Gift card flow.

## Tested scenarios
- An error is shown when a gift card has no balance
- An error is shown when a gift card has balance in a different currency
- An error is shown when a gift card number can't be found

### DIFF
--- a/packages/lib/src/components/Giftcard/Giftcard.tsx
+++ b/packages/lib/src/components/Giftcard/Giftcard.tsx
@@ -20,7 +20,8 @@ export class GiftcardElement extends UIElement {
             paymentMethod: {
                 type: this.constructor['type'],
                 brand: this.props.brand,
-                ...this.state.data
+                encryptedCardNumber: this.state.data?.encryptedCardNumber,
+                encryptedSecurityCode: this.state.data?.encryptedSecurityCode
             }
         };
     }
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -51,8 +52,9 @@ export class GiftcardElement extends UIElement {
             this.props.onBalanceCheck(resolve, reject, this.formatData());
         })
             .then(({ balance }) => {
-                if (!balance) throw new Error('Invalid balance');
-                if (balance?.currency !== this.props.amount?.currency) throw new Error('Unsupported balance currency');
+                if (!balance) throw new Error('card-error'); // card doesn't exist
+                if (balance?.currency !== this.props.amount?.currency) throw new Error('currency-error');
```
pabloai, line 56: Shouldn't `getCardErrorMessage` be used here? Otherwise you're just throwing `'currency-error'`, if I understand correctly.

marcperez: This `throw` will be catched by line `69`, then it will use this message to call `setStatus` on the component. That will trigger a setState (and a rerender), and it will display the message where applicable.

pabloai: Ah, I see 👍

pabloai (approved):

PR #511, opened 2020-11-13
Feature/e2e testing

sponglord:
## Summary
Added "bare bones" framework for running e2e tests in TestCafe.
Added comprehensive set of tests for Korean local cards (KCP)

**Fixed issue**:  Tests will prevent reoccurrence of issues mentioned in FOC-37220

## Usage

⚠️ The `chrome:headless` browser is required. More information can be found [here](https://devexpress.github.io/testcafe/documentation/guides/concepts/browsers.html#test-in-headless-mode).

```bash
yarn test:e2e
```


packages/lib/src/components/internal/SecuredFields/lib/utilities/dom.ts
```diff
@@ -62,4 +62,14 @@ const removeAllChildren = node => {
...
-export { getAttribute, on, off, select, selectOne, removeAllChildren };
+const getSearchParameters = (search = window.location.search) =>
+    search
+        .replace(/\?/g, '')
+        .split('&')
+        .reduce((acc, cur) => {
+            const [key, prop = ''] = cur.split('=');
+            acc[key] = decodeURIComponent(prop);
+            return acc;
+        }, {}) as any;
```
pabloai, line 73: I think this `getSearchParameters` util is good enough for the playground, but for the actual library I'd use the `core-js/web/url-search-params` polyfill as we do on PBL.

packages/lib/src/components/internal/SecuredFields/lib/core/SecuredField.ts
```diff
@@ -78,9 +78,15 @@ class SecuredField extends AbstractSecuredField {
...
 
         /**
-         * Create & reference iframe and add load listener
+         * Configure, create & reference iframe and add load listener
          */
-        const iframeEl: HTMLIFrameElement = createIframe(`${this.iframeSrc}`, processedAriaConfig[this.fieldType].iframeTitle);
+        const iframeConfig = {
+            src: this.iframeSrc,
+            title: processedAriaConfig[this.fieldType].iframeTitle,
+            policy: getSearchParameters().testing === 'testcafe' ? 'no-referrer-when-downgrade' : 'origin'
```
marcperez, line 86: What would you think of checking the `NODE_ENV` variable in this case, something like `process.env.NODE_ENV === 'test'?

packages/playground/tests/e2e/cards/kcpUtils.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const fillCVC = async (t, value = TEST_CVC_VALUE, replace = false) => {
+    return fillIFrame(t, 2, '#encryptedSecurityCode', value, replace);
+};
+
+export const fillDateAndCVC = async (t, dateValue = TEST_DATE_VALUE, cvcValue = TEST_CVC_VALUE) => {
+    await fillDate(t, dateValue);
+    return fillCVC(t, cvcValue);
+};
```
marcperez, line 37: Nice helper functions! Should we maybe move them into some more generic `cardUtils.ts` file, since they are not exclusive to Korean cards?

packages/lib/src/components/internal/SecuredFields/lib/core/SecuredField.ts
sponglord: I don't know if this would work or not, since we are running the tests from the playground. So where do you set the NODE_ENV? Setting it in the package.json of playground doesn't seem to make any difference. We'd have to test.
And I don't know if this would work for the plugins team either

packages/playground/tests/e2e/cards/kcpUtils.js
sponglord: But they kind of are exclusive as I say in the comment. 
Only KCP knows of an iframe index 3 that is a password field.
Likewise a giftcard would have a different index for its securityCode field than a regular card.
And a stored card has a different kind of field at index 1 than a regular card.
Maybe some of it can be abstracted but I see this as a work in progress and this kind of thing can be done later, as we all work on it

packages/lib/src/components/internal/SecuredFields/lib/utilities/dom.ts
sponglord: If you insist... ;-)
I've made the change

packages/playground/src/pages/Dropin/Dropin.js
```diff
@@ -9,13 +9,17 @@ import '../../style.scss';
 const initCheckout = async () => {
     const paymentMethodsResponse = await getPaymentMethods({ amount, shopperLocale });
 
+    //    paymentMethodsResponse.paymentMethods[0].brands.push('elo');
```
marcperez, line 12: Can we clean up the changes on this file?

packages/playground/src/pages/Cards/Cards.js
```diff
@@ -5,20 +5,24 @@ import { handleSubmit, handleAdditionalDetails, handleError } from '../../handle
...
     .then(paymentMethodsResponse => {
+        //getPaymentMethods({ amount, shopperLocale }).then(paymentMethodsResponse => {
         window.checkout = new AdyenCheckout({
             amount, // Optional. Used to display the amount in the Pay Button.
-            originKey,
+            //            originKey,
             clientKey: process.env.__CLIENT_KEY__,
             paymentMethodsResponse,
             locale: shopperLocale,
+            ...(getSearchParameters().countryCode?.toLowerCase() === 'kr' && { countryCode: 'KR' }),
```
marcperez, line 23: this could be replaced by importing `countryCode` from `''../../config/commonConfig'`

packages/lib/src/components/internal/SecuredFields/lib/core/SecuredField.ts
sponglord: Reverted this change since adding a <meta> tag fixes the issue for plugins - and always using `'origin'` works as long as you run components against Test (c.f. localhost)

packages/playground/src/pages/Cards/Cards.js
sponglord: Being able to do this allows the test to run independently of how the playground is set up - meaning the test can always work without having to "reconfigure" the playground every time.
And it is only a playground, right?

packages/playground/src/pages/Dropin/Dropin.js
sponglord: ok

sponglord: ok

packages/playground/src/pages/Cards/Cards.js
sponglord: ok

sponglord: And do we want to always include `countryCode` in the config, which importing it from `commonConfig` will do?

marcperez: That should be all right 👍

sponglord: ok, made the change

packages/lib/src/components/internal/SecuredFields/lib/utilities/dom.ts
marcperez, line 73: I think this method is not used anymore, right? If that's the case, can we remove it?

marcperez (approved):

PR #518, opened 2020-11-17
Feature/Update address component

pabloai:
## Summary
- Update the address component to be able to display different formats and labels for different countries.
- The order of the address fields had change, having the Country field first.

## Tested scenarios
Tested with addresses from BR, CA, GB, NL and US.

**Fixed issue**:  FOC-31982


packages/lib/src/components/internal/Address/Address.tsx
```diff
@@ -78,79 +73,34 @@ export default function Address(props: AddressProps) {
...
-
-            {requiredFields.includes('stateOrProvince') && (
-                <StateField
-                    value={data.stateOrProvince}
-                    errorMessage={!!errors.stateOrProvince}
-                    country={data.country}
-                    onDropdownChange={handleStateChange}
-                />
-            )}
+            {addressSchema.map(field => (field instanceof Array ? getWrapper(field) : getComponent(field, {})))}
```
marcperez, line 103: nice 👍

packages/lib/src/components/internal/Address/constants.ts
```diff
@@ -1,2 +1,47 @@
...
+const [STREET, HOUSE_NUMBER, POSTAL_CODE, CITY, STATE, COUNTRY] = ADDRESS_SCHEMA;
 export const COUNTRIES_WITH_STATES_DATASET = ['BR', 'CA', 'US'];
+
+// prettier-ignore
+export const ADDRESS_SCHEMAS = {
+    CA: [COUNTRY, HOUSE_NUMBER, STREET, [[CITY, 70], [POSTAL_CODE, 30]], STATE],
+    GB: [COUNTRY, [[HOUSE_NUMBER, 30], [STREET, 70]], [[CITY, 70], [POSTAL_CODE, 30]], STATE],
+    US: [COUNTRY, STREET, HOUSE_NUMBER, CITY, [[STATE, 50], [POSTAL_CODE, 50]]],
+    default: [COUNTRY, [[STREET, 70], [HOUSE_NUMBER, 30]], [[POSTAL_CODE, 30], [CITY, 70]], STATE]
+};
```
marcperez, line 11: Maybe we should consider adding and interface here

marcperez (approved):

packages/lib/src/components/internal/Address/constants.ts
pabloai: Good idea!

PR #531, opened 2020-11-19
End to end tests package

marcperez:
## Summary
- Move e2e tests to a new e2e package
- A new dev server is used for this, so we don't depend on the Playground.
- All tests are run automatically (matching any file with the suffix `.test`)


sponglord:
Good job!
A couple of observations:
- I don't think we need two `e2e` directories do we? A top level one is fine then everything else can just sit in the `tests` dir
- I don't see anyway to feed clientScripts defined config objects into the Dropin
- Is all this still running on port 3020? Should it run on another port? Will it clash if we run the tests whilst continuing development in the playground at port 3020?

marcperez:
> Good job!
> Some observations:
> 
> * I don't think we need two `e2e` directories do we? A top level one is fine then everything else can just sit in the `tests` dir
> * I don't see anyway to feed clientScripts defined config objects into the Dropin
> * Is all this still running on port 3020? Should it run on another port? Will it clash if we run the tests whilst continuing development in the playground at port 3020?
> * We could probably drop the `startWithKCP` and `startWithoutKCP` directories and just have them all in the `kcp` directory. The file naming makes it pretty clear what the test does and who the clientScripts is for

Thanks for the feedback @sponglord, I'll make the changes! 👍

sponglord (approved):

PR #570, opened 2020-12-03
Add internal useForm hook

marcperez:
## Summary

> Internal change, should not affect the input/output of any of these components

As of now, every component handles forms in its own way. This means a lot of duplication of code.
This PR introduces a custom hook `useForm`, which extracts common form logic to a common place. This should remove the need of repeating multiple states per component (valid, errors, data). It also handles events and validation.

### Highlights of `useForm`

Supported options:
* `schema`: Every form has a schema (that can change via `setSchema`, if needed). This is used to keep track of the fields of a form, and to prefill other state objects.
* `defaultData`: Used to prefill data in the state.
* `rules`: Set of rules used in the Validator class.
* `formatters`: Set of functions that will run for each field, formatting its value before validating it and storing it in state.

State/methods:
* `data`: Holds the actual values of each form field defined in the `schema`.
* `valid`/`errors`: Hold the validation state of each form field in the `schema`.
* `handleChangeFor(fieldName)`: Will run validations and set the data for a specific form field.
* `triggerValidation`: Revalidates every form field.

### Example

```ts
    const { handleChangeFor, triggerValidation, data, valid, errors, isValid } = useForm({
        schema: ['field1', 'field2'],
        defaultData: { field1: 'Field 1' },
        rules: { field1: (value) => { /* ... */ } },
    });
```

## Roadmap

- [x] IssuerLists migrated to `useForm`
- [x] Blik migrated to `useForm`
- [x] Boleto Bancario migrated to `useForm`
- [x] Dragonpay migrated to `useForm`
- [x] Econtext migrated to `useForm`
- [x] Doku migrated to `useForm`
- [x] MBWay migrated to `useForm`
- [x] Bank Transfer migrated to `useForm`
- [x] Bacs Direct Debit migrated to `useForm`


sponglord:
Does useForm always generate errors on 'blur'? Does this mean that once a field is clicked into - clicking out of it will always generate an error, even if the field is empty?

marcperez:
Hey @sponglord, thanks for your comments.

> * Does useForm always generate errors on 'blur'? Does this mean that once a field is clicked into - clicking out of it will always generate an error, even if the field is empty?

The `useForm` hook doesn't enforce any of this, it all depends on how changes are emitted from the component using it (`handleChangeFor`), and the validation rules for each of the modes (`input`/`blur`). In the components we've converted on this PR, the `blur` validation will happen only when the field has been edited.

> * I see (and like) that you can format a value as you validate it (in `processField`) - but there was that addition that the [Dev from Singapore made](https://github.com/Adyen/adyen-web/pull/202) about whether return an `errorMessge` with the validation (I'm not sure what the use case was, exactly, so I don't know if this is something we might still need? Just wonderin'...)

That's a very good point. For now, the `errors` state is just a simple object with boolean values, but this could be extended further as we convert the relevant payment methods (the ones in #202). Will keep this in mind 👍

sponglord:
From this code in `useForm. updateFieldData `:
```
setErrors(prevErrors => ({ ...prevErrors, [key]: mode === 'blur' && !isFieldValid }));
```
It looks like errors are only set on "blur". 
Am I reading this correctly? What about if you want to generate on "input" for example when the shopper exceeds the expected number of characters in a field

packages/lib/src/utils/useForm.ts
```diff
@@ -0,0 +1,107 @@
...
+import Validator, { ValidatorRules } from './Validator/FormValidator';
+
+function useForm<DataState = { [key: string]: any }>(props: { rules?: ValidatorRules; [key: string]: any }) {
+    const { rules = {}, formatters = {}, defaultData = {} } = props;
+    const validator = new Validator(rules);
+    const [schema, setSchema] = useState<string[]>(props.schema ?? []);
+    const [errors, setErrors] = useState<any>({});
+    const [valid, setValid] = useState<any>({});
+    const [data, setData] = useState<DataState>({} as DataState);
+    const isValid = useMemo(() => schema.reduce((acc, val) => acc && valid[val], true), [valid]);
```
pabloai, line 11: Small suggestion:
```suggestion
    const isValid = useMemo(() => schema.every(val => valid[val]), [valid]);
```

packages/lib/src/utils/useForm.ts
```diff
@@ -0,0 +1,107 @@
...
+    const processField = (key, value, mode) => {
+        const formattedValue = formatters[key] ? formatters[key](value ?? '') : value;
+        const validationResult = validator.validate(key, formattedValue, mode);
+        return [formattedValue, validationResult];
+    };
+
+    const reindexSchema = keys => {
+        const cleanupRemovedFields = (prevData, initialValue) =>
+            keys.reduce((acc, key) => {
+                acc[key] = prevData[key] !== undefined ? prevData[key] : initialValue;
```
pabloai, line 34: Would this work on this case? Or could it be that `prevData[key]` is `null` and we want to keep it that way?
```suggestion
                acc[key] = prevData[key] ?? initialValue;
```

marcperez: That's the thing, not sure how safe it would be to make that kind of assumption. That's why it's currently just setting it to the `initialValue` if it's not defined at all.

pabloai (approved):

Issue #589, opened 2020-12-09
ApplePay - onClick event


lluzak:
**Is your feature request related to a problem? Please describe.**
I need to intercept button click to like i can do with Google Pay.

**Describe the solution you'd like**
Exactly the same as Google Pay button.

**Describe alternatives you've considered**
Rendering button myself and adding logic myself.

marcperez:
Hi @lluzak,

Thanks for opening this issue. We were planning on to adding support for this, but we prioritized it. (See PR #592).
I'll close this ticket, feel free to reopen if you have any more comments or feedback on this.

Hope this helps!

PR #592, opened 2020-12-10
Add onClick event to Apple Pay Component

marcperez:
## Summary
Adds compatibility to the `onClick` event to the Apple Pay component.

```js
    checkout.create('applepay', {
      // ...
      onClick: (resolve, reject) => {
        resolve();
      }
    });
```

> Be aware that depending on what actions are performed in this event, an error could occur when creating the Apple Pay Session (`Must create a new ApplePaySession from a user gesture handler`, see [Creating an Apple Pay Session](https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_js_api/creating_an_apple_pay_session)). This means, for example, starting the session after an asynchronous call will fail in most cases. Refer to the Apple Pay documentation for more information on this.

**Fixed issue**:  #589

### DIFF
--- a/packages/playground/src/pages/Dropin/Dropin.js
+++ b/packages/playground/src/pages/Dropin/Dropin.js
@@ -99,6 +99,12 @@ const initCheckout = async () => {
                     }
                 }
             },
+            applepay: {
+                configuration: {
+                    merchantName: 'Adyen Test merchant',
+                    merchantIdentifier: '000000000200001'
+                }
+            },
             paywithgoogle: {
                 countryCode: 'NL',
                 onAuthorized: console.info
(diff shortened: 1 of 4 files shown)


packages/playground/src/pages/Dropin/Dropin.js
```diff
@@ -99,6 +99,12 @@ const initCheckout = async () => {
                     }
                 }
             },
+            applepay: {
+                configuration: {
+                    merchantName: 'Adyen Test merchant',
+                    merchantIdentifier: '000000000200001'
+                }
```
pabloai, line 106: Aren't we getting this in the `/paymentMethods` response?

pabloai (approved):

packages/playground/src/pages/Dropin/Dropin.js
marcperez: Not always, this is useful for testing (as we have on the Wallets page).

PR #615, opened 2020-12-18
Implement Apple Pay & Google Pay brands from payment methods response

shernaz:
## Summary
Implement apple pay and google pay brands for Pay By Link

## Tested scenarios
To be included in upcoming commits

### DIFF
--- a/packages/lib/src/components/GooglePay/utils.ts
+++ b/packages/lib/src/components/GooglePay/utils.ts
@@ -11,3 +11,16 @@ export function resolveEnvironment(env = 'TEST'): google.payments.api.Environmen
             return 'TEST';
     }
 }
+
+export function mapBrands(brands) {
+    const brandMapping = {
+        mc: 'MASTERCARD',
+        amex: 'AMEX',
+        visa: 'VISA',
+        interac: 'INTERAC',
+        discover: 'DISCOVER',
+        jcb: 'JCB'
+    };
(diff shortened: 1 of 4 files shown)


packages/lib/src/components/GooglePay/utils.ts
```diff
@@ -11,3 +11,15 @@ export function resolveEnvironment(env = 'TEST'): google.payments.api.Environmen
...
+export function mapBrands(brands) {
+    const brandMapping = {
+        mc: 'MASTERCARD',
+        amex: 'AMEX',
+        visa: 'VISA',
+        interac: 'INTERAC',
+        discover: 'DISCOVER'
+    };
+    const allowedCardNetworks = brands.map(brand => brandMapping[brand]);
+    return allowedCardNetworks.reduce((unique, item) => (unique.includes(item) ? unique : [...unique, item]), []);
```
pabloai, line 24: About the reducer to remove duplicated items, why would it be possible for items to be duplicated?

shernaz: yes, it is possible, because, we have elodebit and elo txvariants, however applepay has only txvariant named elo. Hence when we map this case, elo could be duplicated.

pabloai: So we could check for duplicated items only for Apple Pay then?

marcperez (approved):

PR #619, opened 2020-12-22
Add class to card containers showing their funding source

sponglord:
## Summary
Add a class to the top level card containers indicating the `fundingSource` (`"credit"`/`"debit"`) for those card payment methods.
For Dropin especially this will make it easy to identify and, if required, hide via css, a particular set of cards

## Tested scenarios
Tested separating card PMs based on fundingSource - both standalone card components and Dropin card components add an extra class to their top level container

**Fixed issue**:  COWEB-885

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.tsx
@@ -173,7 +173,7 @@ class CardInput extends Component<CardInputProps, CardInputState> {
                 render={({ setRootNode, setFocusOn }, sfpState) => (
                     <div
                         ref={setRootNode}
-                        className={`adyen-checkout__card-input ${styles['card-input__wrapper']} adyen-checkout__card-input-${this.props.fundingSource}`}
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -171,7 +171,10 @@ class CardInput extends Component<CardInputProps, CardInputState> {
                 type={this.props.brand}
                 oneClick={isOneClick}
                 render={({ setRootNode, setFocusOn }, sfpState) => (
-                    <div ref={setRootNode} className={`adyen-checkout__card-input ${styles['card-input__wrapper']}`}>
+                    <div
+                        ref={setRootNode}
+                        className={`adyen-checkout__card-input ${styles['card-input__wrapper']} adyen-checkout__card-input-${this.props.fundingSource}`}
```
pabloai, line 176:
```suggestion
                        className={`adyen-checkout__card-input ${styles['card-input__wrapper']} adyen-checkout__card-input--${this.props.fundingSource}`}
```

pabloai (approved):

PR #620, opened 2020-12-22
Change default challenge iframe sizing.

sponglord:
## Summary
After discussions in the youTrack tickets COWEB-867 & DEV-35803 it has been agreed to make the default 3DS2 challenge iframe size "02" (390px x 400px).
Also it was agreed that the setting "05" should be truly 100% w/h and so we should stop forcing a 16:9 ratio with padding for this setting.

## Tested scenarios
Challenge iframe is now 390px x 400px by default
If challenge iframe _is_ set to "05" the 16:9 aspect ratio is gone and the size is 100% x 100% (this required a height to be set on the top level containers in the playground Dropin and Cards so that the challenge iframe didn't disappear)

**Fixed issue**:  DEV-35803


packages/playground/src/style.scss
```diff
@@ -337,3 +337,11 @@ table {
         }
     }
 }
+
+#dropin-container{
```
marcperez, line 341:
```suggestion
#dropin-container {
```

packages/playground/src/style.scss
```diff
@@ -337,3 +337,11 @@ table {
...
+    min-height: 540px;
+}
+
+.card-field{
```
marcperez, line 345:
```suggestion
.card-field {
```

packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
```diff
@@ -31,7 +31,7 @@ const actionTypes = {
             paymentData: action.paymentData,
             onComplete: props.onAdditionalDetails,
             onError: props.onError,
-            size: '05',
+            size: '02',
```
marcperez, line 34: Just to make sure, can this be overriden via props?

sponglord: No. Just the way it can't be at present, either (unless you instantiate the component directly with `checkout.create('threeDS2Challenge')`)

packages/lib/src/components/ThreeDS2/ThreeDS2.scss
```diff
@@ -6,6 +6,8 @@
     position: relative;
     overflow: hidden;
     width: 100%;
+    min-height: 400px;
```
marcperez, line 9: Is there any reason to force this to `400px`?

packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
marcperez: Would it be possible to allow overriding this value also when not using the new flow?

sponglord: Why suddenly introduce that option when its _never_ been there before?
Why not just move to a new version if you want this config option?

packages/lib/src/components/ThreeDS2/ThreeDS2.scss
sponglord: Yes - 'cos if the merchant has never defined a height on any of the container elements the challenge window will not be visible

marcperez (approved):

PR #632, opened 2020-12-31
Fix Personal details output format

pabloai:
The personal details returns the output data in an outdated format, which is causing issues with Voucher components. This fixes this and aligns the functionality of the Company details component as well.

### DIFF
--- a/packages/lib/src/components/Doku/Doku.tsx
+++ b/packages/lib/src/components/Doku/Doku.tsx
@@ -17,9 +17,9 @@ export class DokuElement extends UIElement {
      */
     formatData() {
         return {
+            ...this.state.data,
             paymentMethod: {
-                type: this.props.type || DokuElement.type,
-                ...this.state.data
+                type: this.props.type || DokuElement.type
             }
         };
     }
--- a/packages/lib/src/components/Econtext/Econtext.tsx
+++ b/packages/lib/src/components/Econtext/Econtext.tsx
@@ -17,9 +17,9 @@ export class EcontextElement extends UIElement {
      */
     formatData() {
         return {
+            ...this.state.data,
             paymentMethod: {
-                type: this.props.type || EcontextElement.type,
-                ...this.state.data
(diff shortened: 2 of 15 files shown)


packages/playground/src/pages/OpenInvoices/OpenInvoices.js
```diff
@@ -56,9 +56,9 @@ getPaymentMethods({ amount, shopperLocale }).then(paymentMethodsData => {
         .create('affirm', {
             countryCode: 'US', // 'US' / 'CA'
             visibility: {
-                personalDetails: 'editable', // editable [default] / readOnly / hidden
-                billingAddress: 'editable',
-                deliveryAddress: 'editable'
+                personalDetails: 'hidden', // editable [default] / readOnly / hidden
+                billingAddress: 'hidden',
+                deliveryAddress: 'hidden'
```
marcperez, line 61: Hey @pabloai, are these changes intentional?

pabloai: They weren't, thanks for checking! I just pushed the rollback.

packages/lib/src/components/Doku/Doku.tsx
```diff
@@ -17,9 +17,9 @@ export class DokuElement extends UIElement {
      */
     formatData() {
         return {
+            ...this.state.data,
```
marcperez, line 20: Since we're refactoring this, could we change this to be more explicit on the fields that are returned?

packages/lib/src/components/Econtext/Econtext.tsx
```diff
@@ -17,9 +17,9 @@ export class EcontextElement extends UIElement {
      */
     formatData() {
         return {
+            ...this.state.data,
```
marcperez, line 20: Same as with Doku, can we make this more explicit?

marcperez (approved):

PR #669, opened 2021-01-14
PIX component

marcperez:
## Summary
Add the new PIX Component for Brazil.

### DIFF
--- a/packages/lib/src/components/Pix/Pix.ts
+++ b/packages/lib/src/components/Pix/Pix.ts
@@ -8,7 +8,7 @@ class PixElement extends QRLoaderContainer {
             delay: 2000, // ms
             countdownTime: 15, // min
             copyBtn: true,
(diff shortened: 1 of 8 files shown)


packages/lib/src/components/Pix/Pix.ts
```diff
@@ -0,0 +1,17 @@
...
+
+class PixElement extends QRLoaderContainer {
+    public static type = 'pix';
+
+    formatProps(props) {
+        return {
+            delay: 2000, // ms
+            countdownTime: 15, // min
+            copyBtn: true,
+            introduction: 'Abra o app com sua chave PIX cadastrada, escolha Pagar com Pix e escaneie o QR Code ou copie e cole o código',
```
pabloai, line 11: Do we need this here and in the translation file?

marcperez: Oops! Good catch!

pabloai (approved):

PR #681, opened 2021-01-19
Add new Arabic language

felipeloge:
## Summary
Add new Arabic language.

Code | Name
-- | --
ar-AE | Arabic (United Arab Emirates)


pabloai:
Hi @felipeloge,

We appreciate your collaboration but unfortunately we don't currently support right-to-left languages.

In the future if you'd like us to support another language you can just create an issue asking for it. We'll then check if it would benefit other merchants as well and if it does, we'll ask our third-party service to generate the translation files (so no need to generate the files yourself).

Also keep in mind that even if we don't support a language you can create your own [custom translations](https://docs.adyen.com/online-payments/drop-in-web/customization#creating-a-localization).

We'll research the best way to display and support right-to-left languages, but it might take a while. I'm closing the PR until then.

Sorry for the trouble, thanks for the effort and feel free to contribute again in the future!

PR #692, opened 2021-01-21
Feature/handling v2 bin lookup

sponglord:
## Summary
Changes needed to use and handle the v2 /binLookup

## Tested scenarios
Added tests.
All tests, old & new, unit & e2e, pass


packages/e2e/tests/cards/binLookup/binLookup.test.js
```diff
@@ -30,7 +30,7 @@ fixture`Testing binLookup response`
     .clientScripts('binLookup.clientScripts.js')
     .requestHooks(logger);
 
-test('Enter number of known dual branded card, ' + 'then inspect body for expected properties ', async t => {
+test.skip('Enter number of known dual branded card, ' + 'then inspect body for expected properties ', async t => {
```
marcperez, line 33: Is this accidental?

sponglord: Nope. binLookup.test.js is for v1. We now have binLookup.v2.test.js. 
Once all is confirmed and we know we're not switching back to v1 then the test file can go

packages/lib/src/components/Card/Card.tsx
```diff
@@ -4,7 +4,7 @@ import CardInput from './components/CardInput';
 import CoreProvider from '../../core/Context/CoreProvider';
 import getImage from '../../utils/get-image';
 import collectBrowserInfo from '../../utils/browserInfo';
-import { CardElementData, CardElementProps, BinLookupResponseObj } from './types';
+import { CardElementData, CardElementProps, BinLookupObject } from './types';
```
marcperez, line 7: Just a suggestion, maybe `BinLookupResponse` is a bit more descriptive for this type.

sponglord: ok

marcperez (approved):

PR #707, opened 2021-01-29
Feature/using cvc policy

sponglord:
## Summary
- In light of the switch to the new v2 `/binLookup` and it's use of a `cvcPolicy` - refactor `Card` & `SecuredField` code to drop the separate `cvcRequired` & `hideCVC` properties and instead use a `cvcPolicy` prop which can cover all scenarios.
- Also implement a new procedure for `CardInput` to trigger validation - validating "every time there's a change in state" rather than "validating every time we set the state"
- Set SF version back to `v3.2.8`

## Tested scenarios
All tests, unit & e2e, run and pass


packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -87,10 +89,23 @@ class CardInput extends Component<CardInputProps, CardInputState> {
...
+            !objectsDeepEqual(prevState.installments, this.state.installments) ||
+            prevState.isSfpValid !== this.state.isSfpValid ||
+            prevState.hideCVCForBrand !== this.state.hideCVCForBrand ||
+            prevState.brand !== this.state.brand ||
+            prevState.additionalSelectValue !== this.state.additionalSelectValue ||
+            // Covers changes to:
+            // - encryptedField values
+            // - KCP authentication
+            // - holder name
+            !objectsDeepEqual(prevState.data, this.state.data)
```
marcperez, line 107: Maybe this could be extracted into a function

sponglord: Why?

sponglord: I can - if it will make you happy ;-) What shall I call it. `checkStateToSeeWhetherToValidateCardInput` :-)

packages/lib/src/components/internal/SecuredFields/lib/utilities/commonUtils.ts
```diff
@@ -88,6 +100,29 @@ function isObjectLike(value) {
...
+    const keys2 = Object.keys(obj2);
+
+    if (keys1.length !== keys2.length) {
+        return false;
+    }
+
+    for (const key of keys1) {
+        if (isObjectLike(obj1[key])) {
+            if (!objectsDeepEqual(obj1[key], obj2[key])) {
+                // console.log('### commonUtils::objectsDeepEqual:: mismatch on key:', key);
```
marcperez, line 114: Can we remove this?

packages/lib/src/components/Card/components/CardInput/CardInput.tsx
marcperez: Just thinking it could maybe make the `CardInput.tsx` file a bit more readable, as you prefer! 👍

marcperez (approved):

packages/lib/src/components/internal/SecuredFields/lib/utilities/commonUtils.ts
sponglord: sure!

PR #730, opened 2021-02-08
Release 3.22.0

sponglord:
## Fixes
Filter payment methods which are not supported in the Drop-in (#722). 

## Changes
In preparation for a future release, we've made some updates to the 3D Secure flow. This should not have any effect on your integration (#729)

## Required API version
Adyen Web v3.22.0 requires API v66 or later.


marcperez:
We also need to update the dependency in the `playground` and `e2e` packages

marcperez (approved):

PR #754, opened 2021-02-12
Prevent the PayPal submit method to be called programatically in the Drop-in

pabloai:
## Summary
Updated the PayPal component to prevent the `submit` method to be called when used in the component.
The `onError` will be triggered when trying to call the `submit` method in both the Drop-in and the component.

## Tested scenarios
Tested in the Drop-in and in the standalone component.

### DIFF
--- a/packages/lib/src/components/PayPal/Paypal.test.ts
+++ b/packages/lib/src/components/PayPal/Paypal.test.ts
@@ -11,11 +11,11 @@ describe('Paypal', () => {
         expect(paypal.isValid).toBe(true);
     });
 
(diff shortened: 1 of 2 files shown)


marcperez (approved):

TarasVerbniak:
Hi!
What's the reason of these changes?
You broke our logic by this PR.
In our website we're using one submit button for all payment methods and now we can't use it or should look for dirty "workarounds".
According to your documentation - there is possibility to disable submit buttons for each method by putting `showPayButton: false` and then use just one submit button wherever you wish on page and programatically submit payment form.
Would be great if you can review your changes one more time.
Thanks!

PR #756, opened 2021-02-12
Fix phone input labels

marcperez:
## Summary
Changes the phone field labels to "Mobile number" on the Qiwi Wallet and MBWay components.


packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -40,7 +40,7 @@ function MBWayInput(props: MBWayInputProps) {
         <div className="adyen-checkout__mb-way">
             <Field
                 errorMessage={!!errors.telephoneNumber && i18n.get('telephoneNumber.invalid')}
```
pabloai, line 42: We should use `mobileNumber.invalid` instead of `telephoneNumber.invalid`.
Could you double check if we have that string in our translation files? If not, you'll need to add it in the base strings file and generate the translation files again.

pabloai (changes requested):

pabloai (approved):

PR #767, opened 2021-02-17
useForm improvements

marcperez:
## Summary
- Improve default data handling
- Refactors Validator class
- Migrates Address internal component to `useForm`


sponglord:
So is there still a FormValidator module or has all the functionality moved into Validator?

marcperez:
> So is there still a FormValidator module or has all the functionality moved into Validator?

FormValidator was a transitional name, it has now been moved to Validator

sponglord (approved):

Issue #725, opened 2021-02-05
Billing address country select


bitbytebit1:
**Steps to reproduce**  
1. Under the billing address click "Choose Country"
2. Type "swe"

**Expected Behavior**  
The select should highlight "Sweden"

**Actual Behavior**  
The select does not highlight Sweden

**Possible fixes**  
Use a select element instead of a button and ul (this will also look better on mobile in my opinion)

**Other comments**
Users are also not able to auto fill the form with their address. All other inputs are auto filled but not country

marcperez:
Hi @bitbytebit1,

Thanks for your feedback, sounds like a good suggestion 👍 

We're currently working on improving the usability of our dropdowns by adding a search functionality, which should solve your use case. We will also take into consideration the suggestion on using a select element on mobile.

We'll update this ticket once that is ready. Hope this helps!

marcperez:
Hi @bitbytebit1,

We were able to prioritize this feature (#776) and it will be released on our next major release (v4.0.0). I'll close this ticket, but feel free to reach us if there's any other comments or feedback you'd like to share. Thanks again!

PR #776, opened 2021-02-19
Select component updates

pabloai:
## Summary
- Make it possible to filter options in the Select component
- Update component to use hooks.
- Filterable lists will be the new default behaviour.
- Added strings:
```
"select.noOptionsFound": "No options found"
"select.filter.placeholder": "Search..."
```

## Tested scenarios
Both filterable and non filterable lists.

**Fixed issue**:  Adyen/adyen-web#725

### DIFF
--- a/packages/lib/src/components/internal/FormFields/Select/Select.tsx
+++ b/packages/lib/src/components/internal/FormFields/Select/Select.tsx
@@ -4,6 +4,7 @@ import cx from 'classnames';
 import SelectButton from './components/SelectButton';
 import SelectList from './components/SelectList';
 import uuid from '../../../../utils/uuid';
+import { keys } from './constants';
 import { SelectItem, SelectProps } from './types';
 import styles from './Select.module.scss';
 import './Select.scss';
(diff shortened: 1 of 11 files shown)


packages/lib/src/language/locales/en-US.json
```diff
@@ -135,6 +135,8 @@
     "select.stateOrProvince": "Select state or province",
     "select.provinceOrTerritory": "Select province or territory",
     "select.country": "Select country",
+    "select.noOptionsFound": "No options found",
+    "select.filter.placeholder": "Search...",
```
marcperez, line 139: Reminder: these need to be added into the translations repository

packages/lib/src/components/internal/FormFields/Select/Select.tsx
```diff
@@ -1,217 +1,201 @@
...
+
+    const active: SelectItem = props.items.find(i => i.id === props.selected) || ({} as SelectItem);
+
+    const keys = {
+        arrowDown: 'ArrowDown',
+        arrowUp: 'ArrowUp',
+        enter: 'Enter',
+        escape: 'Escape',
+        space: ' '
     };
```
marcperez, line 29: This could also be declared outside of this component, and imported here.

marcperez (approved):

packages/lib/src/components/internal/FormFields/Select/Select.tsx
pabloai: Done.

packages/lib/src/language/locales/en-US.json
pabloai: I've created the translation job.

marcperez (approved):

PR #786, opened 2021-02-23
Delay conversion id creation

marcperez:
## Summary
Delays fetching the `conversionId` on the Analytics module until the first telemetry call.

## Tested scenarios
- Creating a new `AdyenCheckout` instance does not call the `conversionId` endpoint
- Mounting a component prompts a `conversionId` call.
- If called multiple times, conversionId is only fetched once.

### DIFF
--- a/packages/lib/src/core/Analytics/Analytics.ts
+++ b/packages/lib/src/core/Analytics/Analytics.ts
@@ -17,28 +17,20 @@ class Analytics {
     private readonly logEvent;
     private readonly logTelemetry;
     private readonly queue = new EventsQueue();
+    public readonly collectId;
 
     constructor({ loadingContext, locale, clientKey, analytics }: CoreOptions) {
         this.props = { ...Analytics.defaultProps, ...analytics };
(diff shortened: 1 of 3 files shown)


packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -17,28 +17,20 @@ class Analytics {
...
     constructor({ loadingContext, locale, clientKey, analytics }: CoreOptions) {
         this.props = { ...Analytics.defaultProps, ...analytics };
         this.logEvent = logEvent({ loadingContext, locale });
         this.logTelemetry = postTelemetry({ loadingContext, locale, clientKey });
+        this.collectId = collectId({ loadingContext, clientKey });
 
         const { conversion, enabled } = this.props;
-
-        if (conversion && enabled) {
+        if (conversion === true && enabled === true) {
```
pabloai, line 29: Which other values could `conversion` and `enabled` have?

pabloai (approved):

PR #787, opened 2021-02-23
Fix Gift card pin field error outline

marcperez:
## Summary
Fix Gift card pin field error outline


sponglord:
This one is my fault - but while you've got a PR open, will you change line 82 to use `error.va.gen.01` (which means "incomplete field") rather than `error.va.sf-cc-num.01` ( which means "failed luhn check").

sponglord (approved):

sponglord (approved):

PR #808, opened 2021-02-26
Differentiate card from custom-card

sponglord:
## Summary
Card component and custom card component now use a different static `type` property.
This will allow any analytics to differentiate between the two.

## Tested scenarios
Checked with relevant people in the analytics team that this will not affect existing dashboards and made sure they are aware of this change.
Reply:
> "There’s currently no logic taking this value explicitly into account, so it would just show up as another group by value in any of the dashboards"


marcperez (approved):
Looks good! 👍

PR #819, opened 2021-03-03
Address component updates

pabloai:
## Summary
- Create Address and PersonalDetails helper components
- Create Helpers section in the Playground
- Update useForm to provide a context object to be used on the Validator.
- Create address specifications object and make it possible to receive specifications as a prop.
- This would make it possible to use the Address component outside payment methods components

## Tested scenarios
- Added and updated tests.
- Tested Address and PersonalDetails as standalone helper components.


packages/lib/src/components/Address/Address.tsx
```diff
@@ -0,0 +1,30 @@
+import { h } from 'preact';
+import UIElement from '../UIElement';
+import Address from '../internal/Address';
+import CoreProvider from '../../core/Context/CoreProvider';
+
+export class AddressElement extends UIElement {
+    get data() {
+        return { ...this.state.data };
```
marcperez, line 8: Is this intentional?

pabloai: No, good catch!

marcperez (approved):

Issue #807, opened 2021-02-26
Can't resolve 'promise/lib/es6-extensions.js' in V4.0.0


nachogarcia:
**Describe the bug**
After trying to update to v4.0.0, the project build results in an error.

```
ERROR in ./node_modules/@adyen/adyen-web/dist/adyen.js
   Module not found: Error: Can't resolve 'promise/lib/es6-extensions.js' in 'node_modules/@adyen/adyen-web/dist'
   @ ./node_modules/@adyen/adyen-web/dist/adyen.js 1:10264-10304
```

We're importing the library with codesplit so we don't compromise our performance, and the CSS as indicated
` const AdyenCheckout = await import('@adyen/adyen-web' /* webpackChunkName: 'adyen' */).then(m => m.default)`
`import '@adyen/adyen-web/dist/adyen.css'`

As a framework we're using nuxt with modern build in client side, babel with @nuxt/babel-preset-app and core.js v3 with proposals.

marcperez:
Hi @nachogarcia,

Thanks for reaching! We haven't been able to reproduce this issue, just to discard an issue with the new ES output and the specific Webpack configuration you're using, could you try the following?

```js
const AdyenCheckout = await import('@adyen/adyen-web/dist/adyen.js' /* webpackChunkName: 'adyen' */).then(m => m.default)`
```
This should load the UMD export (very similar to the one present in `v3.x`). Let us know if this works. Hope this helps!

nachogarcia:
With both imports, the client build works fine; but when it goes to create the modern build (using modules) it breaks.

Promise is set as devDependency not a dependency, maybe it helps. By the way - any reason not to use core-js promise?

nachogarcia:
It works now 👍 

Before
![image](https://user-images.githubusercontent.com/9975630/110505201-3722ad00-80fe-11eb-95ee-a4c6b1d4487b.png)

After
![image](https://user-images.githubusercontent.com/9975630/110505170-2eca7200-80fe-11eb-96fe-293ad3983be0.png)

After, importing directly from '@adyen/adyen-web/dist/es'
![image](https://user-images.githubusercontent.com/9975630/110506212-2e7ea680-80ff-11eb-9487-84f2b80ddd70.png)

Looks like it's not picking the es build. It does with other packages, but it might do with default webpack configs.

PR #822, opened 2021-03-04
Change promise polyfill

nachogarcia:
## Summary
Changes require to import, and uses core-js promise polyfill for it. 

## Tested scenarios
Haven't been able to test in a local project nor analyze bundle size diffs yet

**Fixed issue**:  https://github.com/Adyen/adyen-web/issues/807

### DIFF
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -31,7 +31,6 @@
         "html-webpack-plugin": "^4.5.1",
         "node-sass": "^4.12.0",
         "prettier": "^1.19.1",
-        "promise": "^8.1.0",
         "sass-loader": "^10.1.1",
         "source-map-loader": "^1.1.3",
         "style-loader": "^2.0.0",
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -70,7 +70,6 @@
         "postcss": "^8.2.6",
         "postcss-reporter": "^7.0.2",
         "prettier": "^1.19.1",
-        "promise": "^8.1.0",
         "rollup": "^2.40.0",
         "rollup-plugin-postcss": "^4.0.0",
         "rollup-plugin-terser": "^7.0.2",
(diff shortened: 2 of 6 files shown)


marcperez:
Hi @nachogarcia,

Thanks for opening this PR! This is really helpful 👍 

I noticed the bundle size increased a bit after switching to the core-js promise polyfill though:

| Filename | Size | Change |  |
| :--- | :---: | :---: | :---: |
| `packages/lib/dist/adyen.js` | 151 kB | +5.07 kB (+3%) |  |
| `packages/lib/dist/cjs/index.js` | 151 kB | +5.18 kB (+4%) |  |
| `packages/lib/dist/es/index.js` | 90.6 kB | +5.06 kB (+6%) | 🔍 |

Maybe we should stick to the previous polyfill, and move it from `devDependencies` to `dependencies` to solve the issue in #807. Any thoughts?

nachogarcia:
Other core-js polyfills are always being loaded.
The increase in bundle size is relative, since many people already use core-js polyfills so the promise polyfill wouldn't be duplicated.

I think the best course of action would be to use different targets, so es builds are left unpolyfilled and given the choice to do so.

If you think it's better for now to keep the polyfills as they are, feel free to close the MR and change the dependency.

marcperez:
> Other core-js polyfills are always being loaded.
> The increase in bundle size is relative, since many people already use core-js polyfills so the promise polyfill wouldn't be duplicated.
> 
> I think the best course of action would be to use different targets, so es builds are left unpolyfilled and given the choice to do so.
> 
> If you think it's better for now to keep the polyfills as they are, feel free to close the MR and change the dependency.

In the future we're planning on loading polyfills automatically and as externals, but for now I think it'd make sense indeed to stick to the current polyfill (not to add those extra kb in the bundle). I'll close this PR then and submit a different one with the dependency change.

Thanks again for your contribution, your feedback has been very helpful!

PR #824, opened 2021-03-04
Feature/translated cvc placeholders

sponglord:
## Summary
The `placeholders` for SecuredFields have gone the way of the `ariaLabels` and become something that is only read from the translation files. These translations have 3 digit & 4 digit variations for the CVC field
If the merchant wants to customise them they do so via a translations config object e.g.
```
translations: {
  'en-US': {
    'creditCard.numberField.placeholder': '1111 2222 3333 4444',
    'creditCard.expiryDateField.placeholder': 'mm/yy',
    'creditCard.cvcField.placeholder.3digits': '456',
    'creditCard.cvcField.placeholder.4digits': '4567'
  }
}
```
This also bumps the SecuredFields version to `3.4.0`

## Tested scenarios
Update existing unit tests.
All tests, unit & e2e, pass.
Placeholders are read from the translation files for both Card & Custom Card components
Placeholders can be overwritten, including being set to an empty string, via a translations config object

**Relates to issue**:  COWEB-708

### DIFF
--- a/packages/lib/src/components/Giftcard/Giftcard.tsx
+++ b/packages/lib/src/components/Giftcard/Giftcard.tsx
@@ -4,10 +4,9 @@ import GiftcardComponent from './components/GiftcardComponent';
 import CoreProvider from '../../core/Context/CoreProvider';
 import getImage from '../../utils/get-image';
 import PayButton from '../internal/PayButton';
-import { GIFT_CARD } from '../internal/SecuredFields/lib/configuration/constants';
 
 export class GiftcardElement extends UIElement {
-    public static type = GIFT_CARD;
+    public static type = 'giftcard';
 
     formatProps(props) {
         return props;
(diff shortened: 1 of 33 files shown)


packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -4,9 +4,10 @@ import GiftcardComponent from './components/GiftcardComponent';
 import CoreProvider from '../../core/Context/CoreProvider';
 import getImage from '../../utils/get-image';
 import PayButton from '../internal/PayButton';
+import { GIFT_CARD } from '../internal/SecuredFields/lib/configuration/constants';
 
 export class GiftcardElement extends UIElement {
-    public static type = 'giftcard';
+    public static type = GIFT_CARD;
```
marcperez, line 10: Are we sure we want to use the Secured Fields type for the payment method type? They seem different things (even if they share the same value).

sponglord: I'm using an existing constant so we don't have loads of 'giftcard' strings lying around.
But I can change it back

marcperez: Let's change this one back but keep the others, so it won't break if at some point the other value changes. Rest looks good! 👍

marcperez (approved):

packages/lib/src/components/Giftcard/Giftcard.tsx
sponglord: ok

PR #839, opened 2021-03-08
Amazon APB flow

pabloai:
## Summary
- Add support to the Additional Payment Button flow.
- Make Amazon Pay available in the Drop-in.
- Update Wallets and Drop-in playground pages to support this supported.

### DIFF
--- a/packages/playground/src/pages/Dropin/Dropin.js
+++ b/packages/playground/src/pages/Dropin/Dropin.js
@@ -145,12 +145,12 @@ function handleFinalState(resultCode, dropin) {
 
 function handleRedirectResult() {
     const storedPaymentData = localStorage.getItem('storedPaymentData');
-    const { redirectResult, payload } = getSearchParameters(window.location.search);
+    const { amazonCheckoutSessionId, redirectResult, payload } = getSearchParameters(window.location.search);
 
-    if (storedPaymentData && (redirectResult || payload)) {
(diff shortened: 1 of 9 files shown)


packages/playground/src/pages/Dropin/Dropin.js
```diff
@@ -166,6 +166,23 @@ function handleRedirectResult() {
...
+    if (amazonCheckoutSessionId) {
+        window.amazonpay = checkout
+            .create('amazonpay', {
+                amazonCheckoutSessionId,
+                showOrderButton: false,
+                onSubmit: state => {
+                    makePayment(state.data).then(result => {
+                        handleFinalState(result.resultCode, dropin);
+                    });
+                }
```
marcperez, line 179: Is this intentional? Shouldn't it use the `onSubmit` event of the main instance?

pabloai: It is intentional. If I would use the one in the `onSubmit` in the main instance, it would try to call `component.setStatus` (with `component` being the Amazon Pay component), but I want to show the status in the Drop-in. In a real implementation this would most likely happen on another page without the Drop-in (just as a regular redirect flow) or at least it would show the result in another way. Also I tried to align this way of handling the redirect back with the one it's above, where we call the just `makeDetailsCall` directly.

marcperez (approved):

PR #870, opened 2021-03-19
Add Brazilian social security number to Card Component

marcperez:
## Summary
Add support for the `socialSecurityNumberMode` configuration property. Depending on the `countryCode` value and `socialSecurityNumberMode` we will request the `socialSecurityNumber` to Brazilian shoppers.


packages/lib/src/components/Card/components/CardInput/types.ts
```diff
@@ -35,6 +37,7 @@ export interface CardInputState {
     additionalSelectElements: DualBrandSelectElement[];
     additionalSelectValue: string;
     billingAddress: object;
+    brand: string;
```
marcperez, line 40:
```suggestion
    brand?: string;
```

Shouldn't this be optional @sponglord?

sponglord (approved):

Issue #882, opened 2021-03-24
Configurable field order for card holder name field


ColinMorris83:
**Is your feature request related to a problem? Please describe.**
We would like to use the drop in card component, but our UX designs want the card holder name to be the first form field rather than the last, meaning we will need to resort to using custom secured fields.

**Describe the solution you'd like**
A configurable setting for the position of the card holder field, e.g. top or bottom

**Describe alternatives you've considered**
Have used custom secured fields

pabloai:
Hi @ColinMorris83, thanks for reaching out.

I just created a PR to allow this, meaning it would be included in our next release, which is planned for next Monday.
I hope this helps, have a nice day!

marcperez:
This has now been released as part of `v4.2.0`. I'll close this issue.

PR #891, opened 2021-03-26
Add positionHolderNameOnTop for the Card component

pabloai:
Add option to position the holder name field on top of the Card component.

## Summary
Users have been recurrently asking for the possibility of placing the holder name field at the top of the Card component. After discussing it with the some of the Frontend team and with some of our designers, we've decided to allow this.
Therefore we're adding a `positionHolderNameOnTop` option which accepts a boolean and it's `false` by default.

## Tested scenarios
Added test to check if the positioning is working

**Fixed issue**:  Adyen/adyen-web#882

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.tsx
@@ -172,6 +172,17 @@ class CardInput extends Component<CardInputProps, CardInputState> {
 
         const showAmountsInInstallments = showInstallmentAmounts ?? true;
 
+        const CardHolderNameWrapper = () => (
+            <CardHolderName
+                required={this.props.holderNameRequired}
+                placeholder={this.props.placeholders.holderName}
+                value={this.state.data.holderName}
+                error={!!this.state.errors.holderName}
+                isValid={!!this.state.valid.holderName}
+                onChange={this.handleHolderName}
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -216,6 +216,17 @@ class CardInput extends Component<CardInputProps, CardInputState> {
...
                             <LoadingWrapper status={sfpState.status}>
+                                {hasHolderName && positionHolderNameOnTop && (
+                                    <CardHolderName
+                                        required={this.props.holderNameRequired}
+                                        placeholder={this.props.placeholders.holderName}
+                                        value={this.state.data.holderName}
+                                        error={!!this.state.errors.holderName}
+                                        isValid={!!this.state.valid.holderName}
+                                        onChange={this.handleHolderName}
+                                    />
```
marcperez, line 227: Maybe we could extract this into a variable to a void this duplication.

sponglord (approved):

packages/lib/src/components/Card/components/CardInput/CardInput.tsx
pabloai: Done!

marcperez (approved):

PR #918, opened 2021-04-01
Cleanup - pass an object to destructure rather than separate arguments

sponglord:
## Summary
Cleanup:
 - pass an object to destructure, rather than separate arguments, when handling securedFields encryption & validation.
 - removed unnecessary spread operator in ProcessBrand
 - added comment

## Tested scenarios
All tests pass


marcperez:
nice! 👍

marcperez (approved):

PR #932, opened 2021-04-06
Add KlarnaPayments component

marcperez:
## Summary
Add the [Klarna Widget](https://youtrack.is.adyen.com/issue/DOC-5625) component.

Creating a Klarna Widget component:
```js
checkout.create('klarna').mount('.klarna-container');
```

Creating a Klarna Widget Action component:

```js
checkout.createFromAction(action).mount('.klarna-container');
```


packages/lib/src/components/UIElement.tsx
```diff
@@ -145,7 +145,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> {
     }
 
     public payButton = props => {
-        return <PayButton {...props} amount={this.props.amount} onClick={this.submit} />;
+        return <PayButton amount={this.props.amount} onClick={this.submit} {...props} />;
```
sponglord, line 148: Why have you shifted the position of ...props?
We talked about this once before when I wanted to do this (for BacsDD) and you argued then that it was too fundamental a change and instead the component should override the `payButton` function i.e.
```
 public payButton = props => {
        return <PayButton amount={this.props.amount} onClick={this.submit} {...props} />;
    };
```

marcperez: Good point, moved to the KlarnaPayments element 👍

pabloai (approved):

PR #934, opened 2021-04-07
Add union types for SecuredFields fieldType and encryptedFieldName

gareth-canva:
## Summary

Adding additional typing to SecuredField types in the form of union types `SFFieldType` and `SFEncryptedFieldName` for the common `fieldType` and `encryptedFieldName` used in event payloads for event handlers in Adyen web components. 

We are in the process of upgrading to v4 of `adyen-web`, and had previously been using some hand-rolled types, including unions for these specific fields. This allowed us to (among other things) make use of exhaustive switch statements when handling events, which are a useful construct to ensure thorough handling of all possible field types in an event handler.
The current typings for these properties in the event payloads are generic strings, meaning our existing code would no longer work (you cannot exhaustively check a `string` type).

It seemed like the best first approach would be to see how difficult converting these fields to union types would be and try to contribute that change back upstream if you folks were interested, so here we are.

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/lib/core/handleEncryption.ts
+++ b/packages/lib/src/components/internal/SecuredFields/lib/core/handleEncryption.ts
@@ -7,7 +7,7 @@ import { SFFeedbackObj, CbObjOnFieldValid, EncryptionObj } from '../types';
 
 export function handleEncryption(pFeedbackObj: SFFeedbackObj): void {
     // EXTRACT VARS
-    const fieldType: string = pFeedbackObj.fieldType;
+    const fieldType = pFeedbackObj.fieldType;
 
     // SET FOCUS ON OTHER INPUT - If user has just typed a correct expiryDate - set focus on the cvc field OR typed a correct expiryMonth - focus on year field
     if (this.config.autoFocus) {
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/internal/SecuredFields/lib/core/handleEncryption.ts
```diff
@@ -7,7 +7,7 @@ import { SFFeedbackObj, CbObjOnFieldValid, EncryptionObj } from '../types';
 
 export function handleEncryption(pFeedbackObj: SFFeedbackObj): void {
     // EXTRACT VARS
-    const fieldType: string = pFeedbackObj.fieldType;
+    const fieldType = pFeedbackObj.fieldType;
```
gareth-canva, line 10: This change is not passing typechecking. Lines 14 and 17 (collapsed just below this diff) compare `fieldType` to `ENCRYPTED_EXPIRY_YEAR` and `ENCRYPTED_EXPIRY_MONTH`. However my typings have defined `fieldType` to be of type `SFFieldType`, which purposely does not include these individual month and year expiry fields, but rather only includes the parent `encryptedExpiryDate` field type. This assumption seems to hold elsewhere in the codebase, but I suppose my lack of domain knowledge here is starting to show. Maybe that's an incorrect assumption to hold, or maybe there should be additional, distinct types for the different `fieldType` properties of the various objects. Thoughts?

PR #968, opened 2021-04-15
fix(BankTransfer): showEmailAdress flag

emidev98:
## Summary
We have noticed that you have implemented the feature to send a copy to the email when there is a payment with Bank Transfer. Which is a very good feature but in our case we do not need it and we noticed that the **showEmailAddress** flag was not working.

### DIFF
--- a/packages/lib/src/components/BankTransfer/BankTransfer.tsx
+++ b/packages/lib/src/components/BankTransfer/BankTransfer.tsx
@@ -42,6 +42,10 @@ export class BankTransferElement extends UIElement<BankTransferProps> {
     };
 
     render() {
+        if(this.props.showEmailAddress === undefined || this.props.showEmailAddress === null){
+            this.props.showEmailAddress = BankTransferElement.defaultProps.showEmailAddress;
+        }
+
         if (this.props.reference) {
             return (
                 <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext}>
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/BankTransfer/BankTransfer.tsx
```diff
@@ -42,6 +42,10 @@ export class BankTransferElement extends UIElement<BankTransferProps> {
     };
 
     render() {
+        if(this.props.showEmailAddress === undefined || this.props.showEmailAddress === null){
+            this.props.showEmailAddress = BankTransferElement.defaultProps.showEmailAddress;
+        }
+
```
marcperez, line 48: We can safely assume `this.props` will already contain any `defaultProps` defined in the class (check the constructor on `BaseElement`)

packages/lib/src/components/BankTransfer/BankTransfer.tsx
```diff
@@ -53,10 +57,12 @@ export class BankTransferElement extends UIElement<BankTransferProps> {
...
             return (
                 <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext}>
-                    <BankTransferInput ref={this.handleRef} {...this.props} onChange={this.setState} />
+                    {this.props.showEmailAddress && (
+                        <BankTransferInput ref={this.handleRef} {...this.props} onChange={this.setState} />
+                    )}
                     <RedirectButton {...this.props} name={this.displayName} onSubmit={this.submit} payButton={this.payButton} />
                 </CoreProvider>
-            );
+            )
```
marcperez, line 65: This semicolon could stay

marcperez (changes requested):
Hi @emidev98!

Thanks for opening this PR! That's a nice catch and we really appreciate your help!
I added a couple comments, once they're resolved this can be merged. 🚀

packages/lib/src/components/BankTransfer/BankTransfer.tsx
emidev98: Well seen! Mecí, I have forgot to check that class

marcperez (approved):
Looks good to me! Thanks @emidev98!

emidev98:
Thank you! I'll look forward for more collaborations together

PR #1039, opened 2021-05-05
Allow custom icons and custom brand icons

marcperez:
## Summary
- Allow customizing icons of each payment method.
- Allow customizing brand icons on the card component.

All components accept an `icon` prop. In the Drop-in this will be displayed next to the name, in Components, this will be returned by `component.icon`.

```js
checkout.create('paywithgoogle', {
  someConfig: {},
  icon: 'htttps://...`
});
```

The Card component brand icons can also be customized through the `brandsConfiguration` property.

```js
{
   card: {
      brandsConfiguration: { 
        visa: {
          icon: 'https://...'
        }
      }
   }
}
```

### DIFF
--- a/packages/lib/src/components/Card/Card.tsx
+++ b/packages/lib/src/components/Card/Card.tsx
@@ -28,8 +28,8 @@ export class CardElement extends UIElement<CardElementProps> {
                 ...props.configuration,
                 socialSecurityNumberMode: props.configuration?.socialSecurityNumberMode ?? 'auto',
             },
-            brandsConfiguration: props.brandsConfiguration ?? props?.configuration?.brandsConfiguration ?? {},
-            icon: props.icon ?? props?.configuration?.icon,
+            brandsConfiguration: props.brandsConfiguration || props.configuration?.brandsConfiguration || {},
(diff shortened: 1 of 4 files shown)


packages/lib/src/components/Card/Card.tsx
```diff
@@ -28,6 +28,8 @@ export class CardElement extends UIElement<CardElementProps> {
                 ...props.configuration,
                 socialSecurityNumberMode: props.configuration?.socialSecurityNumberMode ?? 'auto',
             },
+            brandsConfiguration: props.brandsConfiguration ?? props?.configuration?.brandsConfiguration ?? {},
+            icon: props.icon ?? props?.configuration?.icon,
```
pabloai, line 32: Small comments: I would drop the `?` after the `props` in both lines. Also do you see some benefit in using the nullish concealing operator instead of using a regular `||` on this case? because the later means less weight.

marcperez: The size argument makes sense, I'll update this part 👍

pabloai (approved):

PR #1056, opened 2021-05-10
Add adyen-web package.json to exports

gareth-canva:
## Summary

It's not an uncommon pattern for bundlers/tools to do `require.resolve(path.join('@adyen/adyen-web', 'package.json'))` in order to find the location of a package's `package.json` file. However, for packages that have adopted the `exports` field in their `package.json`, this pattern won't work unless `package.json` itself is listed as an 'exported'/accessible file for the module. Here are some related discussions around this pattern: https://github.com/nodejs/node/issues/33460 and https://github.com/nodejs/modules/issues/516

A similar issue arose and was fixed by #837 for the CSS bundle in `adyen-web`.

This change allows the above pattern using `require.resolve` to work for `@adyen/adyen-web`.

@marcperez since you fixed the exports issue for the CSS bundle previously.

### DIFF
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -19,7 +19,8 @@
             "import": "./dist/es/index.js",
             "require": "./dist/cjs/index.js"
         },
-        "./dist/adyen.css": "./dist/adyen.css"
+        "./dist/adyen.css": "./dist/adyen.css",
+        "./package.json": "./package.json"
     },
     "version": "4.4.0",
     "license": "MIT",


marcperez:
Hi @gareth-canva,

Thanks for the PR! Looks good to me. 
Out of curiosity, did you encounter this issue with any particular tool or bundler?

marcperez (approved):

gareth-canva:
Thanks for the quick turnaround @marcperez. The tooling I ran into this issue with was internal, so nothing to worry too much about in this particular instance.

PR #1074, opened 2021-05-17
Accept socialSecurityNumberMode outside configuration on Card component

marcperez:
## Summary
Allows setting `socialSecurityNumberMode` on the root of the Card component configuration as well as in the `configuration` object.


sponglord:
I thought with v4 we were discouraging setting these kind of props directly on the component and that they _had_ to be in the `configuration` object.
For instance since #903 the ability to set `koreanAuthenticationRequired` directly as a prop has been removed from the Card comp

PR #1170, opened 2021-06-18
Add pt-PT translations

openseabrus:
## Summary
Added PT-PT translations to locales


pabloai:
Hi @openseabrus,

We try to be careful with the languages we include as each of them add weight to the library.
That being said, you can always [add your custom locales](https://docs.adyen.com/online-payments/components-web/localization-components#create-localization).

In this case, since the `pt-PT` locale share most of the strings with `pt-BR`, you could use `pt-BR` as a base and just [customize the localization](https://docs.adyen.com/online-payments/components-web/localization-components#customize-localization). Let us know if this would work for you.

I'm closing the PR as it's not something we're looking to add to the library, but we appreciate your contribution and encourage you to reach us out again in the future.

PR #1218, opened 2021-07-17
Fix apple pay button displaying

AlexVapniaruk:
Problem: 
When I tried to use "applepay" button on my integration, I was dissapointed because I cannot see it in any case. 
Later on clicking on the free space I found that button displaying but for some reason "-webkit-appearance: -apple-pay-button"
property is crossed and do not work.
I found that global property "-webkit-appearance: button" overwrite apple one.
So I make small fix for my project to make it work and want to  share it and help others to save time.

## Summary

After overwriting this with right one any type of buttons: "buy", "check-out" and etc. appers. This solution do not affect on other types of payment systems.
## Tested scenarios

To test: You need to use "applepay" paymentMethod.

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/ApplePay/components/ApplePayButton.module.scss
+++ b/packages/lib/src/components/ApplePay/components/ApplePayButton.module.scss
@@ -1,7 +1,7 @@
 @supports (-webkit-appearance: -apple-pay-button) {
     .apple-pay-button {
         display: inline-block;
-        -webkit-appearance: -apple-pay-button;
+        -webkit-appearance: -apple-pay-button !important;
         cursor: pointer;
     }
     .apple-pay-button-black {


sshniro:
The same thing occurred to me as well. Can we get a fix for this? @marcperez @pabloai @shernaz @sponglord 
Screenshot 2021-11-05 at 06 30 53

ribeiroguilherme:
@AlexVapniaruk thanks for point out the problem and contributing with the project.

We do not use `!important` rule in the CSS of our codebase. An alternative solution to fix the problem is by improving the specificity of the styles applied to the `button`. That was done on the PR #1354.

We released new version 5.3.0 which contains a fix for this problem. If there is any issues, feel free to reach out to us again.

PR #1266, opened 2021-08-31
Adds TWINT as stored payment method

m1aw:
## Summary

Adds TWINT payment method as an available stored payment method. Still uses the RedirectElement simply changes labels.

## Tested scenarios

* Displays "saved" after TWINT if it's a stored payment method
* Displays "pay + currency + amount" if it's stored payment method


packages/lib/src/components/Twint/Twint.tsx
```diff
@@ -0,0 +1,47 @@
...
+    public static defaultProps = {
+        type: TwintElement.type,
+        showPayButton: true
+    };
+    /**
+     * Get the element displayable name
+     */
+    get displayName(): string {
+        const { i18n, name, storedPaymentMethodId } = this.props;
+        return storedPaymentMethodId ? `${name} ${i18n.get('twint.saved')}`
```
sponglord, line 21: Does 'twint.saved' exist in the translations file already?

m1aw: No, not yet

m1aw: Updated the files

marcperez (approved):

Issue #906, opened 2021-03-29
Dropdown component not working when running the Drop-In inside a Web Component


bramdebouvere:
Hi,

We're integrating the drop-in (v ~3.17.0) inside a web component, and we're running into issues.

When we click the dropdown in the image below, you'll see that it immediately closes again. This is caused by the event listener that is added to the body (visible in the screenshot).
![image](https://user-images.githubusercontent.com/3641681/112857129-d1985f80-90b0-11eb-90cc-ee7e608ccd82.png)

You'll see that inside the event listener, it uses `e.target`. Since the drop-in is running inside a web-component, e.target will be the web component itself here. My suggestion is to use `e.composedPath()[0]` and fall back to `e.target` when `e.composedPath` is not available.

Kind regards

PR #1272, opened 2021-09-03
Fixes dropdown on Web Components

m1aw:
## Summary

Fixes dropdown click when inside a web component.

## Tested scenarios

* Selecting a iDeal issuer inside a web component version of Dropin
* Selecting a iDeal issuer on a normal test page

**Fixed issue**:  

#906

### DIFF
--- a/packages/lib/src/components/internal/FormFields/Select/Select.tsx
+++ b/packages/lib/src/components/internal/FormFields/Select/Select.tsx
@@ -78,7 +78,11 @@ function Select(props: SelectProps) {
      */
     const handleClickOutside = (e: MouseEvent) => {
         // use composedPath so it can also check when inside a web component
-        if (!e.composedPath().includes(selectContainerRef.current)) {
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/internal/FormFields/Select/Select.tsx
```diff
@@ -77,7 +77,8 @@ function Select(props: SelectProps) {
      * @param e - MouseEvent
      */
     const handleClickOutside = (e: MouseEvent) => {
-        if (!selectContainerRef.current.contains(e.target)) {
+        // use composedPath so it can also check when inside a web component
+        if (!e.composedPath().includes(selectContainerRef.current)) {
```
marcperez, line 81: [`composedPath`](https://caniuse.com/?search=composedPath) seems not to be available on some older browsers (e.g. IE11). We should add a fallback for those.

marcperez (approved):

Issue #1147, opened 2021-06-08
[Accessibility] - iOS VoiceOver prevents customers from selecting payment methods


johndoeblocks:
**Describe the bug**
At the checkout, it is not possible to change the payment method to something else than what is already selected. Since credit card is the default option, that is the only choice given to someone using VoiceOver. Any screen reader user who wishes to use another payment method will be unsuccessful in completing the order. 

By inspecting the code, we can see that these radio buttons are not native HTML radio buttons. And as such different screen readers might have problem getting focus. 

**Steps to reproduce the behavior:**

With an iPhone with VoiceOver on:
Settings > Accessibility > VoiceOver

1. Go to https://www.mystoredemo.io/#/ 
2. Add an item
3. Open cart and select checkout
4. Try switch between payment methods
5. See error

**Expected behavior**
As other screen readers the expected behavior would be to be able to switch between payment methods with a simple slide on the screen.

**Video**
Video displaying the bug.
https://user-images.githubusercontent.com/85570868/121215884-4e497600-c878-11eb-87a5-ea83610f3771.mp4

**Smartphone (please complete the following information):**
 - Device: iPhone 8
 - OS: iOS14.5
 - Browser: Chrome 

**Additional Context**
We are using the Adyen Web SDK v3.15.1 

Thank you in advance.

marcperez:
Hi @joaooliveiramanteigas,

Thanks for reaching out.

We're currently reviewing the accessibility of our Drop-in component. Your comment is very useful to us and we have added this issue to our internal backlog. Please let us know if you have any other accessibility feedback.

Issue #610, opened 2020-12-16
Drop-in component is not keyboard accessible


Mesoptier:
**Describe the bug**
There is an accessibility issue with the Adyen Drop-in component: when navigating with the keyboard (using the <kbd>Tab</kbd> key) it is not possible to navigate out of the Drop-in without also selecting a different payment method.

**To Reproduce**
This can easily be reproduced in the example provided with the Drop-in documentation (https://docs.adyen.com/checkout/prebuilt-ui#drop-in):

1. Make sure the "Credit Card" payment method is selected.
2. Place focus on one of the visible input fields.
3. Use the <kbd>Tab</kbd> key to navigate through the input fields, until reaching the final field (the "Pay" button).
4. Press the <kbd>Tab</kbd> key again. Now the next payment method is activated.

**Expected behavior**
What should happen after step 3 (from an accessibility perspective), is for the next focusable element after all payment methods to be focused.

The current behavior makes it practically impossible to use the form with a keyboard, especially if a custom submit button outside of the Drop-in is used (which is officially supported behavior, with the `showPayButton` option).

Suggested fix: remove the `tabindex` attribute from the payment method `<li>` elements. This would make the <kbd>Tab</kbd> behavior work properly. Then, to allow keyboard users to switch between payment methods, standard radio button behavior should be applied to the (currently faked) radio buttons.

**~~Screenshots~~ Screencasts**
I recorded a short screencast showing the reproduction steps:
https://www.youtube.com/watch?v=u6Yi-xL4vMU

**Desktop (please complete the following information):**
 - OS: Ubuntu & Windows
 - Browser: Google Chrome
 - Version: 88

**Smartphone (please complete the following information):**
N/A

**Additional context**

pabloai:
Hi @Mesoptier,

In my opinion the current situation when using the Pay button of the Drop-in works fine. The user can navigate through the payment methods and their fields by pressing <kbd>Tab</kbd> , and once they reach to the one they want to use, they can fill the details and press the Pay button. There's no need to jump to the first next element after the Drop-in, since the whole purpose of the Drop-in is to make a payment, and after that the Drop-in should not be there anymore.

That being said, I do agree there should be a way to navigate correctly when using a custom Pay button. I'll create an internal ticket to research this and get back to you once we have a solution.

Thanks for reaching out and for the detailed report!

bitbytebit1:
Would like to chime in here that press enter on mobile does not focus the next relevant element.

PR #1299, opened 2021-09-27
Fix accessible navigation between payment methods 

m1aw:
## Summary

Changes the behavior of Dropin payment method / list to act as an accordion following W3 accessibility guidelines, allow for navigation through payment method with using screen reader controls and without automatically switch between payments methods allowing users to "Tab" to the bottom of the page.

## Tested scenarios

* Ability to navigate the page using keyboard controls on Desktop
* Ability to navigate the page using desktop version of VoiceOver
* Buttons are read aloud with proper descriptions identifying payment methods including stored cards

**Fixed issue**:  

#1147 #610


sponglord:
Looks good to me. 
You could maybe add some e2e tests for the navigation on Desktop - but they don't have to be part of this PR

sponglord (approved):

packages/lib/src/style/utils.scss
```diff
@@ -0,0 +1,15 @@
+.adyen-u-visuallyhidden:not(:focus):not(:active) {
```
marcperez, line 1: I might be missing something, but Is this used anywhere?

marcperez (approved):

PR #1302, opened 2021-09-29
Fetch the right properties from the paymentMethods response 

marcperez:
## Summary
- Fetch the right properties from the paymentMethods response on components with multiple txVariants

## Tested scenarios
- `klarna_paynow` properties are set correctly, not defaulting to the base `klarna` ones.

### DIFF
--- a/packages/lib/src/core/core.ts
+++ b/packages/lib/src/core/core.ts
@@ -133,12 +133,13 @@ class Core {
          * Once we receive a valid class for a Component - create a new instance of it
          */
         if (isValidClass) {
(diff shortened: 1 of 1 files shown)


packages/lib/src/core/core.ts
```diff
@@ -133,12 +133,13 @@ class Core {
          * Once we receive a valid class for a Component - create a new instance of it
          */
         if (isValidClass) {
-            const paymentMethodsDetails = !options.supportedShopperInteractions ? this.paymentMethodsResponse.find(PaymentMethod.type) : [];
+            console.log(PaymentMethod.type, options.type);
```
m1aw, line 136: Console log sneaked in there

marcperez: Woops, good catch 👍

m1aw (approved):

PR #1324, opened 2021-10-13
Fixing e2e tests

ribeiroguilherme:
## Summary

- Fixed broken e2e tests after the v5 release
- Adjusted `customCards.config.js` to get errors from the `onChange` callback and apply them in the custom card fields

## Tested scenarios
- Run `yarn test:e2e-remote` in the packages/e2e repo

### DIFF
--- a/packages/e2e/app/src/pages/CustomCards/CustomCards.js
+++ b/packages/e2e/app/src/pages/CustomCards/CustomCards.js
@@ -4,7 +4,7 @@ import { handleSubmit, handleAdditionalDetails } from '../../handlers';
 import { amount, shopperLocale, countryCode } from '../../services/commonConfig';
 import '../../style.scss';
 import './customcards.style.scss';
-import { setCCErrors, setFocus, onBrand, onConfigSuccess, onBinLookup, onChange } from './customCards.config';
+import { setFocus, onBrand, onConfigSuccess, onBinLookup, onChange } from './customCards.config';
 
 const initCheckout = async () => {
     window.checkout = await AdyenCheckout({


sponglord:
Good fixes! Nice to have the e2e tests working again for v5. :-)

Regarding the change to: "customCards.config.js to get errors from the onChange callback " - if this is working to help the tests pass then great.
However the thing to consider is that the contents of the `playground` directory (or the equivalent directory in the e2e tests: `pages`) mimic what the merchant has to create in order to run Checkout. 
So the change to `customCards.config.js` represents what the merchant would have to do to in order to keep the same functionality in v5 as they used to have with the v4 `onError` handler.

I think a more robust solution is to enhance the object stored in `state.errors`, for a particular securedField, with the properties that we used to send to the `onError` handler.

This would mean that the `mapErrorsToValidationRuleResult` function in `SecuredFieldsProvider` should now look like this:
```
    /**
     * Map SF errors to ValidationRuleResult like objects, for CardInput component
     */
    public mapErrorsToValidationRuleResult(): object {
        const errorKeys = Object.keys(this.state.errors);
        const sfStateErrorsObj = errorKeys.reduce((acc, key) => {
            if (this.state.errors[key]) {
                acc[key] = {
                    isValid: false,
                    errorMessage: getError(this.state.errors[key]),
                    // For v5 the object found in state.errors should also contain the additional properties that used to be sent to the onError callback, namely: translation, errorCode & ref to rootNode
                    errorI18n: this.props.i18n.get(this.state.errors[key]),
                    error: this.state.errors[key],
                    rootNode: this.rootNode
                   // end v5 changes
                };
            } else {
                acc[key] = null;
            }
            return acc;
        }, {});
        return sfStateErrorsObj;
    }
```
(and there would need to be a small change to `SecuredFieldsInput.tsx` to call this function the way that `CardInput` currently does).

Issue #337, opened 2020-09-24
Reduce Bundle Size (excluding unused i18n and payment methods?)


JStonevalley:
**Is your feature request related to a problem? Please describe.**
The bundle size of the SDK is huge. There seems to be lots of i18n in the bundle that I will never use. As well as functionality for unused payment methods.

**Describe the solution you'd like**
It would be great if it was possible to include only the languages and paymentmethods that you need in the bundle to reduce the size.

**Describe alternatives you've considered**
Same as above

**Additional context**
-

marcperez:
Hi @JStonevalley,

Thanks for reaching! That's indeed something we want to improve and is in our roadmap and will be part of an upcoming major release.

JStonevalley:
Ok so there are no such options available at the moment?
Is it part of the next major release or is it further down the road?

marcperez:
Hi @JStonevalley,

Unfortunately, we don't have a timeline at this point, but it's something we are prioritizing for the next major version.

nachogarcia:
It might be great to have also a ES6 build so we can use it for modern browsers and reduze some size.

steffenmllr:
`121.5kB Minified + Gzipped` is more than all our other vendor packages combined https://bundlephobia.com/result?p=@adyen/adyen-web@3.18.2 - would be great if you could prioritize this.

matsinfiltratingvy:
```
Bundle without adyen: 232 KB (76 KB gzipped)
Bundle with adyen: 856 KB (227 KB gzipped)
```

So our entire webapp is about 200 KB with all code, css, libraries etc. Until we import adyen-web and it explodes in size. Given the millions of times this code is downloaded each day, you really should spend some effort on making it smaller.

marcperez:
Hi @matsinfiltratingvy,

Thank you for your comment. We're aware of this and currently working on a solution. 
The first step will include dynamically loading the language translation files and it will be released as part of our next major version (`v4.0.0`, which is planned for release in the next few weeks). After that, we're also planning on a similar solution for the different payment methods included in the library.

Please, let us know if you have any specific suggestions or feedback on this. Hope this helps!

matsinfiltratingvy:
Hi, thanks for getting back to me.

One thing could be to write it in such a way that it's easier to tree-shake/only include what's being used. Right now one passes in a string for which component to be rendered, which means you have to look it up dynamically and thus have everything in the bundle.

[...]

PR #1343, opened 2021-10-27
Use babel to transpile typescript, remove global polyfill and add modern bundle

m1aw:
## Summary

This ticket has 3 main objectives:
- Remove the pollution happening with globally scoped polyfills
- Atomically handle new polyfills
- Provide an unpolyfilled / "modern" version of the lib to reduce the transferred size

To achieve this we start using core-js-pure via babel-runtime, this explains why babel was introduced. 
For fast babel transpilation we started using babel-typescript-preset instead of doing 2 stages (tsc then babel).
For the "modern" build we use babel and babel-preset env with `target:"module"` and `bugfixes:true`.
Also the umd build now only runs when building the production bundle and not in dev to speed up compilation.

## Tested scenarios

- Unit test passed successfully 
- There was no increase in failed e2e tests
- The resulting bundles where tests on chrome for global scoped pollution (checked if Function.toString() was being polyfilled for example)
- Tested normal dropin flow on IE11

**Fixed issue**:  

Might help #337


packages/lib/config/rollup.config.js
```diff
@@ -32,9 +33,30 @@ const watchConfig = {
...
 
-async function getPlugins({ compress, analyze, version, useTypescript = true }) {
+const extensions = ['.js', '.jsx', '.ts', '.tsx'];
+
+const polyfillPlugin = [
+    '@babel/plugin-transform-runtime',
+    {
+        corejs: 3,
+        absoluteRuntime: true
+        //useESModules: true
```
marcperez, line 43: Is this necessary? Can it go?

packages/playground/config/polyfills.js
```diff
@@ -19,10 +19,3 @@ import 'whatwg-fetch';
         });
     });
 })([Element.prototype, CharacterData.prototype, DocumentType.prototype]);
```
marcperez, line 21: Will the new polyfills remove the need for this `ChildNode.remove()` one?

marcperez: (same for `fetch`)

ribeiroguilherme: If `fetch` is polyfiled by `core-js-pure`, we might be able to remove the dependency `whatwg-fetch`

ribeiroguilherme: If `fetch` is polyfiled by `core-js-pure`, we might be able to remove the dependency `whatwg-fetch`

m1aw: I just left this for the playground because there is still code we are not transpilling in there.

ribeiroguilherme: Sorry, I wasn't clear enough. I didn't mean in the playground; we use `whatg-fetch` as polyfill in our `lib` , on `src/core/Services/fetch.ts` . Can we ditch that polyfill too?

m1aw: we cannot remove this, using corejs there where errors on IE11

marcperez (approved):

lynn1286:
image

我在使用nextjs构建adyen ，但是 webpack chunks 异常庞大，是我的使用方式有误吗？
下面是我的package.json

```
"@adyen/adyen-web": "^5.15.0",
 "next": "12.1.6",
```

下面是我的代码：
```
import type { ChangeEventHandler } from 'react'
import React, { useEffect, useRef, useState } from 'react'
import { useRouter } from 'next/router'
import classnames from 'classnames'
import useSWR from 'swr'
import adyenCheckout from '@adyen/adyen-web'
import type Dropin from '@adyen/adyen-web/dist/types/components/Dropin'
import type { PaymentResponse } from '@adyen/adyen-web/dist/types/types'
import { paymentMethodsApi } from 'https/adyen-request'
import MessageAlert from '@/components/message/index'
import { usePaymentState } from '@/hooks/index'
import type { SessionsResponse } from '@/types/adyen'
import type { BillingAddress, ResultCode } from '@/hooks/use-payment-state'
import { PaymentStateAction, PaymentModelState } from '@/hooks/use-payment-state'
import '@adyen/adyen-web/dist/adyen.css'
```

PR #1349, opened 2021-11-01
bancontact.visa.test fixed

sponglord:
## Summary
- e2e DropinPage model created
- bancontact.visa.test converted to use this and failing test fixed

## Tested scenarios
bancontact.visa.test passes again

### DIFF
--- a/packages/e2e/tests/_models/Dropin.page.js
+++ b/packages/e2e/tests/_models/Dropin.page.js
@@ -1,0 +1,71 @@
+import { ClientFunction, Selector } from 'testcafe';
+import BasePage from './BasePage';
+import { getIframeSelector } from '../utils/commonUtils';
+import cu from '../cards/utils/cardUtils';
+
+/**
+ * The Page Model Pattern is a test automation pattern that allows you to create an
+ * abstraction of the tested page, and use it in test code to refer to page elements
+ */
+export default class DropinPage extends BasePage {
+    constructor(pmSelector) {
+        super('');
+
+        const BASE_EL = '.adyen-checkout__dropin';
+
+        /**
+         * ###################
+         * Credit card related
+         * ###################
+         */
+        this.brandsHolder = Selector(`${BASE_EL} .adyen-checkout__payment-method__brands`);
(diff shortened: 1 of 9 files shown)


packages/e2e/tests/_models/Dropin.page.js
```diff
@@ -0,0 +1,71 @@
...
+         */
+        this.dualBrandingIconHolder = Selector(`${BASE_EL} .adyen-checkout__card__dual-branding__buttons`);
+        this.dualBrandingIconHolderActive = Selector(`${BASE_EL} .adyen-checkout__card__dual-branding__buttons--active`);
+        this.dualBrandingImages = this.dualBrandingIconHolderActive.find('img');
+
+        /**
+         * iframe utils
+         */
+        this.iframeSelector = getIframeSelector(`${BASE_EL} ${pmSelector} iframe`);
+        this.cardUtils = cu(this.iframeSelector);
```
ribeiroguilherme, line 62: Just brainstorming regarding the page model pattern. Can't we reuse the logic of `CardComponent.page.js` here? 

As this class evolves and we add more Payment methods, this class will grow a lot. It would be great to reuse the logic of our Components classes. This approach would allow us to isolate code for our Components, and reuse them within the Dropin page.

One way that I imagine that 

```javascript

export default class DropinPage extends BasePage {
    constructor({ components }) {
        Object.assign(this, components);
         ...
     }
     ...
}
```

```javascript

// dropin.test.js

const dropinPage = new DropinPage({
    components: {
        creditCard: new CardComponent('adyen-checkout__payment-method--bcmc'), 
        ideal: new IssuerListComponent('adyen-checkout__payment-method--ideal'),
        molpay: new IssuerListComponent('adyen-checkout__payment-method--molpay')

fixture`Testing Dropin with ...
```

ribeiroguilherme (approved):

packages/e2e/tests/_models/Dropin.page.js
sponglord: Yes. I agree we need to find a method to avoid code duplication.
We'll have to investigate your proposal...

PR #1351, opened 2021-11-03
DropinPage model is instantiated with an object containing refs to other page models

sponglord:
## Summary
re. e2e tests
Following a suggestion from @ribeiroguilherme - the `DropinPage` model is now instantiated with an object containing refs to other page models.
This allows us to isolate code in our Component pages and reuse them within the Dropin page.

## Tested scenarios
All e2e tests pass


ribeiroguilherme (approved):
Very nice 🚀

PR #1353, opened 2021-11-05
Fix/a11y issues

sponglord:
## Summary
Tackling a11y issues in COWEB-982:
- adding `aria-hidden="true"` on cvc hint div
       re. FOC-54217 (a11y spreadsheet ticket 4)
- adding `aria-hidden="true"` on card brand logos in CardNumber field
      re. COWEB-1032 (a11y spreadsheet ticket 9.3)
- Add possibility to configure securedFields iframe `title` to not exist and instead replace it with `role="presentation"`.  
      re. COWEB-1032 (a11y spreadsheet ticket 9.1)
- Removed `focusable` attr on the `` elements for the card brands list in the Dropin
      re. COWEB-983 (a11y spreadsheet ticket 1.1)
- Removed deprecated `scrolling` & `frameborder` attributes from securedFields iframe
      re. COWEB-983 (a11y spreadsheet ticket 1.2)

## Tested scenarios
re. removing `scrolling` & `frameborder` attributes from securedFields iframe - iframe still works in IE11 (tested in Browserstack Windows 8.1 & 10)

**Fixed issue**:  (see refs above)

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/lib/utilities/createIframe.ts
+++ b/packages/lib/src/components/internal/SecuredFields/lib/utilities/createIframe.ts
@@ -2,7 +2,11 @@ export default function createIframe({ src, title = 'iframe element', policy = '
     const iframeEl = document.createElement('iframe');
     iframeEl.setAttribute('src', src);
     iframeEl.setAttribute('class', 'js-iframe');
-    iframeEl.setAttribute('title', title);
+    if (title !== '' && title.trim().length !== 0) {
+        iframeEl.setAttribute('title', title);
+    } else {
+        iframeEl.setAttribute('role', 'presentation');
+    }
     iframeEl.setAttribute('frameborder', '0'); // deprecated but still necessary for IE TODO re-test this on next round of IE testing
     iframeEl.setAttribute('scrolling', 'no');
     iframeEl.setAttribute('allowtransparency', 'true');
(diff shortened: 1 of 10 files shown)


packages/lib/src/components/internal/SecuredFields/lib/utilities/createIframe.ts
```diff
@@ -1,10 +1,17 @@
...
     iframeEl.setAttribute('src', src);
     iframeEl.setAttribute('class', 'js-iframe');
-    iframeEl.setAttribute('title', title);
-    iframeEl.setAttribute('frameborder', '0'); // deprecated but still necessary for IE TODO re-test this on next round of IE testing
-    iframeEl.setAttribute('scrolling', 'no');
+    if (title !== '' && title.trim().length !== 0) {
+        iframeEl.setAttribute('title', title);
+    } else {
+        iframeEl.setAttribute('role', 'presentation');
+    }
```
ribeiroguilherme, line 14: Can you elaborate what the `role=presentation` attribute means on the iframe usage?

sponglord: You need to refer to the a11y spreadsheet I shared in CheckoutDevs chat: Ticket 9.1
Alternatively look at my last comment, point 1 in COWEB-1032

ribeiroguilherme: Clear. Thanks!

PR #1354, opened 2021-11-05
Instant Payments on Drop-in

ribeiroguilherme:
## Summary
Merchants should be able to configure Drop-in to show instant payments on top of the regular payment method list.

- Introduced property `instantPaymentTypes` as part of the `dropin` configuration, which can contain the payment types which will show on top of the regular payment method list.
  - Only `Google Pay` and `Apple Pay` will be allowed for now
- Other payment methods passed as part of the configuration which are not included as part of the supported payment methods are discarded
- The order of the instant payment methods are kept during the render

Usage example:
```javascript
const dropin = checkout.create('dropin', {
    instantPaymentTypes: ['paywithgoogle', 'applepay']
}).mount('#dropin-container');
```

## Tested scenarios

- Tested performing payments with `Google Pay`
- Tested performing payments with `Apple Pay` using Simulator app
- Tested opening it on Safari to see both buttons piled on each other
- Added unit tests for Drop-in parsing the new prop; PaymentMethodList rendering the new component; UIElement closing active payment methods if instant payment is clicked


packages/lib/src/components/ApplePay/components/ApplePayButton.module.scss
```diff
@@ -1,7 +1,15 @@
 @supports (-webkit-appearance: -apple-pay-button) {
+
+    /*
+    *  Combination of both classes improve the specificity, avoiding
+    *  overwrite of the -webkit-appearence by the button native css
+    */
+    .apple-pay,
     .apple-pay-button {
```
ribeiroguilherme, line 8: This should fix #1218, as increase the specificity of the apple pay button

marcperez (approved):
Looks good to me!

PR #1370, opened 2021-11-17
Fixes padding for payment method focus state

m1aw:
## Summary

Focus state border was right against the text and icon of the payment method button. This improves it by adding some padding and remove some padding from its container.

## Tested scenarios

- Tested focus on payment methods
- Tested partial payments / order payments for side effects

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
+++ b/packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
@@ -55,15 +55,16 @@
     justify-content: space-between;
     font-weight: 400;
     font-size: $font-size-medium;
-    padding: $spacing-medium - $spacing-xsmall;
-    padding-left: $spacing-xxlarge - $spacing-xsmall;
+    $payment-button-padding: $spacing-xsmall;
+    padding: $spacing-medium - $payment-button-padding;
+    padding-left: $spacing-xxlarge - $payment-button-padding;
     position: relative;
     transition: background 0.1s ease-out;
     width: 100%;
 
     [dir='rtl'] & {
-        padding: $spacing-medium - $spacing-xsmall;
-        padding-right: $spacing-xxlarge - $spacing-xsmall;
+        padding: $spacing-medium - $payment-button-padding;
+        padding-right: $spacing-xxlarge - $payment-button-padding;
     }
 
     .adyen-checkout__payment-method--standalone & {
--- a/packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
+++ b/packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
@@ -55,15 +55,15 @@
     justify-content: space-between;
     font-weight: 400;
     font-size: $font-size-medium;
-    padding: 16px;
-    padding-left: 48px;
+    padding: $spacing-medium - $spacing-xsmall;
+    padding-left: $spacing-xxlarge - $spacing-xsmall;
     position: relative;
     transition: background 0.1s ease-out;
     width: 100%;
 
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
```diff
@@ -55,15 +55,15 @@
     justify-content: space-between;
     font-weight: 400;
     font-size: $font-size-medium;
-    padding: 16px;
-    padding-left: 48px;
+    padding: $spacing-medium - $spacing-xsmall;
```
ribeiroguilherme, line 58: Why not use `$spacing-medium-small` here? Goal is to use `12px`, right?

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
```diff
@@ -55,15 +55,15 @@
...
+    padding-left: $spacing-xxlarge - $spacing-xsmall;
     position: relative;
     transition: background 0.1s ease-out;
     width: 100%;
 
     [dir='rtl'] & {
-        padding: 16px;
-        padding-right: 48px;
+        padding: $spacing-medium - $spacing-xsmall;
+        padding-right: $spacing-xxlarge - $spacing-xsmall;
```
ribeiroguilherme, line 66: Can you elaborate why do we need this calculation here to set `44px`?

ribeiroguilherme (commented):
It would be great if you add some prints of the changes next time. It will make easier to understand them, as there is no ticket on our board about this task

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
m1aw: Because in my opinion makes it more obvious why there is -4px, it's the padding of the button element.

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
m1aw: We don't need a calculation, but 44px is a bit of a magic number, this hopefully gives a bit more reasoning on why. 
In this case because the spacing was moved from the padding of the container, to the padding of the button itself.

m1aw:
Old focus state
image

New focus stat
Screen Shot 2021-11-16 at 4 24 12 PM

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
m1aw: Also, if we all agree this is ugly I will just change it to 44px and 12px. And maybe add a comment somewhere, the objective is readability.

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
ribeiroguilherme: I understand. It makes sense from your perspective. But if someone works on this file at some point, he/she might wonder why this. If the goal is to not use `$spacing-medium-small` in order to make it more obvious the calculation being done here for the padding, you could do:

```
$button-padding: $spacing-xsmall;
padding: $spacing-medium - $button-padding;
```

It would make it more explicit why is like that.

ribeiroguilherme (approved):

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodItem.scss
m1aw: I do like this even more, I will update it to add that exactly that.

PR #1378, opened 2021-11-22
Paypal - Updating currency after re-mounting

ribeiroguilherme:
## Summary
PayPal component isn't updating the currency after re-mounting with a different currency value. This PR fixes it, and also makes sure that merchant can load multiple PayPal buttons on the screen.

## Tested scenarios

The feature was tested using the following approach:
- Assembled Playground with 3 Paypal buttons, which one of them  is loaded after a timeout. This assures that merchant can load multiple buttons in different intervals
- Created global function`reload()`, which can be called on the console in order to unmount and mount the component using a different currency

If you would like to test the feature in the modified playground, you can:
```
git checkout 4eb03d01f71e8c4cdcfe22e84c5f543e38a33948
```
which will load the commit with the customized playground.

### DIFF
--- a/packages/lib/src/utils/Script.ts
+++ b/packages/lib/src/utils/Script.ts
@@ -33,7 +33,10 @@ class Script {
 
             this.script.src = this.src;
             this.script.async = true;
-            this.script.onload = resolve;
+            this.script.onload = event => {
+                this.script.setAttribute('data-script-loaded', 'true');
+                resolve(event);
+            };
             this.script.onerror = () => {
                 this.remove();
                 reject(new Error(`Unable to load script ${this.src}`));
@@ -43,7 +46,9 @@ class Script {
             const addedScript = container.querySelector(`script[src="${this.src}"]`);
 
             if (addedScript) {
-                addedScript.addEventListener('load', resolve);
+                const isLoaded = addedScript.getAttribute('data-script-loaded');
+                if (isLoaded) resolve(true);
+                else addedScript.addEventListener('load', resolve);
             } else {
                 container.appendChild(this.script);
             }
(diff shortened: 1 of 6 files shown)


packages/lib/src/utils/Script.ts
```diff
@@ -33,7 +33,10 @@ class Script {
 
             this.script.src = this.src;
             this.script.async = true;
-            this.script.onload = resolve;
+            this.script.onload = event => {
+                this.script.setAttribute('data-script-loaded', 'true');
```
pabloai, line 37: Maybe we could rename this attribute to something more component-specific like `data-paypal-script-loaded`. This way we could use a similar approach in other components if needed.

pabloai (changes requested):

packages/lib/src/utils/Script.ts
ribeiroguilherme: I'd rather keep it generic as it is now. The `Script` class is a generic class, and it is used by other UI components. We shouldn't couple it to PayPal here.

This code fixes that issue which the `resolve` wasn't being called when loading the script twice. Reason for that is that the script was already loaded, so the event listener isn't fired again. 

https://github.com/Adyen/adyen-web/blob/341b082213b480a36692c9ec7fe22468834223a5/packages/lib/src/utils/Script.ts#L51

Proposed solution here is that as soon as the script is loaded, we flag is as loaded, and then if it gets called again, we simply resolve the promise instead of adding an event listener which will never be called.

I believe that other payment methods will benefit from this change. Btw, I tested with GooglePay locally and it worked fine.

pabloai: Ah, my mistake I see the change is done in Script.ts directly. Thanks for the clarification.

pabloai (approved):

Issue #1315, opened 2021-10-07
Angular SSR: ReferenceError: Element is not defined


drafenous:
**Describe the bug**
I'm trying to enable SSR on my Angular project with SAP Spartacus, but when i instance with 'new AdyenCheckout(config)' and enable Angular Universal SSR, then crashes with message:
ReferenceError: Element is not defined

**To Reproduce**
Implement Angular Universal SSR on project.

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/8341200/136464907-623b145c-88b5-43c8-ab00-e81a6e5cdca6.png)
If uncomment the line 151, then crashes.

**Desktop (please complete the following information):**
 - OS: Windows 10 x64
 - Browser: Chrome 94
 - Node Version: 12.22.1
 - Npm Version: 6.14.12
 - Package Version: "@adyen/adyen-web": "^4.7.3",
 - Angular Version: 10.2.3

**Additional context**
I tryed to isolate with isPlatformBrowser() from '@angular/common' and make a conditional import of component, but no success.

ribeiroguilherme:
Hi @drafenous 

Thanks for reaching out. Apparently the issue is that Angular Universal is executed in the server, therefore some of the browser APIs and capabilities might be missing there, like the `window`, `document`, `navigator`, etc. And `adyen-web` interacts with some of these API's. This can be the reason why you get this error.

You can read more about it here: 
https://angular.io/guide/universal#working-around-the-browser-apis

As it states in the docs, Angular provides some abstractions for `Location` and `Document` browser API's. But you still need to go around and adapt it for the `window` object. There is a blog post here that might be useful: http://michaelborisov.com/blog

I'll close this ticket, but feel free to reopen if you think this is still an issue related to this library.

Hope this helps!

drafenous:
Sorry for the delay, lost a busy week with the first release of our ecommerce.

We've tried so far:
- Isolate the intantiation of AdyenCheckout(this.configuration) validating with isPlatformBrowser, but without success. Below is an example.

![image](https://user-images.githubusercontent.com/8341200/139482999-f5f4b656-f3b3-4872-9828-2c0efb76c347.png)
...

![image](https://user-images.githubusercontent.com/8341200/139483044-fcc2a8dc-9979-4d06-aaaa-2d19985f997b.png)


[...]

Issue #902, opened 2021-03-29
Importing adyen-web in next.js


trnc-ck:
hey,

we are unable to include the adyen web-package in our next.js app, as the following error occurs as soon as we import the lib:

`Server Error

ReferenceError: Element is not defined
This error happened while generating the page. Any console logs will be displayed in the terminal window.
Source

external%20%22@adyen/adyen-web%22 (1:0) @ eval
> 1 | module.exports = require("@adyen/adyen-web");
`

marcperez:
Hi @trnc-ck,

Thanks for reaching. Could you share how are you installing and importing the library? What version are you using?

trnc-ck:
hey,

we just install your lib using `yarn add @adyen/adyen-web`

and add it like this:
`import AdyenCheckout from '@adyen/adyen-web';`
`import '@adyen/adyen-web/dist/adyen.css';`

we didn't write any further code. the error occurs just using the import. regarding next.js we use the latest version.

trnc-ck:
Ok, we fixed it… You need to load the component dynamically and disable the SSR-feature:
https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr

Thanks anyway!

maheshk06:
Hi @trnc-ck
Hope you are doing good.

In Next.JS I am getting a similar issue.
I tried what you suggested.

`const AdyenCheckout = dynamic( () => import( '@adyen/adyen-web' ), { ssr: false } );`
throwing an error.
![image](https://user-images.githubusercontent.com/57614818/194690229-c93fd304-576b-4a10-a97c-190c25a60d41.png)

these line are making issues.
`      const checkout = yield call( AdyenCheckout, configuration );
      checkout.createFromAction( action ).mount( '#credit-card-con' );`

![image](https://user-images.githubusercontent.com/57614818/194690284-8cfc04ca-437b-4d8b-9043-5de1eb0788c8.png)

saga file code - 
`import Router, { useRouter } from 'next/router';
import Logger from 'xps-utils/logger';
import { call, select, put } from 'redux-saga/effects';
import { serviceRequest, getPaymentVersion } from 'library/store/payment/service-request';
import { selectTotalGrossAmt, selectBillingAddress, selectSubscriptionItemGrossAmt, selectShouldReAuthorised, selectCustom } from 'library/store/cart/selector';
import { selectPaymentId, selectPaymentDetail, selectActivePayment } from 'library/store/payment/selector';
[...]

Issue #233, opened 2020-08-31
Make code crash-safe in SSR environments


Knusperchicken:
**Is your feature request related to a problem? Please describe.**
When importing this package into a React component which is being rendered on the server, the code is getting executed on the server side resulting in a crash with messages like `ReferenceError: window is not defined`.

**Describe the solution you'd like**
As I don't always have control over the component being rendered on the server or client side it would be nice for the code to not crash if this code is being executed on the server side.

**Describe alternatives you've considered**
As mentioned above I don't have control over the general architecture and config of the project thus I can't control whether this packge should only be imported on the client side or vice versa. There is this hacky way that I tried:

```
let AdyenCheckout

try {
  // eslint-disable-next-line global-require
  AdyenCheckout = require('@adyen/adyen-web')
} catch (_) {
  AdyenCheckout = class {}
}
```

This works but it isn't the way we should do it when using a React component, right? 😉

**Additional context**
I've tried myself to fix this by adding a `globalThis` polyfill and replacing all occurences of `window` with `globalThis` but it was to no avail. Build was good, tests too and the sample pages seemed to work fine but when I copied the built files over to my project, the dev server crashed imemdiately upon startup. Maybe I just missed something and additionally I'm not a TypeScript pro ... 🤔

pabloai:
Hi @Knusperchicken,

Thanks for reaching and for providing a clear explanation of the issue.
We'll look into it and let you know once we have a solution.

Have a nice day!

pabloai:
Hi again @Knusperchicken,

Would using [dynamic imports](https://nextjs.org/docs/advanced-features/dynamic-import) help in you scenario?
Here's a quick demo: https://codesandbox.io/s/adyen-web-nextjs-xkvdb

Let us know what you think, cheers!

pabloai:
Closing this issue. Feel free to reach us again.
Cheers!

Knusperchicken:
Hi @pabloai ,

I'm sorry for the late response. We might have found another solution for our framework which seems to work. Thank you anyway for your effort.
🍪

mhb-yt:
Hey @Knusperchicken, 

What solution have you found?

Knusperchicken:
Hi @mhb-yt,

disregarding the "hack" (which works fine though) above I was able to solve this by importing and creating the checkout object at another place in our framework which was executed only on the client side. I'm afraid this answer won't really help you though 😕 
[...]

PR #1384, opened 2021-11-24
Fix Element reference error when importing library outside browser context

jonlambert:
## Summary

When importing the `@adyen/adyen-web` library in a server-side context, [the app will crash](https://github.com/Adyen/adyen-web/issues/1315#issuecomment-977815471) just through the import alone. This means you can't simply perform a runtime check before instantiating the AdyenCheckout class, and requires you to dynamically import any components that use this library which adds quite a lot of overhead. 

This PR checks for the presence of `Element`, `navigator` and `window` objects before referencing at the top-level of each file. I haven't bothered adding presence checks within function calls, etc, as full SSR support would fall outside the scope of this PR. But at least now the library can be imported in a server-side context without crashing or requiring dynamic importing.

I managed to spin up a dev environment locally to reproduce and test this; hopefully someone with a more in-depth knowledge of the codebase can review 🙂 Would be great to get thoughts.

Thanks a lot.

**Fixed issue**:  #1315, #902, #233


jonlambert:
Have updated to check for the presence of `Element`, `navigator` and `window `objects whenever they're used at the top level of a file.

jonlambert:
(There may be some crossover with work done in #1343 here)

jonlambert:
Hey @ribeiroguilherme – further to our chat in #1315 do you have any thoughts on this? Sorry to nudge, but we're using my fork at the moment and before we proceed much further it would be good to get an idea of whether this might be merged or not 🙂 Thanks a lot.

ribeiroguilherme (approved):
@jonlambert thanks for contributing to the project. I tested in a `next.js` project and worked well. 👍

PR #1388, opened 2021-11-26
Feature/consolidated error panel

sponglord:
## Summary
The issue with how the card component presents errors that can be picked up by screenreaders is a constantly recurring one.
Current behaviour has these errors set on an individual basis for each input & they are read, individually, as they occur - but many merchants have complained that: 
- it's hard  for visually impaired shoppers to know that a field is still in error when they move to a new field 
- that when pressing the pay button it is impossible to know which fields are now in error
- that it's hard for the shopper to know how to navigate to a field that is in error i.e it should be possible to immediately remedy errors can  with minimum interaction by the shopper

So this PR consolidate credit card errors into a single field for the screenreader to broadcast:
- This field sits in a new `ErrorPanel` component that acts as an `aria-describedby` for the `CardInput` "form"
- Errors accumulate in this panel as they occur
- This panel is always "active" -  so, for example, it would inform the shopper that the card number field was in error even when in another field
- When validation is applied, i.e. the pay button is pressed - the first error in this list indicates an element to set focus on: which we then do
- The original, visual, errors still exist below their respective input's - but these are now hidden from the screenreader and no longer act as an `aria-describedby` for that input

A new property `SRConfig` can be set in the card config that allows us to turn this new functionality on/off:
```
SRConfig:{
  collateErrors: true|false // defaults to true
}
```
When this functionality is on it is also possible to configure whether we automatically switch focus to the first field in error when the pay button is pressed. 
(Note: whilst some merchants have specifically asked for this feature, other opinions say that it is very disorienting for the shopper to sudden have focus shifted away to another field - for this reason it defaults to false): 
```
SRConfig:{
  moveFocus: true|false // defaults to false
}
```
This `ErrorPanel` is currently only applied to the Card component, since this is where most of the a11y feedback has been directed, but it has been designed such that it should be easy to roll out to other PMs/components

## Tested scenarios
New e2e tests added specifically to test this new functionality:
- errors are in the expected order, reflecting the layout of the credit card form (which varies massively depending on how the card is configured - does it have a holderName? Where is that positioned? Does it have AVS fields? Does the order of the errors change as AVS country changes? Does it have SSN or KCP fields etc?
- the relevant element gets focus when the pay button is pressed (or doesn't if the config says it shouldn't)
- the config works to turn all this new functionality off and presents the Card as it used to be
- the panel is hidden from view
- all the other e2e tests run and pass

**Relates to issue**:  COWEB-960

## Note to reviewers
A big PR this, it touched many files - but most are cosmetic and just involve passing on the value of the `collateErrors` config prop.
The significant changes are in:
- `CardInput.tsx`
- `CardInput/uitls.ts`
- `Errors/ErrorPanel.tsx`


sponglord:
After a suggestion from @ribeiroguilherme - have used `CoreContext` to alleviate some of the prop drilling required to pass on the fact of whether we are collating errors or not

packages/lib/src/components/Card/components/CardInput/types.ts
```diff
@@ -94,3 +99,46 @@ export interface CardInputState {
...
+    showPanel?: boolean;
+}
+
+interface FieldError {
+    errorMessage?: string;
+    errorI18n?: string;
+    // error: string;
+    // rootNode?: HTMLElement;
+    // isValid?: boolean;
+    // shouldValidate?: boolean;
```
ribeiroguilherme, line 115: can they be removed?

packages/lib/src/core/Errors/ErrorPanel.tsx
```diff
@@ -0,0 +1,45 @@
...
+    errors,
+    callbackFn = null,
+    showPanel = false
+}: ErrorPanelProps) {
+    if (!errors) return null;
+
+    const { errorMessages } = errors;
+
+    // Perform passed callback, if specified
+    callbackFn?.(errors);
```
ribeiroguilherme, line 29: Is it intentional to leave the callback here instead of putting it inside an `useEffect` function?

Every time that set state happens, this component likely will render and the callback will be called

ribeiroguilherme (approved):

packages/lib/src/components/Card/Card.tsx
```diff
@@ -17,6 +17,9 @@ export class CardElement extends UIElement<CardElementProps> {
     };
 
     formatProps(props: CardElementProps) {
+        // Extract &/or set defaults for the screenreader error panel
+        const { collateErrors = true, moveFocus = false, showPanel = false } = props.SRConfig ?? {};
```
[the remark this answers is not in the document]
sponglord: Not on the `Card.tsx` we don't - but you're right, we can define one

packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -15,17 +15,30 @@ import StoreDetails from '../../../internal/StoreDetails';
...
+import { ValidationRuleResult } from '../../../../utils/Validator/Validator';
 
 function CardInput(props: CardInputProps) {
     const sfp = useRef(null);
     const billingAddressRef = useRef(null);
+    const isValidating = useRef(false);
+
+    const errorFieldId = 'creditCardErrors';
+
+    const { collateErrors = true, moveFocus = false, showPanel = false } = props.SRConfig;
```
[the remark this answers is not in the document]
sponglord: true

packages/lib/src/components/Card/components/CardInput/types.ts
sponglord: yup

packages/lib/src/core/Errors/ErrorPanel.tsx
sponglord: Good point - we have a check, in the callback, that prevents the callback from running again - so I wasn't noticing it being called excessively... but this is better

Issue #1331, opened 2021-10-20
setStatus to ready doesn't work when setStatusAutomatically is true


wesleyhf:
**Describe the bug**
When I try to setStatus to `ready` inside of `onSubmit` it is not work when `setStatusAutomatically` is `true`.

**To Reproduce**
1. set `setStatusAutomatically` to `true`
2. inside of `onSubmit` try to call `dropin.setStatus('ready');`

**Expected behavior**
After the `onSubmit` call to my back-end I want to set the `dropin` to `ready` state.

**Desktop (please complete the following information):**
 - OS: MacOS 11.6
 - Browser: Chrome
 - Version: 94.0.4606.71 (Official Build) (x86_64)

**Additional context**

```js
const checkout = new AdyenCheckout({
    setStatusAutomatically: true,
    onSubmit: (state, dropin) => {
        handleRequest(state, dropin);
        dropin.setStatus('ready');
    },
});
```

wesleyhf:
Hello @ribeiroguilherme, sorry for it 😅 .

I was trying to use the `4.9.0` version, through the CDN files:

- https://checkoutshopper-test.adyen.com/checkoutshopper/sdk/4.9.0/adyen.js
- https://checkoutshopper-test.adyen.com/checkoutshopper/sdk/4.9.0/adyen.css

ribeiroguilherme:
Hi @wesleyhf ,

Indeed there is an issue with `setStatusAutomatically` with our Drop-in component. I created a ticket in our backlog to fix it. Thanks for letting us know. I will update this ticket as soon as we fix it.

If you would like to still use it with `setStatusAutomatically: true`, here is an workaround:

```javascript
const checkout = new AdyenCheckout({
    setStatusAutomatically: true,
    onSubmit: async (state, dropin) => {
        dropin.setStatus('loading');
        await handleRequest(state, dropin);
        dropin.setStatus('ready');
    },
});
```

ribeiroguilherme:
Hi @wesleyhf ,

We released the version 5.6.2 which contains the bugfix for this issue. Thanks again for reaching out!

PR #1394, opened 2021-11-30
Fixing setStatusAutomatically configuration property

ribeiroguilherme:
## Summary

The property `setStatusAutomatically` isn't working properly after v5 release.

This PR:
- Fixes issue which user could interact with other payment methods while a payment is being submitted. This is caused because the loading is set on the active payment method but not in the whole drop-in;
- Adjusted `setStatusAutomatically`  to work with the sessions flow;
- Introduced interface of UIElement which shows what methods are available outside the class

## Tested scenarios
- Tested using `setStatusAutomatically=false` with sessions.
- Validated that the whole drop-in is in `loading` state properly; not only the active payment method
- Validated that the problem reported on #1331 is gone
- Adjusted unit tests

**Fixed issue**: #1331 
**Reference**: COWEB-1037


packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -78,7 +78,7 @@ class ApplePayElement extends UIElement<ApplePayElementProps> {
                         this.setState({ applePayToken: btoa(JSON.stringify(event.payment.token.paymentData)) });
                     }
 
-                    this.onSubmit();
+                    super.submit();
```
sponglord, line 81: Why the change to explicitly call `super`?

packages/lib/src/components/UIElement.tsx
```diff
@@ -71,46 +80,49 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> {
...
+        if (this.props.isDropin && this.elementRef?.setStatus) {
+            this.elementRef.setStatus(status, props);
+        } else if (this.componentRef && this.componentRef.setStatus) {
+            this.componentRef.setStatus(status, props);
+        }
         return this;
     }
 
-    submitPayment(data): Promise<void> {
-        this.setStatus('loading');
```
sponglord, line 94: So is this setting of status no longer needed?

packages/lib/src/components/ApplePay/ApplePay.tsx
ribeiroguilherme: There is already a `submit` method declared in this component. In order to call the one of the parent component, I had to use `super`.

sponglord: Ok, it's just that it used to call `super.onSubmit()` directly (albeit via the subclass' `this.onSubmit()`) but now it calls `super.submit()` which first performs validation before it calls `this.onSubmit` in the `UIElement`.
It makes sense to validate first - but I just wanted to check on the slight change in behaviour

packages/lib/src/components/UIElement.tsx
ribeiroguilherme: The status is set here before calling `submitPayment`, in this part:

https://github.com/Adyen/adyen-web/blob/b397d48b8850e42172f1904cd567bb8dd5968917/packages/lib/src/components/UIElement.tsx#L51-L53

sponglord: ok

packages/lib/src/components/ApplePay/ApplePay.tsx
ribeiroguilherme: I get it. As I defined the interface of the `UIElement`, the `onSubmit` isn't available anymore on the subclasses since it is a private method. It is better to force the components to go through the `submit` method , which will validate before doing the actual request.

sponglord: ok, cool. thanks for the explanation :-)

sponglord:
So it all looks good to me. 
@marcperez - any thoughts?

isaacmg410:
Good afternoon @marcperez  and @ribeiroguilherme 
We are using the WebComponent 5.2.0 version and we've seen that when user double-cliks the "PayButton" it creates duplicate payments with the credit card Component (if it doesn't need an additional step like 3ds2 for example) Do you think this "setStatus" bug is the cause? and with this code will be solved our problem?

ribeiroguilherme:
@isaacmg410 I believe that this PR might be related to your problem. If the user is able to click twice on PayButton, that means that the setStatus isn't setting the status to 'loading' as expected. 

It depends how you are implementing it though. There are different ways to block the button: you can be using the property  `setStatusAutomatically`, or handling the status value by yourself on `onSubmit` like showed [here](https://github.com/Adyen/adyen-web/issues/1331)

PR #1407, opened 2021-12-08
Adjusting text field to handle fullwidth characters

ribeiroguilherme:
## Summary

Our text field wasn't handling fullwidth characters gracefully, due to the fact that we were not using `composition` events in order to accept text inputed by an IME.

When typing Japanese for example, the IME is activated and the user combine letters and the IME suggest words. This interaction wasn't working before.

## Tested scenarios
- Unit tests , e2e-tests
- Used iOS Simulator to simulate iPhone using Japanese keyboard
- Tested on Windows 10 with Japanese keyboard

**Fixed issue**:  
COWEB-1049

### DIFF
--- a/packages/lib/src/components/internal/FormFields/InputBase.tsx
+++ b/packages/lib/src/components/internal/FormFields/InputBase.tsx
@@ -1,5 +1,5 @@
 import { h } from 'preact';
-import { useState } from 'preact/hooks';
+import { useState, useCallback, useRef } from 'preact/hooks';
 import classNames from 'classnames';
 import { convertFullToHalf } from './utils';
 import { ARIA_ERROR_SUFFIX } from '../../../core/Errors/constants';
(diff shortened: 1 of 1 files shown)


sponglord:
I like the solution - looks nice and clean.
So just to be clear - the fact that we were always doing `convertFullToHalf` in the `handleInput` handler wasn't enough?

sponglord (approved):

ribeiroguilherme:
@sponglord the problem itself was quite weird. It wasn't the `convertFullToHalf` function, but actually just by changing the `e.target.value` value it would cause the glitchy characters to appear, or duplicated characters and so on while using Japanese keyboard.

By simply doing: 
```javascript
e.target.value = e.target.value
``` 
the code was already crashing. If I am not mistaken, this `e.target.value` should be read-only, so mutating it might be the root cause of the issue 🤔 

---

About the question if we still need the `convertFullToHalf` function, I don't really know tbh. I checked the tests for that function and I saw that it was being used to handle an specific type of full width characters but not Japanese ones:
https://github.com/Adyen/adyen-web/blob/629d32b7e8119d846e30f0ec2d4f15819d2c9007/packages/lib/src/components/internal/FormFields/utils.test.ts#L7

I believe we still need to keep it to handle that type of input

sponglord:
> I believe we still need to keep it to handle that type of input

We definitely still need it - but do we need it in the `handleInput` function or can it now be moved to `handleOnCompositionEnd`?
(If there's any doubt - just keep it where it is!)

ribeiroguilherme:
@sponglord I got your point. I am not 100% sure about moving it out from there; better to keep it as it is for now!

PR #1431, opened 2021-12-22
Improving error message for ClientKey, 3DS2Fingerprint and generic Error responses

ribeiroguilherme:
## Summary
While using sessions, the misconfigured client key error is showing as generic `NETWORK_ERROR`, which makes hard for merchants to find out what is the problem. Besides that, errors for `submitThreeDS2Fingerprint` weren't reported properly.  

This PR has the following changes:
- Passing error message about misconfigured ClientKey to the http request code
- Added `errorLevel: fatal` to `submitThreeDS2Fingerprint` call, so we throw proper error in the console
- Added handler which parses error response from the backend and uses its message property if available

## Tested scenarios
- e2e tests . Got failures on the following tests: `expiryDatePolicies.optional.separate.test.js , expiryDatePolicies.hidden.separate.test.js,  expiryDatePolicies.hidden.regular.test.js, threeDS2.default.size.test.js` , which does not seem to be related to this changes
- Tested on playground making payments with card, google pay, stored card 

**Fixed issue**:  FOC-54920, COWEB-995

### DIFF
--- a/packages/lib/src/core/Services/http.ts
+++ b/packages/lib/src/core/Services/http.ts
@@ -10,8 +10,10 @@ interface HttpOptions {
     loadingContext?: string;
     method?: string;
     path: string;
-    errorLevel?: 'silent' | 'info' | 'warn' | 'error' | 'fatal';
+    errorLevel?: ErrorLevel;
 }
+
+type ErrorLevel = 'silent' | 'info' | 'warn' | 'error' | 'fatal';
 
 type AdyenErrorResponse = {
     errorCode: string;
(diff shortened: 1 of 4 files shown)


sponglord:
These e2e tests (and others) are failing now unless the imports are changed in the relevant pages files to use the es bundles for the js and css.
So you can ignore the failures (or re-run them again, changing the imports in `pages/Cars/Cards.js`.
(Once I've confirmed with @m1aw about the validity of these imports (he still has doubts they pull in the _latest_ code) I'll make a PR to fix them

packages/lib/src/core/Services/http.ts
```diff
@@ -60,4 +94,7 @@ function handleFetchError(message: string, level: string) {
 }
 
 export const httpGet = (options: HttpOptions, data?): Promise<any> => http({ ...options, method: 'GET' }, data);
```
sponglord, line 96: I guess you could add the same generics `<T>` to `httpGet` as you did for `httpPost`

sponglord (approved):

packages/lib/src/core/Services/http.ts
ribeiroguilherme: Done!

PR #1441, opened 2022-01-06
Add support for French MealVoucher payment method

m1aw:
## Summary

The objective to this pull request is to add MealVoucher as a supported payment method. For this I extend the Gift Card component. A new "GiftcardFields" was created to allow injecting a new "Fields" type component.

This is mainly because there is a very big change in the fields layout for meal voucher. It could have been achieved with lots of very custom CSS to flexbox everything and change the order of the fields, but I felt this would have been the cleaner approach. 

## Tested scenarios

Not yet.

**Fixed issue**:


sponglord:
I like the approach - seems very elegant.
@marcperez - do you have any comments on the use of the Preact `h` function in the render function in `GiftcardComponent`?

I have made a couple of changes and then tested this branch with a local version of securedFields that accepts expiryDate as a valid gift card field and it all works as expected.

packages/lib/src/components/Giftcard/components/GiftcardComponent.tsx
```diff
@@ -107,55 +111,17 @@ class Giftcard extends Component<GiftcardComponentProps> {
...
-                                            'adyen-checkout__input--error': sfpState.errors.encryptedSecurityCode,
-                                            'adyen-checkout__input--focus': focusedElement === 'encryptedSecurityCode'
-                                        })}
-                                    />
-                                </Field>
-                            )}
-                        </div>
-                    )}
+                    render={({ setRootNode, setFocusOn }, sfpState) =>
+                        h(this.props.fieldsLayoutComponent, {
```
marcperez, line 115: Hey @m1aw, maybe a fragment (instead of calling `h`) would make this one a bit easier to understand, what do you think?

m1aw: Fixed

packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -29,8 +29,7 @@ export class GiftcardElement extends UIElement {
             paymentMethod: {
                 type: this.constructor['type'],
                 brand: this.props.brand,
-                encryptedCardNumber: this.state.data?.encryptedCardNumber,
-                encryptedSecurityCode: this.state.data?.encryptedSecurityCode
+                ...this.state.data
```
marcperez, line 32: We might want to conditionally add the `encryptedExpiryDate` (and leave the previous ones) to prevent exposing incorrect data in case anything ends up accidentally in state.data in the future.

sponglord: Well we shouldn't be _accidentally_ doing anything! (and this is how the `Card` comp works).

But if you do want to be more specific then it would be `encryptedExpiryMonth` & `encryptedExpiryYear` - since securedFields breaks the date down into these 2 properties

packages/lib/src/components/Giftcard/components/types.ts
```diff
@@ -0,0 +1,20 @@
+import Language from '../../../language/Language';
+
+export type GiftcardFieldsProps = {
+    setRootNode: (input: HTMLElement) => void;
+    i18n: Language;
+    pinRequired: boolean;
+    sfpState: any; //TODO
```
ribeiroguilherme, line 7: If I am not mistaken, we have the type of this property defined as `SFPState` on `SFP/types.ts` file.

Shall we add the missing types here too?

ribeiroguilherme:
Is it worth to add some e2e tests for this PM?  It could be done in another ticket - I'm just wondering if that would help us in the future

packages/lib/src/components/Giftcard/components/GiftcardComponent.tsx
```diff
@@ -19,6 +17,10 @@ interface GiftcardComponentProps {
     amount: PaymentAmount;
     showPayButton?: boolean;
     payButton: (config) => any;
+
+    pinRequired: boolean;
+    expiryDateRequired?: boolean;
+    fieldsLayoutComponent: any; //TODO
```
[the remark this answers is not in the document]
m1aw: Fixed

packages/lib/src/components/Giftcard/components/types.ts
m1aw: Fixed

m1aw:
Not sure exactly what to test, it should be mainly covered by Giftcard, but I think we can follow up this on another ticket.

PR #1449, opened 2022-01-14
Chore/splitting card input code

sponglord:
## Summary
Refactoring `CardInput.tsx`
This module was getting too long - so some of the code has been split out to other modules/utilities

## Tested scenarios
All e2e tests pass


packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -1,47 +1,43 @@
...
+import { InstallmentsObj } from './components/Installments/Installments';
 
-function CardInput(props: CardInputProps) {
+const CardInput: FunctionalComponent<CardInputProps> = props => {
     const sfp = useRef(null);
     const billingAddressRef = useRef(null);
     const isValidating = useRef(false);
 
+    const cardInputRef = useRef<CardInputRef>({});
+    props.setComponentRef(cardInputRef.current);
```
ribeiroguilherme, line 31: Is it intentional to leave this `props.setComponentRef(cardInputRef.current);` outside of an effect? As it is right now, I expect it to set the component ref multiple times without need

packages/lib/src/components/Card/components/CardInput/components/Installments/GetInstallmentsComp.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const getInstallmentsComp = ({ amount, brand, installmentOptions, handleInstallments, showAmountsInInstallments }) => (
+    <Installments
+        amount={amount}
+        brand={brand}
+        installmentOptions={installmentOptions}
+        onChange={handleInstallments}
+        type={showAmountsInInstallments ? 'amount' : 'months'}
+    />
+);
```
ribeiroguilherme, line 12: Out of curiosity, why not just render the `<Installments />` on both `CardFieldsWrapper` and `StoredCardFieldsWrapper` component instead of creating this helper function to render it?

packages/lib/src/components/Card/components/CardInput/components/StoredCardFieldsWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const StoredCardFieldsWrapper = ({
+    sfpState,
+    setFocusOn,
+    cvcPolicy,
+    focusedElement,
+    hasInstallments,
+    handleInstallments,
+    showAmountsInInstallments,
+    ...props
+}) => {
```
ribeiroguilherme, line 15: It would be nice to define a proper interface for the `Props` of these new components, narrowing down the types for the components. 

I don't know how much work that is though; and how that could be reused accross this wrapper, the Card wrapper, and any other Card component that share these props.

ribeiroguilherme (approved):

packages/lib/src/components/Card/components/CardInput/components/Installments/GetInstallmentsComp.tsx
sponglord: I was just porting something that was already there. But you're right we don't really gain anything from that util now we've split the storedCard and card components into separate files - so I'll remove the util.

packages/lib/src/components/Card/components/CardInput/CardInput.tsx
sponglord: I wasn't quite sure what to do here. `useMemo` seemed like overkill for such a minor operation. So I've found another way to ensure this is only called once

packages/lib/src/components/Card/components/CardInput/components/StoredCardFieldsWrapper.tsx
sponglord: I agree. I've always thought it's very _opaque_ just sending through all the props - it's not clear what is actually needed further down the line in the card/securedField components and what is just unnecessary "baggage".

I tried, some time ago, to list all the props in the `SFPProps` interface in `/internal/SecuredFields/defaultProps.ts` - and there are a lot of them (and this list must now be out of date). 

But I don't know if trying to isolate them all and pass them through, individually, to the card related components is worth it or not? Will it just add length to the file that I've just managed to shave off? Should we just embrace the spread operator and let it work its opaque magic?

It might be worth looking into... in another PR.

ribeiroguilherme (approved):

PR #1458, opened 2022-01-21
Error handling for 3DS fingerprint submit request

ribeiroguilherme:
## Summary
`callSubmit3DS2Fingerprint` wasn't handling exceptions throw by the http service, leading to an endless spinner.

With the following change, it will propagate the exception to the `UIElement#handleError` which will set the status of the component to 'ready'.

## Tested scenarios
- When error happens on `callSubmit3DS2Fingerprint`, the status of the Drop-in is set to 'ready', making it usable again

**Fixed issue**:  COWEB-995

### DIFF
--- a/packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
+++ b/packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
@@ -1,6 +1,7 @@
 import { httpPost } from '../../core/Services/http';
 import { pick } from '../internal/SecuredFields/utils';
 import { ThreeDS2FingerprintResponse } from './types';
+import AdyenCheckoutError from '../../core/Errors/AdyenCheckoutError';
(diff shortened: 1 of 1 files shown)


sponglord (approved):

packages/lib/src/components/ThreeDS2/callSubmit3DS2Fingerprint.ts
```diff
@@ -14,38 +15,43 @@ export default function callSubmit3DS2Fingerprint({ data }): void {
             errorLevel: 'fatal'
         },
         {
-            ...data
+            ...data,
+            paymentData: data.paymentData + 'x'
```
sponglord, line 19: What is this for?

ribeiroguilherme: Fixed

PR #1460, opened 2022-01-24
Fix broken url

ThermScissorpunch:
## Tested scenarios

**Fixed issue**:


marcperez (approved):

marcperez:
Good catch @ThermScissorpunch, thanks for the fix! 👍

PR #1466, opened 2022-01-26
AmazonPay Recurring Payments

m1aw:
## Summary

This ticket adds support for recurring payments via AmazonPay. For this we add support for 2 new properties, `chargePermissionType` and `recurringMetadata`.

## Tested scenarios

- A payment was made on TEST in `Recurring` mode
- A payment was made on TEST with `OneTime` mode

**Fixed issue**:

### DIFF
--- a/packages/playground/src/pages/Wallets/Wallets.js
+++ b/packages/playground/src/pages/Wallets/Wallets.js
@@ -30,13 +30,13 @@ getPaymentMethods({ amount, shopperLocale }).then(async paymentMethodsResponse =
     const step = urlSearchParams.get('step');
 
     const chargeOptions = {
-        chargePermissionType: 'Recurring',
-        recurringMetadata: {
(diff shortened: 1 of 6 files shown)


packages/playground/src/pages/Wallets/Wallets.js
```diff
@@ -26,13 +26,22 @@ getPaymentMethods({ amount, shopperLocale }).then(async paymentMethodsResponse =
     const amazonCheckoutSessionId = urlSearchParams.get('amazonCheckoutSessionId');
     const step = urlSearchParams.get('step');
 
+    const chargeOptions = {
+        // chargePermissionType: 'Recurring',
```
sponglord, line 30: Are we just keeping this commented out code here to show the potential shape of the object?
Can we do that with a type instead?

sponglord (approved):

PR #1469, opened 2022-01-27
Chore/specify SFP props

sponglord:
## Summary
Specifically pass the props that SecuredFieldsProvider actually needs rather than just spreading all the props from the parent.
A previous PR (#1449) applied this for the `CardInput` - this PR does the same for `SecuredFieldsInput` and `ACHInput`

## Tested scenarios
All tests, unit & e2e, pass


packages/lib/src/components/Card/Card.tsx
```diff
@@ -33,7 +33,7 @@ export class CardElement extends UIElement<CardElementProps> {
             hasCVC: !((props.brand && props.brand === 'bcmc') || props.hideCVC),
             // billingAddressRequired only available for non-stored cards
             billingAddressRequired: props.storedPaymentMethodId ? false : props.billingAddressRequired,
-            ...(props.brands && !props.groupTypes && { groupTypes: props.brands }),
+            // ...(props.brands && !props.groupTypes && { groupTypes: props.brands }),
```
ribeiroguilherme, line 36: Can we delete this comment? Commented code might cause confusion in the future

packages/lib/src/components/Card/components/CardInput/defaultProps.ts
```diff
@@ -10,10 +10,10 @@ export default {
     hasHolderName: false,
     holderNameRequired: false,
     enableStoreDetails: false,
-    hideCVC: false,
+    // hideCVC: false,
     hasCVC: true,
-    hasStoreDetails: false,
-    storedDetails: null,
+    // hasStoreDetails: false,
+    // storedDetails: null,
```
ribeiroguilherme, line 16: Can we delete these commented default props?

packages/lib/src/components/internal/SecuredFields/SFP/types.ts
```diff
@@ -0,0 +1,24 @@
+import { CVCPolicyType, DatePolicyType } from '../lib/types';
+import { AddressData } from '../../../../types';
+
+export interface SFPState {
+    status?: string;
+    brand?: string;
+    errors?: object;
+    valid: object;
+    data: object;
```
ribeiroguilherme, line 9: Is it possible to add proper type instead of `object` for these properties? Or is it kinda complex/dynamic objects?

ribeiroguilherme (approved):
Great PR! Good to have proper typed components, so we know what properties to expect and also should make it easier to refactor/debug in the future. 🙌

packages/lib/src/components/Card/Card.tsx
sponglord: Funny - I'm not seeing this in my file...

packages/lib/src/components/Card/components/CardInput/defaultProps.ts
sponglord: yup

packages/lib/src/components/internal/SecuredFields/SFP/types.ts
sponglord: They are complex-ish objects. I've typed one in another PR so I'll port that over. The rest I'll deal with in a subsequent PR

packages/lib/src/components/Card/Card.tsx
sponglord: I'll try recommitting it

PR #1475, opened 2022-02-03
Feature/handle pan length

sponglord:
## Summary
A `panLength` prop has been added to the `/binLookup` response.
This should have the effect of enabling us to jump focus from the PAN when the card number becomes valid.
We use the existing `autoFocus` config prop to turn this possibility on and off. (This prop is already used to control whether we jump focus from the date fields to the cvc.)
Because the credit card form can have variable, configurable, fields - logic has been added so we know to which field focus should be moved. (This logic is detailed in the YouTrack ticket, COWEB-1058)
For full functionality we will need securedFields version 3.9.0

## Tested scenarios
Added new e2e tests for receiving panLength value from binLookup and ensuring focus is sent to correct field at correct time.
All tests, unit & e2e, new & existing, pass.

**Fixed issue**:  COWEB-1058


packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
```diff
@@ -34,6 +38,49 @@ export default parent => {
             ).then((data: BinLookupResponseRaw) => {
                 // If response is the one we were waiting for...
                 if (data?.requestId === currentRequestId) {
+                    if (process.env.NODE_ENV === 'development') {
```
ribeiroguilherme, line 41: Is it intended to leave this code-block here? Seems like it was used for dev purposes

ribeiroguilherme (approved):

packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
sponglord: Yeah, it is deliberate.
It's a way, when you're developing, to intercept and mock the binLookup response.
I'm continually having to copy & paste this block of code into `triggerBinLookup` - so I thought I'd leave it there, to only be included when building in dev mode, so devs can easily test binLookup. For a production build it will be excluded

PR #1448, opened 2022-01-14
Postal Code validation

ribeiroguilherme:
## Summary
Adding validation for Postal code input for Address component.
- Validations were ported from KYC project. Therefore the files `validate.formats.ts `, `validate.ts` and `validator-utils.ts ` were copied over.
- Adjusted `Address.tsx` to use  the `addressFormatters` and also to pass the property `maxlength` to `FieldContainer`
- Adjusted `FieldContainer` to better handle the error messages: now it can receive an error message object which contains the translation key and values, so interpolation can be used while creating the error message

## Tested scenarios
- Added e2e tests for Address - Postal code validation
- Adjusted useForm unit tests to handle the format function

**Fixed issue**:  COWEB-1022

PR #1476, opened 2022-02-04
Feature/configurable giftcard names

sponglord:
## Summary
Currently we allow merchants to edit the logos for gift cards based on brand using the `icon` property in the `brandsConfiguration` object within `paymentMethodsConfiguration`.
Some merchants also want to edit the names for giftcards using this same logic.
This PR enables that with a `brandsConfiguration` object that looks like this:
```
            giftcard: {
                brandsConfiguration: {
                    svs: {
                        icon: 'https://mymerchant.com/icons/customIcon.svg',
                        name: 'custom gift card name'
                    }
                }
            }
        }
```

## Tested scenarios
Added new e2e test to see that custom name and logo are read from `brandsConfiguration`

**Fixed issue**:  DEV-51927

### DIFF
--- a/packages/e2e/tests/giftcards/brandsConfiguration/giftcard.dropin.brandsConfiguration.test.js
+++ b/packages/e2e/tests/giftcards/brandsConfiguration/giftcard.dropin.brandsConfiguration.test.js
@@ -1,0 +1,25 @@
+import DropinPage from '../../_models/Dropin.page';
+import GiftCardPage from '../../_models/GiftCardComponent.page';
+
+const dropinPage = new DropinPage({
+    components: {
+        giftcard: new GiftCardPage('.adyen-checkout__payment-method--giftcard')
+    }
+});
+
+fixture`Testing Giftcard with brandsConfiguration in Dropin`
+    .beforeEach(async t => {
+        await t.navigateTo(dropinPage.pageUrl);
+    })
+    .clientScripts('./brandsConfiguration.clientScripts.js');
+
+test.only('#1 Check Giftcard comp receives custom name and icon from brandsConfiguration object', async t => {
+    // Wait for el to appear in DOM
+    await dropinPage.giftcard.pmHolder();
+
+    // Custom name
+    await t.expect(dropinPage.giftcard.nameSpan.innerText).eql('Gifty mcGiftface');
+
+    // Custom card icon
+    await t.expect(dropinPage.giftcard.brandingIcon.getAttribute('src')).contains('mc.svg');
+});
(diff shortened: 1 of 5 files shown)


packages/e2e/tests/giftcards/brandsConfiguration/giftcard.dropin.brandsConfiguration.test.js
```diff
@@ -0,0 +1,25 @@
+import DropinPage from '../../_models/Dropin.page';
+import GiftCardPage from '../../_models/GiftCardComponent.page';
+
+const dropinPage = new DropinPage({
+    components: {
+        giftcard: new GiftCardPage('.adyen-checkout__payment-method--giftcard')
```
ribeiroguilherme, line 6: Minor thing - it seems weird to instantiate `GiftCardPage` inside `DropinPage` .

IMO it makes more sense to create a model like `GiftCard.component.js` which isn't a Page (does not extend from `BasePage`), but represents the Component. I did something similar [here](https://github.com/Adyen/adyen-web/pull/1448/files#diff-87008f4e9d7bbcdf1bb6aeed8beda9baf914bf3781a3b734e2a7109903ba6aeeR3-R4) for Address component.

We can brainstorm which way makes more sense and align that, so we can have a clear definition of page/model architecture for our E2E tests and write those following the same principles.

packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -48,7 +48,7 @@ export class GiftcardElement extends UIElement {
     }
 
     get displayName() {
-        return this.props.name;
+        return this.props.brandsConfiguration[this.props.brand]?.name || this.props.name;
```
ribeiroguilherme, line 51: Don't we need to add the typescript definition for this new property/configuration?

ribeiroguilherme (approved):

packages/e2e/tests/giftcards/brandsConfiguration/giftcard.dropin.brandsConfiguration.test.js
sponglord: I seem to recall it was your idea originally to do it like this...? (re. bancontact.visa.test.js using the CardCompoentPage)
But we can talk about it later.
And the point here is that the GiftCard comp that we need to test is the one that is found _inside_ the Dropin. When it's standalone it's already up to the merchant can display the name for the comp.

ribeiroguilherme: > I seem to recall it was your idea originally to do it like this...? (re. bancontact.visa.test.js using the CardCompoentPage)
But we can talk about it later.

Yeap! My point here is that might make sense to create isolated `GiftCard.component.js` (without extending `BasePage`) and pass it to `DropinPage` as you did. Just trying to make clear the separation between page and reusable models/components 👍 

On `bancontact.visa.test.js` I missed this part of passing the CardCompoentPage to DropinPage


> And the point here is that the GiftCard comp that we need to test is the one that is found inside the Dropin. When it's standalone it's already up to the merchant to display the name for the comp.

Got it!

packages/lib/src/components/Giftcard/Giftcard.tsx
ribeiroguilherme: 👍

PR #1484, opened 2022-02-08
Restoring missing detectedBrands prop to error object

sponglord:
## Summary
The change with how we handle errors in v5 meant that a prop, `detectedBrands` was no longer being propagated when `/binLookup` failed.
In keeping with previous solutions to this problem we now store that prop so it can be added to the stored error object and so can be accessed when required.
This is of particular importance for the Custom card component.

## Tested scenarios
All test pass.
`detectedBrands` can be accessed (in the `onChange` callback) and passed into other merchant defined handlers e.g. `onError` handlers for the Custom card component.

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/SFP/types.ts
+++ b/packages/lib/src/components/internal/SecuredFields/SFP/types.ts
@@ -11,7 +11,7 @@ export interface SFPState {
     isSfpValid?: boolean;
     autoCompleteName?: string;
     billingAddress?: AddressData;
-    hasUnsupportedCard?: string[];
+    detectedUnsupportedCardsArray?: string[];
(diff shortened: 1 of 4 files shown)


packages/lib/src/components/internal/SecuredFields/SFP/types.ts
```diff
@@ -11,7 +11,7 @@ export interface SFPState {
     isSfpValid?: boolean;
     autoCompleteName?: string;
     billingAddress?: AddressData;
-    hasUnsupportedCard?: boolean;
```
ribeiroguilherme, line 14: Just a bit weird the naming of the variable. It made more sense when it was a `boolean` type :)

ribeiroguilherme (approved):

PR #1487, opened 2022-02-09
Card: removing 'installments' property if shopper selects one time payment

ribeiroguilherme:
## Summary

In case the installments is configured for Card, and shopper selects 'One time payment' or  'Installment: 1x', we shouldn't  add the `installments` property to the payload. 

## Tested scenarios

Added e2e tests

**Fixed issue**: COWEB-1070

### DIFF
--- a/packages/e2e/tests/_models/Installments.component.js
+++ b/packages/e2e/tests/_models/Installments.component.js
@@ -1,0 +1,24 @@
+import { t, Selector } from 'testcafe';
+
+export default class InstallmentsComponent {
+    constructor(baseEl = 'div.adyen-checkout__installments') {
+        this.baseEl = Selector(baseEl).with({ timeout: 0 });
+
+        this.radioButtonOneTimeInstallment = this.baseEl.find('.adyen-checkout__radio_group__input[value=onetime]');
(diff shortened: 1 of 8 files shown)


sponglord:
Don't suppose you fancy writing an (e2e) test for this do you?
We currently don't have any for a card with installments

packages/e2e/tests/_models/Installments.component.js
```diff
@@ -0,0 +1,24 @@
+import { t, Selector } from 'testcafe';
+
+export default class InstallmentsComponent {
+    constructor(baseEl = 'div.adyen-checkout__installments') {
+        this.baseEl = Selector(baseEl).with({ timeout: 0 });
```
ribeiroguilherme, line 5: For some reason, if I don't set the `timeout: 0`, every time I invoke `selectInstallment()` or `selectRevolving()`  it takes `5s` for TestCafe to interact with the page elements.

ribeiroguilherme:
@sponglord I added some tests cases

packages/e2e/tests/_models/Installments.component.js
sponglord: I have seen this kind of thing before - testcafe can be very slow to interact with things like radio buttons and checkboxes

PR #1507, opened 2022-02-23
Showing Instant Payment Method if showPayButton is set to false

ribeiroguilherme:
## Summary
Forcing instant payment method button to show up if `showPayButton` is set to `false`. By default, it should always show up.

**Fixed issue**: COWEB-1075


sponglord:
I presume you've tested this? And all is working as it should?

sponglord (approved):

ribeiroguilherme:
I tested it on our playground using Drop-in. It works fine!

PR #1522, opened 2022-03-03
Throw an error if we notice that the environment & type of clientKey don't match

sponglord:
## Summary
Throw an error if we notice that the environment & type of clientKey don't match.
This should also help to prevent the "Invalid client key" error that securedFields can sometimes throw

## Tested scenarios
Error throw when using Live environment but Test clientKey & vice versa

### DIFF
--- a/packages/lib/src/core/core.ts
+++ b/packages/lib/src/core/core.ts
@@ -195,6 +195,12 @@ class Core {
         this.paymentMethodsResponse = new PaymentMethodsResponse(this.options.paymentMethodsResponse ?? this.options.paymentMethods, this.options);
         delete this.options.paymentMethods;
 
+        // Check for clientKey/environment mismatch
+        const clientKeyType = this.options.clientKey.substr(0, 4);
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/Card/types.ts
```diff
@@ -100,9 +101,35 @@ export interface CardElementProps extends UIElementProps {
...
+    region?: string;
+
+    // Common to GooglePay & ApplePay
+    merchantName?: string;
+
+    // Common to GooglePay & AmazonPay
+    merchantId?: string;
+
+    // Paypal
+    intent?: string;
```
marcperez, line 124: Maybe we should extract this ones to either a global or more specific type

sponglord: This was really just an early attempt to clarify what a `CardConfiguration` object looks like.
If you think it needs more work I'll move it to another PR - since I was just tacking it on to this "error message" PR

marcperez (approved):

PR #1529, opened 2022-03-09
Card - Showing brands beneath CardNumber field

ribeiroguilherme:
## Summary
Placing available brands of Card component beneath the Card number field. On this PR:
- Added property `showBrandsInCard` to Card configuration object which handles if the available brands show beneath the card number field or not. By default, they do not show

For Drop-in, once the `showBrandsInCard` is set to `true`:
- Payment Method Item it will render maximum of 4 brand images - if there are more than 4, then UI will show 3 brand images and the label saying how much is left. 
- As soon as the shopper expands the Payment Method, the brands disappear from Payment Method Item and they all show up underneath the Card Number field

For Component, once the `showBrandsInCard` is set to `true`:
- Available brands will show up underneath the Card Number field

## Tested scenarios
- Added e2e tests
- Added unit tests

**Fixed issue**:  COWEB-1077


sponglord:
I think we agreed that the default for the card comp should be true (on by default) - double check with Arjen

Also, I think the `CardFields` component is now defunct (replaced by `StoredCardFieldsWrapper` & `CardFieldsWrapper`). My bad, I didn't remove it in the last `CardInput` refactor.
Can you check and delete it, plz?

sponglord (approved):

PR #1538, opened 2022-03-17
Allow ApplePay's onPaymentAuthorized event to complete with merchant provided status and customized errors.

alinmbb:
The OnPaymentAuthorized event on ApplePay can complete with the [ApplePayAuthorizationResult](https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymentauthorizationresult) object that allows custom error messages targeted for specific fields using [ApplePayError](https://developer.apple.com/documentation/apple_pay_on_the_web/applepayerror) objects. 

Currently Adyen is only passing positive or negative status towards ApplePay without any way of customizing negative outcomes.

## Summary

**Fixed issue**:  This PR passes along the response from `onPaymentAuthorized` promise towards the `ApplePaysession.completePayment`

### DIFF
--- a/packages/lib/src/components/ApplePay/ApplePayService.ts
+++ b/packages/lib/src/components/ApplePay/ApplePayService.ts
@@ -66,8 +66,8 @@ class ApplePayService {
      */
     onpaymentauthorized(event: ApplePayJS.ApplePayPaymentAuthorizedEvent, onPaymentAuthorized) {
         return new Promise((resolve, reject) => onPaymentAuthorized(resolve, reject, event))
-            .then(() => {
-                this.session.completePayment(ApplePaySession.STATUS_SUCCESS);
+            .then(response => {
+                this.session.completePayment(response);
             })
             .catch(() => {
                 this.session.completePayment(ApplePaySession.STATUS_FAILURE);


packages/lib/src/components/ApplePay/ApplePayService.ts
```diff
@@ -66,8 +66,8 @@ class ApplePayService {
      */
     onpaymentauthorized(event: ApplePayJS.ApplePayPaymentAuthorizedEvent, onPaymentAuthorized) {
         return new Promise((resolve, reject) => onPaymentAuthorized(resolve, reject, event))
-            .then(() => {
-                this.session.completePayment(ApplePaySession.STATUS_SUCCESS);
+            .then(response => {
+                this.session.completePayment(response);
             })
             .catch(() => {
                 this.session.completePayment(ApplePaySession.STATUS_FAILURE);
```
ribeiroguilherme, line 73: If we are talking about passing the customized errors property, wouldn't make more sense for them to be passed in the catch block?

alinmbb: Using the catch block can also work. As long as we are giving the possibility - in any way- to send a customized result. 

But I am not sure, would it not be also a breaking chang if we are using the catch block instead ?

alinmbb:
Is there anything I can modify to address the typescript type error ? I am not sure where it would be

packages/lib/src/components/ApplePay/ApplePayService.ts
ribeiroguilherme: I just asked to brainstorm a bit where to handle the errors - but yea, it will be a breaking change if implemented as you proposed

ribeiroguilherme:
> Is there anything I can modify to address the typescript type error ? I am not sure where it would be

https://github.com/Adyen/adyen-web/runs/5601030699?check_suite_focus=true

`response` is of type `unknown` which is not valid to be passed to the `completePayment` function. So the type needs to be explicitly set. Next to it, the change needs to be backwards compatible

ribeiroguilherme:
Going to close this PR as the proposed changes isn't backwards compatible. Feel free to reach out if you find an alternative solution to that. We can always re-open this PR.

Regardless of this, I added to our backlog this improvement 👍

PR #1541, opened 2022-03-22
useLabelElement=false can be used on Field to wrap everything in a div

sponglord:
## Summary
When specifying a Field component you can now use the `useLabelElement` prop to specify whether the `Field` should wrap its content in a `<label>` element (usual, default, behaviour) or not.

The reason for this is that `RadioGroup` elements are exceptions to the usual fields in that they provide their own `<label>` elements. Without this change we end up with labels inside labels - which is not allowed according to the [spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) ("no descendant label elements").

This change could be timely considering the prospect of upcoming accessibility audits.
It also aligns `Field.tsx` with how it works in KYC (with an eye on the sharedComponents scenario)

## Tested scenarios
`PersonalDetails` comp, which uses the `RadioGroup`, no longer leads to a  label-inside-a-label in the markup
All tests pass (except the "availableBrands" e2e tests - which are failing for other reasons)

### DIFF
--- a/packages/lib/src/components/internal/PersonalDetails/PersonalDetails.tsx
+++ b/packages/lib/src/components/internal/PersonalDetails/PersonalDetails.tsx
@@ -82,7 +82,7 @@ export default function PersonalDetails(props: PersonalDetailsProps) {
             )}
 
             {requiredFields.includes('gender') && (
-                <Field errorMessage={!!errors.gender} classNameModifiers={['gender']} name={'gender'}>
+                <Field errorMessage={!!errors.gender} classNameModifiers={['gender']} name={'gender'} useLabelElement={false}>
                     {renderFormField('radio', {
                         i18n,
                         name: generateFieldName('gender'),
--- a/packages/lib/src/components/Card/components/CardInput/components/CardNumber.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/components/CardNumber.tsx
@@ -25,10 +25,10 @@ export default function CardNumber(props: CardNumberProps) {
             onFocusField={() => onFocusField(ENCRYPTED_CARD_NUMBER)}
             errorMessage={error && i18n.get(error)}
             isValid={isValid}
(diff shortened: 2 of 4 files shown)


packages/lib/src/components/internal/PersonalDetails/PersonalDetails.tsx
```diff
@@ -82,7 +82,7 @@ export default function PersonalDetails(props: PersonalDetailsProps) {
             )}
 
             {requiredFields.includes('gender') && (
-                <Field errorMessage={!!errors.gender} classNameModifiers={['gender']} name={'gender'}>
+                <Field errorMessage={!!errors.gender} classNameModifiers={['gender']} name={'gender'} useLabelElement={false}>
                     {renderFormField('radio', {
                         i18n,
```
ribeiroguilherme, line 87: Wouldn't be better to analyze the component child at the `<Field>` component implementation, and if it is a Radio element, then we decide there how to proceed (label or div)?

Something like:
```js
const isRadio = toChildArray(children).some((child: ComponentChild) => child.type === RadioGroup);
```
Then we do not need to always set the `useLabelElement` prop when using `radio` input. Having to always do this can lead to inconsistency as we can forget to set this property.

ribeiroguilherme (approved):

packages/lib/src/components/internal/PersonalDetails/PersonalDetails.tsx
sponglord: > ...we can forget to set this property

You could say this is true of _any_ property.

Also then `Field` needs to know of the existence of a thing called `RadioGroup` (and any other component types that need a `div` rather than a `label`. KYC, for example, has other such components).

I do see what you're saying however - I'll bear it in mind, I think what you suggest might be something that can be implemented in the "shared-components" version of this Field comp (which has a project specific wrapper around it, where knowledge of things like `RadioGroup` might make more sense

PR #1544, opened 2022-03-25
Button - Updating secondary button CSS

ribeiroguilherme:
## Summary
- Added prop 'variant' to `Button` which is used to set the variant of the button instead of setting the variants by  using true/false (Example:`<Button  secondary={true} action={true}  ghost={true} ... />`
- Updated secondary variant CSS to match according to our designs
- The 'old' secondary variant now is called 'action' variant
- Updated Components which were using the 'old' secondary variant to use the 'action' variant

## Tested scenarios
- Added unit tests checking classnames

**Fixed issue**:   CODS-1031

### DIFF
--- a/packages/lib/src/components/internal/Button/Button.tsx
+++ b/packages/lib/src/components/internal/Button/Button.tsx
@@ -8,9 +8,9 @@ import { ButtonProps, ButtonState } from './types';
 class Button extends Component<ButtonProps, ButtonState> {
     public static defaultProps = {
         status: 'default',
+        variant: 'primary',
         disabled: false,
         label: '',
-        secondary: false,
         inline: false,
         target: '_self',
         onClick: () => {}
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/internal/Button/Button.tsx
```diff
@@ -50,7 +50,7 @@ class Button extends Component<ButtonProps, ButtonState> {
             loading: <Spinner size="medium" />,
             redirect: (
                 <span className="adyen-checkout__button__content">
-                    <Spinner size="small" inline />
+                    <Spinner size="medium" inline />
```
sponglord, line 53: Does the design dictate a medium size spinner now?

sponglord (approved):

packages/lib/src/components/internal/Button/Button.tsx
ribeiroguilherme: Yes - it is the one that is being used on our current designs

PR #1548, opened 2022-03-29
Add APSE (Asia Pacific South East) environment

astiskala:
## Summary

Checkoutshopper is now available in Singapore. This adds the ability to specify that environment.

**Fixed issue**: IS-6195


ribeiroguilherme (changes requested):
`src/components/GooglePay/utils.ts` `resolveEnvironment()` method also needs to get updated with the new environment label

astiskala:
> `src/components/GooglePay/utils.ts` `resolveEnvironment()` method also needs to get updated with the new environment label

Good spot! That's embarrassing, considering I picked that up for `live-au`. Fixed :)

ribeiroguilherme (approved):

PR #1550, opened 2022-03-30
Convert Field.tsx to use Preact hooks

sponglord:
## Summary
Converted `Field.tsx` to use Preact hooks (was previously a class based component)

## Tested scenarios
All tests, unit & e2e, pass


packages/lib/src/components/internal/FormFields/Field/Field.tsx
```diff
@@ -1,53 +1,62 @@
...
+
+    // The means by which focussed/filled is set for securedFields
+    if (propsFocused != null) setFocused(!!propsFocused);
+    if (propsFilled != null) setFilled(!!propsFilled);
+
+    // The means by which focussed/filled is set for other fields - this function is passed down to them and triggered
+    const onFocusHandler = useCallback((event: h.JSX.TargetedEvent<HTMLInputElement>) => {
+        setFocused(true);
+        onFocus?.(event);
+    }, []);
```
ribeiroguilherme, line 50: `onFocus` needs to be part of the dependency array here, otherwise `onFocusHandler` won't get updated if the parent component changes the `onFocus` prop reference.

Same for the `onBlurHandler` and `renderContent` dependency arrays

packages/lib/src/components/internal/FormFields/Field/Field.tsx
```diff
@@ -116,84 +124,71 @@ class Field extends Component<FieldProps, FieldState> {
...
             </Fragment>
         );
-    }
-
-    render({
-        name,
-        className = '',
-        classNameModifiers = [],
+    }, [
+        // Original list: name, children, errorMessage, helper, inputWrapperModifiers, isLoading, isValid, label, showValidIcon, isCollatingErrors, dir
```
ribeiroguilherme, line 128: Is it intended to leave the comment?

ribeiroguilherme, line 128: Is it intended to leave the comment?

packages/lib/src/components/internal/FormFields/Field/Field.tsx
```diff
@@ -1,53 +1,62 @@
-import { Component, cloneElement, toChildArray, h, ComponentChild, VNode, Fragment } from 'preact';
 import classNames from 'classnames';
-import './Field.scss';
```
ribeiroguilherme, line 3: Are they styles being imported somewhere else? Or are they not needed anymore?

ribeiroguilherme, line 3: Are they styles being imported somewhere else? Or are they not needed anymore?

ribeiroguilherme (approved):

packages/lib/src/components/internal/FormFields/Field/Field.tsx
sponglord: We only expect to set `onFocusHandler` and `onBlurHandler` once.
These handlers replace class based code that looked like this:
```
 onFocus(e) {
        this.setState({ focused: true }, () => {
            if (this.props.onFocus) this.props.onFocus(e);
        });
    }
```
`renderContent` has a dependency array

packages/lib/src/components/internal/FormFields/Field/Field.tsx
sponglord: no :-)

packages/lib/src/components/internal/FormFields/Field/Field.tsx
sponglord: whoops! :-)

packages/lib/src/components/internal/FormFields/Field/Field.tsx
ribeiroguilherme: Setting up these functions only once does not mean that we shouldn't add their external dependencies to the `useCallback` dependency array. 

This `props.onFocus` prop might get updated at some point from the parent component - maybe this is not the case now, but it can happen in the future. And that will lead the UI to misbehave as we will be using a stale version of the `props.onFocus` inside the `onFocusHandler`. Adding it to the dependency array makes sure that this function will be always up-to-date with its dependencies.

`renderContent` has a dependency array but it is omitting objects like `label`, `onFocusHandler`, etc. which can impact its behavior at some point

sponglord: I see you point but I would argue that the aim of this PR was to do a "one for one" replacement of existing functionality rather than future proofing it

ribeiroguilherme: The `Field` component defined as class based component does not have issues with stale props because such problem does not exist for this type of component. As it gets migrated to functional component, it has.  That is why I mentioned to use the dependency array.

In my view we would not be future proofing it in this particular case. We would be making it compatible with the current way that it works now as class component.

PR #1552, opened 2022-03-30
Add support for Ratepay Direct Debit

m1aw:
## Summary

This adds support for Ratepay Direct Debit. Also modifies OpenInvoice to support BankDetails in the future. 
For this changes where made, IbanInput was extracted from the SEPA component. This change also affects the translations.

## Tested scenarios

* Tested payment with Ratepay Direct Debit
* Tested SEPA componenent for regression
* Tested normal Ratepay 

**Fixed issue**:


packages/lib/src/components/internal/IbanInput/IbanInput.tsx
```diff
@@ -104,14 +109,16 @@ class IbanInput extends Component<IbanInputProps, IbanInputState> {
...
         // calculate cursor's new position
         const cursor = e.target.selectionStart;
-        const previousIban = this.state.data['sepa.ibanNumber'];
+        const previousIban = this.state.data['ibanNumber'];
         const newCursorPosition = getNextCursorPosition(cursor, iban, previousIban);
 
         this.setState(
             prevState => ({
-                data: { ...prevState.data, 'sepa.ibanNumber': iban },
+                data: { ...prevState.data, 'ibanNumber': iban, 'countryCode': countryCode },
```
ribeiroguilherme, line 121: Minor thing: no need for single quotes here. And you can just pass the variable `countryCode` , you don't need to define the name and assign to it if they have the same name

packages/lib/src/components/internal/IbanInput/utils.ts
```diff
@@ -183,3 +183,9 @@ export const getNextCursorPosition = (cursor, iban, previousIban) => {
 
     return cursor;
 };
+
+/**
+ * @param electronicFormatIban -
+ * @returns countryCode string
+ */
+export const getCountryCode = electronicFormatIban => electronicFormatIban.slice(0, 2);
```
ribeiroguilherme, line 191: `electronicFormatIban` has type `any`. It is better if we can define proper types whenever is possible. This allow us to later on force more strict typescript rules and have well typed codebase

packages/lib/src/components/internal/OpenInvoice/types.ts
```diff
@@ -6,6 +6,13 @@ export interface OpenInvoiceVisibility {
     personalDetails?: FieldsetVisibility;
     billingAddress?: FieldsetVisibility;
     deliveryAddress?: FieldsetVisibility;
+    bankAccount?: FieldsetVisibility;
+}
+
+export interface BankDetailsSchema {
+    countryCode?: string,
+    iban?: any,
```
ribeiroguilherme, line 14: shouldn't be `string` instead of `any`?

packages/e2e/tests/openInvoices/afterpay/ratepay.test.js
```diff
@@ -0,0 +1,42 @@
...
+        .typeText('.ratepay-direct-field .adyen-checkout__input--telephoneNumber', '612345678')
+        .typeText('.ratepay-direct-field .adyen-checkout__iban-input__owner-name', 'A. Schneider')
+        .typeText('.ratepay-direct-field .adyen-checkout__iban-input__iban-number', 'DE87123456781234567890')
+        .typeText('.ratepay-direct-field .adyen-checkout__input--street', mockAddressGermany.street)
+        .typeText('.ratepay-direct-field .adyen-checkout__input--houseNumberOrName', mockAddressGermany.houseNumberOrName)
+        .typeText('.ratepay-direct-field .adyen-checkout__input--city', mockAddressGermany.city)
+        .typeText('.ratepay-direct-field .adyen-checkout__input--postalCode', mockAddressGermany.postalCode)
+        // Can't use the checkboxLabelGender trick to speed up the click 'cos this label contains a link - so use a Selector with a timeout
+        .click(payButton)
+        .wait(5000);
```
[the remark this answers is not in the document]
m1aw: I will update this in the future, for now the Test is not actually working since we still have to solve filling in the `dateOfBirth`

PR #1554, opened 2022-03-31
Atome payment method

ribeiroguilherme:
## Summary
- Added Atome payment method
- Started creating the interface of `OpenInvoiceContainer` as there was no interface defined before. I added the props used by this PR, but the remaining props need to be added afterwards.
- Introduced two new props on `OpenInvoiceContainer`: 
   - `billingAddressRequiredFields`: similar to the current `personalDetailsRequiredFields`, but for billing address
   - `billingAddressSpecification`: The billing address form needs to be customized in certain way that some of the default fields won't be shown for this payment method. Therefore the UI needs to be adjusted accordingly to not have empty space gaps. This property allows to pass an `specification` which dictates which fields to show and their position in the UI

## Tested scenarios
- Tested manually the whole payment flow using Atome as Component and as part of Drop-in
- Added unit tests to make sure only required fields are being passed to the OpenInvoice component

**Fixed issue**: COWEB-1051

### DIFF
--- a/packages/lib/src/components/helpers/OpenInvoiceContainer/OpenInvoiceContainer.tsx
+++ b/packages/lib/src/components/helpers/OpenInvoiceContainer/OpenInvoiceContainer.tsx
@@ -2,8 +2,19 @@ import { h } from 'preact';
 import UIElement from '../../UIElement';
 import OpenInvoice from '../../internal/OpenInvoice';
 import CoreProvider from '../../../core/Context/CoreProvider';
+import { UIElementProps } from '../../types';
+import { AddressSpecifications } from '../../internal/Address/types';
 
-export default class OpenInvoiceContainer extends UIElement {
+interface OpenInvoiceElementProps extends UIElementProps {
+    consentCheckboxLabel?: h.JSX.Element;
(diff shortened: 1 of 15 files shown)


packages/lib/src/components/helpers/OpenInvoiceContainer/OpenInvoiceContainer.tsx
```diff
@@ -2,8 +2,19 @@ import { h } from 'preact';
...
+import { UIElementProps } from '../../types';
+import { AddressSpecifications } from '../../internal/Address/types';
 
-export default class OpenInvoiceContainer extends UIElement {
+interface OpenInvoiceElementProps extends UIElementProps {
+    consentCheckboxLabel?: h.JSX.Element;
+    billingAddressRequiredFields?: string[];
+    billingAddressSpecification?: AddressSpecifications;
+
+    // TODO: add other props for OpenInvoiceElement
```
m1aw, line 13: I have made a similar thing for RatePay, this looks more complete though. We should check when merging both to master.

m1aw (approved):

PR #1565, opened 2022-04-06
Fixed e2e test that was failing because of secondary button styling changes

sponglord:
## Summary
Fixed e2e test that was failing because of secondary button styling changes

## Tested scenarios
Test passes


ribeiroguilherme (approved):
Thanks for fixing it

PR #1581, opened 2022-04-14
Fix/disabling i os arrow keys

sponglord:
## Summary
The keyboard arrow buttons on iOS devices do not work for the card component's securedFields (web integration).
From iOS/Safari ± `v14.5` the keyboard arrow buttons appear, activated, when a securedField input is clicked into - but do nothing. (Prior to this Safari <= `v14.3` the arrow buttons never activate).

The reason the keyboard arrow buttons don't work is linked to a [change made in Safari 13](https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes):
> “Changed the behavior for third party iframes to prevent them automatically navigating the page.”

In other words the change prevents 3rd party (i.e. cross-origin) iframes from programmatically focusing before the user makes a direct interaction with the frame.
This restriction unfortunately also applies to the arrow keys on the "soft" iOS keyboard (a test shows that if the iframes are the same origin as the hosting site then the arrow keys behave as expected).

Since this is a situation created by Apple that we have no control over - we resolved that if we can't achieve keyboard navigation in iOS browsers then we shouldn't even appear to offer it i.e. that we would, if possible, turn off these arrow keys.

There is no direct and easy way to access the keyboard arrow buttons (i.e. through JS or CSS) _but_ it was discovered that if all other inputs in the Card Payment method are disabled when an input is clicked into then the arrow keys do not activate.

So the code in this PR (which is _only applied when iOS is detected_) detects `touch` events on Card Payment method inputs and directs all the inputs, including securedFields, _that are not the one that has just been interacted with_ to disable themselves.

This PR relies on a new version of SecuredFields (`v3.10.0`) which is now (as of 14/04/22) on Test.

## Tested scenarios
Huge amounts of manual testing across multiple iOS devices (via Browserstack)
Still TODO - create (unit) tests, if possible.

**Fixed issue**:  Relates to issue COWEB-1020


packages/lib/src/components/internal/Address/Address.tsx
```diff
@@ -98,6 +98,7 @@ export default function Address(props: AddressProps) {
                 specifications={specifications}
                 maxlength={getMaxLengthByFieldAndCountry(countrySpecificFormatters, fieldName, data.country, true)}
                 trimOnBlur={true}
+                disabled={disablingTrigger && disablingTrigger !== fieldName}
```
ribeiroguilherme, line 101: Minor thing and personal opinion here: 

I find this new property `disablingTrigger` on `Address` confusing to understand.  It is used only by the Card component, and can be hard for other devs that dig into the `Address` to understand what it implies. I am wondering if we can change it in a way which can be used by other components later on as well and can be easier to understand.

From what I understand, we need to disable all fields except the one that is currently focused. What if we provide a property like `enabledFields: string[]` which can be set one or more enabled fields - if nothing is set, then all fields are enabled by default.

In this way, we would pass to address the disablingTrigger:
```js
<Address
  {...props}
  enabledFields={disablingTrigger && [disablingTrigger]}
/>
```

I think doing this way it is more clear what is happening and might be reused in the future on other payment methods. Let me know your thoughts!

ribeiroguilherme (approved):

packages/lib/src/components/internal/Address/Address.tsx
sponglord: I understand your concern. I found naming this particular property and some of the other props & functions required for this PR quite difficult.

My issue with your suggested `enabledFields` prop is that it then requires a knowledge form the parent of _all_ the fields in a component i.e it has to list them all in order for them to be enabled.

What if we had a `disabledFields` prop - which is more in line with what I'm already doing, namely just stating the field to exclude (rather than listing all the fields that need to be _included_)?

ribeiroguilherme: Fair enough - I totally agree with the `disabledFields` property 👍  But how are you going to implement on this case which you have only the field name that should be enabled? You will need to set all other fields that needs to be disabled, right? Is it possible here?

sponglord: I'll figure it out. :-)
For now I'm merging this - 'cos I want this new functionality in place.
I'll put the new, more minor, change in a new PR

PR #1588, opened 2022-04-25
Fix broken URL

TimvdLippe:
## Summary
The previous URL 404's and based on https://docs.adyen.com/online-payments/web-drop-in/ and the sidemenu, I think this is the correct URL.


ribeiroguilherme (approved):

ribeiroguilherme:
Nice catch @TimvdLippe . Thanks!

PR #1593, opened 2022-05-04
Add onOrderCancel to sessions

m1aw:
## Summary

This objective of this PR is to add onOrderCancel handling when using the dropin with sessions. 

## Tested scenarios

* Checked if orderCancel is trigged correctly with sessions
  * Checked if it handles endpoint failure
* Checked if onOrderCancel is trigger in the manual implementation
* Checked if onOrderCancel is not passed the button is not shown

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Dropin/components/DropinComponent.tsx
+++ b/packages/lib/src/components/Dropin/components/DropinComponent.tsx
@@ -41,6 +41,8 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
                 }
             }
         );
+
+        this.onOrderCancel = this.getOnOrderCancel();
     };
 
     public setStatus = (status: UIElementStatus, props: DropinStatusProps = {}) => {
@@ -93,12 +95,12 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
     }
 
     /**
-     * onOrderCancelBuilder decides which onOrderCancel logic should be used, manual or sessions
+     * getOnOrderCancel decides which onOrderCancel logic should be used, manual or sessions
      */
-    private onOrderCancelBuilder = () => {
+    private getOnOrderCancel = () => {
         if (this.props.onOrderCancel) {
             return (data: onOrderCancelData) => {
-                return this.props.onOrderCancel(data);
+                this.props.onOrderCancel(data);
             };
         }
         if (this.props.session) {
@@ -110,6 +112,8 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
         }
         return null;
     };
+
+    private onOrderCancel: (data: onOrderCancelData) => void;
 
     render(props, { elements, instantPaymentElements, status, activePaymentMethod, cachedPaymentMethods }) {
         const isLoading = status.type === 'loading';
@@ -140,7 +144,7 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
                                 cachedPaymentMethods={cachedPaymentMethods}
                                 order={this.props.order}
                                 orderStatus={this.state.orderStatus}
-                                onOrderCancel={this.onOrderCancelBuilder()}
+                                onOrderCancel={this.onOrderCancel}
                                 onSelect={this.handleOnSelectPaymentMethod}
                                 openFirstPaymentMethod={this.props.openFirstPaymentMethod}
                                 openFirstStoredPaymentMethod={this.props.openFirstStoredPaymentMethod}
--- a/packages/lib/src/components/Dropin/components/DropinComponent.tsx
+++ b/packages/lib/src/components/Dropin/components/DropinComponent.tsx
@@ -5,7 +5,6 @@ import getOrderStatus from '../../../core/Services/order-status';
 import { DropinComponentProps, DropinComponentState, DropinStatusProps } from '../types';
 import './DropinComponent.scss';
 import { UIElementStatus } from '../../types';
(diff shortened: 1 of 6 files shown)


packages/playground/src/pages/Dropin/session.js
```diff
@@ -28,6 +28,9 @@ export async function initSession() {
         onError: (error, component) => {
             console.error(error.message, component);
         },
+        // onOrderCancel: data => {
+        //     console.error('onOrderCancel', data);
+        // },
```
ribeiroguilherme, line 33: Can be removed?

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -92,6 +92,25 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
         this.setState({ activePaymentMethod: null });
     }
 
+    private onOrderCancel = data => {
```
ribeiroguilherme, line 95: Can the `data` type and return type of the function be defined here?

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -92,6 +92,25 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
+        if (this.props.onOrderCancel) {
+            return new Promise(() => {
+                return this.props.onOrderCancel(data);
+            });
```
ribeiroguilherme, line 99: Why do we need to do `new Promise` here? Can't just call the `this.props.onOrderCancel` ?

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -92,6 +92,25 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
+        }
+        if (this.props.session) {
+            return this.props.session
+                .cancelOrder(data)
+                .then(data => {
+                    console.log('response', data);
```
ribeiroguilherme, line 105: `log` should be removed

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -92,6 +92,25 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
+                    this.props._parentInstance.update({ order: null });
+                })
+                .catch(error => {
+                    this.setStatus(error?.message || 'error');
+                });
```
ribeiroguilherme, line 110: Did you manage to simulate an error response and trigger this flow?

ribeiroguilherme:
Do we have any sort of E2E tests for the cancel flow? It would be great to cover this use-case if there are no tests.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
m1aw: I typed the return functions, I can also add types to this one, but they where being inferred correctly

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -92,6 +92,25 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
         this.setState({ activePaymentMethod: null });
     }
 
+    /**
+     * onOrderCancelBuilder decides which onOrderCancel logic should be used, manual or sessions
+     */
+    private onOrderCancelBuilder = () => {
```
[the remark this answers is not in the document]
ribeiroguilherme: I think it is a bit overcomplicating to create a builder to return the function that will cancel the gitftcard order. It is also not very efficient as every render you inject a new function into the `<PaymentMethodList />`  when `this.onOrderCancelBuilder()` is called, therefore forcing the render.

Why not create normal handler function to do that and pass it as callback? Is there any problem doing that?

Ex:
```ts
private handleOnOrderCancel = async (data: OrderData): Promise<void> => {
  if (this.props.session) {
    try {
       await this.props.session.cancelOrder(data);
       this.props._parentInstance.update({order: null});
      } catch (error) {
        this.setStatus(error?.message || 'error');
      }
      return;
    }    
    if (this.props.onOrderCancel) this.props.onOrderCancel(data);
}
```

ribeiroguilherme (approved):

packages/lib/src/components/Dropin/components/DropinComponent.tsx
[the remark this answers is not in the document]
m1aw: I agree with the performance concern. I can move the function call to the constructor and that would make more sense. 
That being said, the logic in that function is different, because for several reasons:
1. We are always passing an function to the component PaymentMethodList and that makes the button will always be active even when there is not a valid implementation of the function (ie: there is no `onOrderCancel` and not in sessions)
2. Sessions in that case take precedence over onOrderCancel, which I don't think makes sense since the merchant can still chose to implement his own logic for onOrderCancel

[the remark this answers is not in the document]
ribeiroguilherme: 1. A button being active just because it has a callback sounds a bit off. Might be worth to revisit this logic and see if it can be improved
2. Can merchant that is using sessions implement his own `onOrderCancel`? I am not sure if that makes sense from a merchant perspective. From my understanding the session endpoint is called internally by the SDK only. I don't see the use-case for a merchant providing its own `onOrderCancel`. Let me know what I am missing

PR #1601, opened 2022-05-13
Feature/mb way international phone number

sponglord:
## Summary
Adding an international phone prefix dropdown to the current phone number field to allow international phone numbers in the MBWay component.

Previously this component just had a single input into which the user was expected to type a Portuguese mobile phone number (either with or without the international "+351" prefix).

Now the component consists of 2 parts:
- an international country code selector (shows flags + international phone prefix + 2 character country code)
- an input into which to type the domestic part of the phone number

## Tested scenarios
All existing unit and e2e tests pass

**Fixed issue**:  COWEB-835 
Also resolves this [issue for the magento plugin](https://github.com/Adyen/adyen-magento2/issues/1438)

### DIFF
--- a/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
+++ b/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
@@ -48,10 +48,6 @@ function MBWayInput(props: MBWayInputProps) {
     }, []);
 
     const onChange = ({ data, valid, errors, isValid }) => {
-        console.log('\n### MBWayInput::onChange:: data=', data);
-        // console.log('### MBWayInput::onChange:: valid=', valid);
-        // console.log('### MBWayInput::onChange:: errors=', errors);
-        // console.log('### MBWayInput::onChange:: isValid=', isValid);
         props.onChange({ data, valid, errors, isValid });
     };
 
--- a/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
+++ b/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
@@ -1,13 +1,12 @@
 import { h } from 'preact';
-import { useState, useLayoutEffect, useRef } from 'preact/hooks';
+import { useState, useRef } from 'preact/hooks';
 import useCoreContext from '../../../../core/Context/useCoreContext';
 import { MBWayInputProps } from './types';
 import './MBWayInput.scss';
-import getDataset from '../../../../core/Services/get-dataset';
-import { DataSet } from '../../../../core/Services/data-set';
 import PhoneInput from '../../../internal/PhoneInputNew';
 import AdyenCheckoutError from '../../../../core/Errors/AdyenCheckoutError';
 import LoadingWrapper from '../../../internal/LoadingWrapper';
+import usePhonePrefixes from '../../../internal/PhoneInputNew/usePhonePrefixes';
 
 function MBWayInput(props: MBWayInputProps) {
     const { i18n, loadingContext } = useCoreContext();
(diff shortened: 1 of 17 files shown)


packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -1,69 +1,74 @@
...
+    const { allowedCountries = [] } = props;
+
+    const [phonePrefixes, setPhonePrefixes] = useState<DataSet>([]);
+    const [status, setStatus] = useState<string>('loading');
 
     this.setStatus = setStatus;
-    this.showValidation = triggerValidation;
+    this.showValidation = phoneInputRef?.current?.triggerValidation;
+
+    useLayoutEffect(() => {
```
ribeiroguilherme, line 25: Minor thing: seems that this whole logic here could be extracted to a custom hook (ex: `usePhonePrefixes`):

```javascript
const { isLoadingPrefixes, phonePrefixes } = usePhonePrefixes();
```

That can declutter a bit the code :-) It will look simpler and can improve reusability

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -1,69 +1,74 @@
...
+            </div>
+        </LoadingWrapper>
     );
 }
 
 MBWayInput.defaultProps = {
-    onChange: () => {}
+    onChange: () => {},
+    phoneNumberKey: 'mobileNumber',
+    phoneNumberErrorKey: 'mobileNumber.invalid'
```
ribeiroguilherme, line 71: Do we support such customization for i18n? Is it needed?

ribeiroguilherme (approved):

packages/lib/src/components/internal/SecuredFields/lib/utilities/commonUtils.ts
```diff
@@ -222,4 +222,4 @@ export function partial(...args) {
 }
 
 // Not null or undefined or only spaces
-export const isEmpty = input => !!(input == null || /^[\s]*$/.test(input));
+// export const isEmpty = input => !!(input == null || /^[\s]*$/.test(input));
```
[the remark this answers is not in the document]
sponglord: yes

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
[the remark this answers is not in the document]
sponglord: I don't know exactly! 
@pabloai determined that `useLayoutEffect` was necessary in other components that call `getDataset`

sponglord: What are you talking about extracting? The phone prefix loading? I'm curious as to why the `this.showValidation` is also highlighted?

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
sponglord: I think so - we might want to say "telephone number (invalid)" or "phone number (invalid)" or "mobile phone number (invalid)". 
It offers flexibility depending on what exactly the component is being used for and what the Ux-ers want to show for labels/error messages

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
ribeiroguilherme: Sorry I wasn't clear - I didn't intend to highlight the `this.showValidation` part 😅 

I meant the whole logic inside the `useLayouEffect`: fetching the data from the API and handling the loading status of this action. This could be extracted to a custom hook.

sponglord: Good idea - have created a custom hook

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -16,46 +15,23 @@ function MBWayInput(props: MBWayInputProps) {
...
-            })
-            .catch(error => {
-                props.onError(new AdyenCheckoutError('ERROR', error));
-                setPhonePrefixes([]);
-                setStatus('ready');
-            });
-    }, []);
+    if (error) {
+        props.onError(new AdyenCheckoutError('ERROR', error));
+    }
```
ribeiroguilherme, line 27: This part should be placed as you did before: inside the try-catch block. Otherwise, every time that component re-render, we will trigger the `onError` (if there is an error).

You can pass the error to the hook and do it there on its try-catch block:
```javascript
const { loadingStatus: prefixLoadingStatus, phonePrefixes } = usePhonePrefixes({ allowedCountries, loadingContext, handleError: props.onError });`
```

An alternative without passing the handler to the custom hook is to react using useEffect:

```javascript
useEffect(() => {
   if (error) props.onError(...);
}, [error]);
```

ribeiroguilherme (approved):

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
sponglord: alright

PR #1614, opened 2022-06-01
Click to Pay - Initialization mechanism

ribeiroguilherme:
## Summary
This PR is part of the umbrella branch `feature/click-to-pay`, meaning that it is not the final PR.

It targets the mechanism to load the schemes SDK efficiently within the Card payment method. There are UI components created to test the whole flow, but they are going to be adjusted/refactored in the upcoming PR's to match with the designs.

There are console.logs and TODO comments spread around, but they will be removed the later on.

Summary of how it works:
- `constructor` of Card payment method detects that there is a CtP configuration available, then it creates the `ClickToPayService` and initializes it
- `clickToPayService` is passed to the `ClickToPayProvider` which determines if should render the CtP Component or not.
- As soon as the `initialize()`  is called in the `Card` `constructor()`, the  `ClickToPayService` loads the required 3rd party SDK's and verify if the user has cookies.:
  - If cookies are found, then the service fetches the user cards and keep them in memory to show on the Card component.
  - If cookies are not found, the user is not recognized during the initialization. Then an identification flow needs to be done. The identification flow starts when the Card component is fully rendered - the 3rd party SDK sends the OTP message to the shopper. Shopper enters the OTP, then get his Cards, then he is able to perform the checkout
- If the shopper is not enrolled in the CtP system (email isn't valid), then we render only the normal Card fields

## Tested scenarios
- Tested that as soon as Drop-in is rendered, the CtP is initialized and shopper cards are loaded in memory
- Tested rendering the CtP as part of the Card Component
- Tested sending OTP to the shopper and OTP validation part
- Tested the checkout flow through the 3rd party SDK receiving the approved transaction


packages/lib/src/components/Card/Card.tsx
```diff
@@ -171,18 +192,33 @@ export class CardElement extends UIElement<CardElementProps> {
...
+                <ClickToPayProvider clickToPayService={this.clickToPayService}>
+                    <ClickToPayContext.Consumer>
+                        {({ ctpState }) => (
+                            <div>
+                                {ctpState !== CtpState.NotAvailable && ctpState !== CtpState.Idle && (
+                                    <Fragment>
+                                        <ClickToPayComponent onSubmit={this.handleClickToPaySubmit} />
+                                        <ContentSeparator classNames={['adyen-checkout-ctp__separator']} label="Or enter card details manually" />
+                                    </Fragment>
+                                )}
```
sponglord, line 204: It would be neater if this could all be handled inside one top-level component e.g. `<ClickToPayWrapper>`
Inside this would be the `<ClickToPayProvider>` and the `<ClickToPayContext.Consumer>`

The `<CardInput>` could either be accessed through the `children` prop or else provided directly through a 'getter' function e.g. `<ClickToPayWrapper cardInput={cardInputField}>`
``` 
const cardInputField = (
        <CardInput
            setComponentRef={this.setComponentRef}
            {...this.props}
            {...this.state}
            onChange={this.setState}
            ...
        />
    );
```

sponglord (approved):

packages/lib/src/components/Card/components/ClickToPay/components/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }, [triggerValidation]);
+
+    useImperativeHandle(ref, () => ({ validateInput }));
+
+    useEffect(() => {
+        props.onChange({ data, valid, errors, isValid });
+    }, [data, valid, errors]);
+
+    return (
+        <Field label={i18n.get('One time code')} errorMessage={props.errorCode || !!errors.otp} classNameModifiers={['otp']}>
```
sponglord, line 44: Fields like this should have a `name` prop e.g. `name={'oneTimeCode}`
This will ensure that the generated `label` then has a `for` attribute that will link to a unique `id `generated on the enclosed text field

packages/lib/src/components/Card/Card.tsx
ribeiroguilherme: Yes, I totally agree that it could be neater 👍  

I am working in another branch which has more changes related to this part - it has kinda the final state of how the 'render' in the `Card.tsx` file will look like . So I propose for us to iterate on this again on my next PR so we can come up with the best approach :-)

packages/lib/src/components/Card/components/ClickToPay/components/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
ribeiroguilherme: I will add on the follow up PR 👍

sponglord: ok

PR #1626, opened 2022-06-10
UPI Payment method

ribeiroguilherme:
## Summary

Adding support for the new txVariant `upi`.

Summary:
- `upi` is the parent txVariant which has two sub variants which are always configured: `upi_qr` and `upi_collect`
- This payment method has two flows which the shopper opt to pay: 
  -  Click on `Generate QR code` which does the /payments with the `upi_qr` , and then triggers the QR code flow
  -  Enter the VPA value, then click on `Continue`, which does the /payments with the `upi_collect` and then triggers the Await flow

## Tested scenarios
- Manual test authorizing the payment using the BillDesk simulator

**Fixed issue**: COWEB-891


packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -0,0 +1,108 @@
...
+                        payButton={this.payButton}
+                        onChange={this.setState}
+                        onSubmit={this.submit}
+                        onGenerateQrCodeClick={this.handleGenerateQrCodeClick}
+                    />
+                );
+        }
+    }
+
+    public render(): h.JSX.Element {
```
m1aw, line 98: Do we need to type the render function?
Also why h.JSX.Element? Is this what we should use? I did something for i18n an was using Preact.VNode

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -0,0 +1,108 @@
...
+ * 'upi' tx variant is the parent one.
+ * 'upi_collect' and 'upi_qr' are the sub variants which are submitted according to the shopper interaction.
+ */
+const UPI_COLLECT = 'upi_collect';
+const UPI_QR = 'upi_qr';
+
+class UPI extends UIElement {
+    public static type = 'upi';
+
+    constructor(props: any) {
```
m1aw, line 20: I know the UPI component has props, but could we have them here too?
I feel like this makes more sense since it's where it will interface with outside of the library

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+                status: isSubmittingAndFlowIsQrCode ? 'default' : status,
+                disabled: status === 'loading'
+            })}
+
+            <ContentSeparator label={i18n.get('qrCodeOrApp')} />
+
+            <Button
+                icon={getImage({ loadingContext: loadingContext, imageFolder: 'components/' })('qr_dark')}
+                status={isSubmittingAndFlowIsQrCode ? 'loading' : 'default'}
+                disabled={status === 'loading'}
```
m1aw, line 51: Can we set status to loading instead? Why create a new disabled state?
It's a bit confusing, since that is already the internal logic of the button, but we also use the status?

ribeiroguilherme: This payment method differs a bit from the norm, as there are two buttons to submit the payment instead of one. The shopper can choose to 'Generate QR code', or to submit the VPA field .

So, when the shopper clicks to 'Generate QR code' , this button must show a loading state (as the user interacted with it), and the other button (the one that submits the VPA) must be disabled, and vice-versa. That is why the 'disabled' state is being used here. Otherwise both buttons would show 'loading' state, which isn't a good UX.

packages/lib/src/components/UPI/UPI.tsx
ribeiroguilherme: There is actually no specific props for the UPI component.

This `props` variable is here as it needs to be passed when calling `super` inside the constructor, as the UiElement constructor expects that.

packages/lib/src/components/UPI/UPI.tsx
ribeiroguilherme: We don't necessarily need to type as we aren't using strict mode. In reality we don't need to type anything 😅 . But I don't see why not type it - if we turn on the strict typescript mode at some point, we will have less issues with untyped code.. and I believe that makes sense to push for more strict typing in the future

About `h.JSX.Element` vs `preact.VNode` , I am not fully sure which one is recommended to go for. I see few usages of both in the codebase. I used the `h.JSX.Element` as it is the type inferred by the Typescript compiler in any 'render' function that returns JSX.

packages/lib/src/components/UPI/UPI.tsx
m1aw: I see, and then UIElement actually doesn't have a typed props too. Because otherwise we could just extend that.

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
m1aw: I see, button already handles the logic of "loading", I had only read the logic of <PayButton>.
Maybe we should pass all this logic in the future to <Button> it would make it a cleaner interface.

packages/lib/src/components/UPI/UPI.tsx
m1aw: Well, I don't think render is going to cause issues in strict mode since we extend Component for the component.
I did search and I did not find consensus too. I will change some of my code to be `JSX.Element` since it seems to be a bit more framework agnostic.

packages/lib/src/components/internal/PayButton/PayButton.tsx
```diff
@@ -9,10 +9,12 @@ export interface PayButtonProps {
      * Class name modifiers will be used as: `adyen-checkout__image--${modifier}`
      */
     classNameModifiers?: string[];
-
     label?: string;
     amount: PaymentAmount;
     status?: string;
+    disabled?: boolean;
+    icon?: string;
```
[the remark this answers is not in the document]
m1aw: Yeah, that was my initial confusion but then I understood.

m1aw (approved):

PR #1627, opened 2022-06-10
3DS2 handling challenge error

sponglord:
## Summary
A scenario exists, described in when it is not possible to retrieve a `transStatus` result for a 3DS2 challenge.
In this scenario an "error" object is sent (with `errorCode` & `errorDescription` properties).
The frontend 3DS2 component should treat this response as a valid outcome and handle it accordingly (show error icon, call merchant defined `onError` callback)

## Tested scenarios

**Fixed issue**:  COWEB-1113

### DIFF
--- a/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
+++ b/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
@@ -4,6 +4,10 @@ import { createChallengeResolveData, handleErrorCode, prepareChallengeData, crea
 import { PrepareChallenge3DS2Props, PrepareChallenge3DS2State } from './types';
 import { ThreeDS2FlowObject } from '../../types';
 import '../../ThreeDS2.scss';
+import Img from '../../../internal/Img';
+import { getImageUrl } from '../../../../utils/get-image';
+import './challenge.scss';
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
```diff
@@ -48,18 +58,49 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
...
             );
         }
 
+        if (this.state.status === 'error') {
+            return (
+                <div className="adyen-checkout__threeds2-challenge-error">
+                    <Img
+                        className="adyen-checkout__status__icon adyen-checkout__status__icon--error"
+                        src={getImageUrl({ loadingContext: this.props.loadingContext, imageFolder: 'components/' })('error')}
+                        alt={this.props.i18n.get('error.message.unknown')}
```
m1aw, line 97: We don't need alt here since we already use `error.message.unkown` in the text message.

m1aw (approved):

packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
sponglord: Good point!

PR #1628, opened 2022-06-10
Loading securedFields file that now uses JWE

sponglord:
## Summary
Loading securedFields version (`4.0.0`) that now uses JWE, rather than CSE, to encrypt data.

For Edge <= 18 & IE11 we fall back to a bulkier, "Compat", version of the file with the necessary polyfills.

## Tested scenarios
Tested in (Browserstack):
- Windows 10, Chrome 45
- Windows 10, Firefox 40
- Safari 11.1
- Edge 15
- IE11

Tested on desktop:
-  latest versions of Chrome(102.0.5005), Firefox (90.0.2), Safari (15.5) on Mac OS Monterey

All tests, unit & e2e, pass for both the "regular" and the "compat" version

**Fixed issue**:  COWEB-1062

**Update**: latest version is `4.1.0` which is smaller in size and which also restores the "event logging" i.e. the tracking of the user behaviour within the PAN & cvc fields. All tests still pass.

### DIFF
--- a/packages/e2e/app/src/pages/Cards/Cards.js
+++ b/packages/e2e/app/src/pages/Cards/Cards.js
@@ -5,6 +5,7 @@ import { amount, shopperLocale, countryCode } from '../../services/commonConfig'
 import '../../style.scss';
 
 const initCheckout = async () => {
+    // window.TextEncoder = null; // Comment in to force use of "compat" version
     window.checkout = await AdyenCheckout({
         amount,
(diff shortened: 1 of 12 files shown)


ribeiroguilherme (approved):

ribeiroguilherme:
Small note: The secured fields version is 4.1.0 and not 4.0.0 as described in the summary :-)

PR #1631, opened 2022-06-16
Adds onOrderCreated callback to support GiftCard component in sessions flow

m1aw:
## Summary

This ticket is meant to add onOrderCreated callback. This callback tries to solve the following:

> While trying to make partial payments using gift cards on checkout sessions + web components, there is no way to inform the merchant and the shopper of a successful gift card payment and the remaining amount to be paid, after a part of the payment is successfully paid using a giftcard.

We solve this by creating this new callback, it should only get triggered when an order is created, because there was not enough balance in the giftcard to cover the full amount. If the gifcard can cover the full amount we still use the normal flow of asking for confirmation.

This callback will allow merchants to retrieve the `order` object, this should have the `remainingAmount` as well as informations of the payment methods already used. This `order` object can then be used while setting up the new payment method component, together with the old session.

## Tested scenarios

* Created E2E scenario to test when the callback is trigger and when it's not
* Playground as been updated with an example 

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Giftcard/Giftcard.tsx
+++ b/packages/lib/src/components/Giftcard/Giftcard.tsx
@@ -5,7 +5,7 @@ import CoreProvider from '../../core/Context/CoreProvider';
 import getImage from '../../utils/get-image';
 import PayButton from '../internal/PayButton';
 import AdyenCheckoutError from '../../core/Errors/AdyenCheckoutError';
-import { PaymentAmount } from '../../types';
+import { Order, PaymentAmount } from '../../types';
 
 export class GiftcardElement extends UIElement {
     public static type = 'giftcard';
@@ -71,6 +71,15 @@ export class GiftcardElement extends UIElement {
 
         if (this.props.session) {
             return this.props.session.createOrder();
+        }
+    };
+
+    protected handleOrder = (order: Order) => {
+        this.elementRef._parentInstance.update({ order });
+        if (this.props.session && this.props.onOrderCreated) {
+            return new Promise((resolve, reject) => {
+                return this.props.onOrderCreated(resolve, reject, order);
+            });
         }
     };
 
--- a/packages/playground/src/pages/GiftCards/GiftCards.js
+++ b/packages/playground/src/pages/GiftCards/GiftCards.js
@@ -1,8 +1,8 @@
 import AdyenCheckout from '@adyen/adyen-web';
 import '@adyen/adyen-web/dist/es/adyen.css';
 import { handleChange, handleSubmit } from '../../handlers';
-import { amount, shopperLocale, countryCode } from '../../config/commonConfig';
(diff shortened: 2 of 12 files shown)


packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -74,6 +74,15 @@ export class GiftcardElement extends UIElement {
         }
     };
 
+    protected handleOrder = (order: Order) => {
+        this.elementRef._parentInstance.update({ order });
```
sponglord, line 78: Could just call `super.handleOrder(order)` here

packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -74,6 +74,15 @@ export class GiftcardElement extends UIElement {
...
+        if (this.props.session && this.props.onOrderCreated) {
+            return new Promise((resolve, reject) => {
```
sponglord, line 80: Does this _need_ to return a Promise? 
The function it overrides `UIElement.handleOrder` doesn't.

packages/lib/src/components/Giftcard/Giftcard.tsx
m1aw: Because there is no super.handleOrder. I did think of this, but since we are assigning an anonymous function to the object state, we can't use super, because props that get overwritten.

packages/lib/src/components/Giftcard/Giftcard.tsx
m1aw: It has to do with keeping the same API has we have for other session callback like `onOrderRequest` and `onBalanceCheck`.

sponglord: ok

sponglord (approved):

packages/playground/src/pages/GiftCards/GiftCards.js
```diff
@@ -44,4 +44,65 @@ import '../../style.scss';
...
+            onOrderCreated: async (resolve, reject, data) => {
+                await afterGiftCard(data);
+                resolve();
+            }
+        })
+        .mount('#giftcard-session-container');
+
+
+    const afterGiftCard = async (order) => {
+        const checkout = await AdyenCheckout({
```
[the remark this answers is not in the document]
m1aw: Yes, that is definitely true. I was trying stuff out and left the example that way. There was no reason to do so. I have updated the example.

packages/lib/src/components/Giftcard/Giftcard.tsx
m1aw: More specifically doing it that way results in the following:
```
Cannot read properties of undefined (reading 'call') 
```

PR #1641, opened 2022-06-29
Click to Pay - Login and OTP 

ribeiroguilherme:
## Summary

This PR covers the UI components for the Login and the OTP feature, including all asynchronous code done by these components.

It targets the shopper identification/login flow into the CtP system, which covers:
- In case the shopper has valid email which exists in the CtP system, then it shows the OTP component
- In case the shopper clicks to be logged out, then it shows the 'Login' component
- In case the shopper has cookies, it shows directly the 'Card' list
- In case the shopper is not recognized as part of the CtP system, then we show the default Card component without the Login flow.

Some highlights of the PR:
- `<ClickToPayWrapper />` component which helps to orchestrate how the UI for the Card payment method renders:
  -  There are scenarios where the Click to Pay is the primary input and the Card is the secondary, therefore the Pay button can have different colors, and that can be toggled depending on the shopper interaction
  - There are scenarios where we don't show the Card input - we show a button instead which the shopper needs to click to see the card input
- In case the shopper wants to resend the OTP code, that can be done through `<CtPResendOtpLink />` component. After sending the OTP again, there is a count down timer of 60s 
- `<Field />` component was adjusted to support having a Component passed as prop which can be displayed on the right corner over the input field. In this PR, the 'Resend code' (`<CtPResendOtpLink />`) is being placed there

## Tested scenarios
- Manual testing:
  - Attempt to login with wrong credentials
  - Logout feature
  - Resend otp feature
  - Entering wrong otp code and showing proper error

## Note
- This PR is part of the umbrella branch `feature/click-to-pay`, meaning that it is not the final PR.


packages/lib/src/components/Card/Card.tsx
```diff
@@ -193,31 +208,9 @@ export class CardElement extends UIElement<CardElementProps> {
                 commonProps={{ isCollatingErrors: this.props.SRConfig.collateErrors }}
             >
                 <ClickToPayProvider clickToPayService={this.clickToPayService}>
```
sponglord, line 210: It's not possible to also put the `ClickToPayProvider` inside the `ClickToPayWrapper`?

ribeiroguilherme: The reason I didn't put it there is because the `ClickToPayWrapper` consumes the context values from the `ClickToPayProvider` through the `useClickToPayContext` hook . 

`ClickToPayProvider` uses the context do decide which payment method should be primary/secondary ones (Card or CTP)

ribeiroguilherme:
> Clicking on the "Not you?" link on the OTP page seemed to hang for a moment - is there much going on here that would cause a delay?

There is an asynchronous call done to the Scheme SDK to properly logout the user (they clean the cookies, but I believe that there are more things involved) - that is why it takes more time. 

Maybe I can forward the shopper to the 'Login' screen before doing the async `unbindAppInstance` (aka logout) call , so there is no delay. In this approach I'd be assuming that the logout would always work, but I believe that the odds that logout mechanism from the scheme to not work are very small.

packages/lib/src/components/Card/Card.tsx
ribeiroguilherme: > The whole idea being that the CtP stuff is as self-contained & separated as possible, with the minimum disruption on the Card.tsx markup

I totally agree. In the current situation here, the 'ClickToPayWrapper' is acting mainly as a 'Card UI manager' in a sense that it is used to decide how to render the UI: only the 'card', only the 'loader', or the 'ctp + card' - it all depends of the 'state' of the Click to Pay (if it is available, initializing, etc)

I tried to encapsulate the logic in the Wrapper, so in case we provide a standalone ClickToPay component in the future, we won't have to mix the logic about Card component. We could simply do:

```jsx
<ClickToPayProvider clickToPayService={this.props.clickToPayService}>
    <ClickToPayComponent />
</ClickToPayProvider>
```

But I want to revisit this again and see if I can improve that more. I will do it after wrapping up the checkout flow.

PR #1645, opened 2022-07-06
OnlineBanking component for CZ and SK

m1aw:
## Summary

This PR adds a new OnlineBanking component. This is an issuer list component. 

The IssuerList was modified to additionally display a disclaimer that ToS will be accepted upon continuation. This change was made to IssuerList since it might be reused for other components in the future, and seemed generic enough.

There is now also an interpolation function to help place links, or other html elements / react elements, into a string that comes from the translations files. For this we use the `%#` delimiter. 

Additionally CZ and SK currencies where added to the playground.

## Tested scenarios

It was tested that the issuer list was rendering and redirecting on the playground.

**Fixed issue**:


packages/lib/src/language/utils.ts
```diff
@@ -127,3 +128,21 @@ export const loadTranslations = async (locale: string, customTranslations: objec
...
+    // splits by regex group, it guarantees that it only splits with 2 tokens (%#)
+    const matches = translation.split(/%#(.*?)%#/gm);
+    // the map will create an array of JSX / string elements, this syntax in accepted in JSX/react to render elements
+    return matches.map((term, index) => {
+        // math to get the index of the renderFunction that should be used
+        // since we split on tokens, that means the index of the render function is half of the index of the string
+        const indexInFunctionArray = Math.floor(index / 2)
+        return index % 2 === 0 ? term : renderFunctions[indexInFunctionArray](term);
+    });
+};
```
sponglord, line 148: Just a thought: should this function test that the length of the `matches` array matches that of the `renderFunctions` array i.e that it is being passed the same number of components to render as there are place to put them?

ribeiroguilherme: Follow up on this in the PR #1682

ribeiroguilherme (approved):

PR #1652, opened 2022-07-11
Change onOrderCallback to not return promise

m1aw:
## Summary

We concluded that we don't need to use a promise for the onOrderCreated callback. This PR changes to normal function callback.

## Tested scenarios

- Ran e2e tests fro onOrderCreated
- Tested playground integration

**Fixed issue**:

### DIFF
--- a/packages/playground/src/pages/GiftCards/GiftCards.js
+++ b/packages/playground/src/pages/GiftCards/GiftCards.js
@@ -76,14 +76,14 @@ import '../../style.scss';
         .create('giftcard', {
             type: 'giftcard',
             brand: 'svs',
-            onOrderCreated: async (data) => {
-                await afterGiftCard(data);
+            onOrderCreated: (data) => {
+                afterGiftCard(data);
             }
         })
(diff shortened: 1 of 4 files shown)


packages/playground/src/pages/GiftCards/GiftCards.js
```diff
@@ -76,9 +76,8 @@ import '../../style.scss';
         .create('giftcard', {
             type: 'giftcard',
             brand: 'svs',
-            onOrderCreated: async (resolve, reject, data) => {
+            onOrderCreated: async (data) => {
```
ribeiroguilherme, line 79: Is there is a need for the `async` prefix in the function and also for the `afterGiftCard` to be async? I believe that there is no need for that here. The code is all synchronous

m1aw: Kinda just left it for testing purposes. I agree this makes it more clear.

ribeiroguilherme (approved):

PR #1660, opened 2022-07-20
Add support for installmentOptions for Sessions

Dominique2509:
## Summary
This PR covers the handling of configuration.installmentOptions returned on the internal `/setup` call of `/sessions`. 

## Tested scenarios
- When installmentOptions are provided in the `/sessions` call, they are correctly shown on the card component.
- When installmentOptions are provided both in the config and the `/sessions` call, `/sessions` takes precedence. 
- When installmentOptions are provided only in the config, they are correctly shown on the card component.

**Fixed issue**: COWEB-1122

### DIFF
--- a/packages/lib/src/core/core.ts
+++ b/packages/lib/src/core/core.ts
@@ -42,6 +42,7 @@ class Core {
                 .setupSession(this.options)
                 .then(sessionResponse => {
                     const amount = this.options.order ? this.options.order.remainingAmount : sessionResponse.amount;
+                    this.session.configuration = sessionResponse.configuration
                     this.setOptions({ ...sessionResponse, amount });
                     return this;
                 })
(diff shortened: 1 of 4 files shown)


packages/lib/src/core/core.ts
```diff
@@ -42,6 +42,7 @@ class Core {
                 .setupSession(this.options)
                 .then(sessionResponse => {
                     const amount = this.options.order ? this.options.order.remainingAmount : sessionResponse.amount;
+                    this.session.configuration = sessionResponse.configuration
```
marcperez, line 45:
```suggestion
                    this.session.configuration = sessionResponse.configuration;
```

packages/lib/src/types/index.ts
```diff
@@ -1,6 +1,7 @@
 import paymentMethods from '../components';
 import { ADDRESS_SCHEMA } from '../components/internal/Address/constants';
 import actionTypes from '../core/ProcessResponse/PaymentAction/actionTypes';
+import {InstallmentOptions} from "../components/Card/components/CardInput/components/types";
```
marcperez, line 4:
```suggestion
import { InstallmentOptions } from '../components/Card/components/CardInput/components/types';
```

packages/lib/src/core/core.ts
m1aw: Surprise that the linter didn't catch this

PR #1664, opened 2022-07-29
Click to Pay - Adjusting logos

ribeiroguilherme:
## Summary

This PR covers loading the scheme logos according to the available schemes configured by the merchant.

Highlights:
- Adjusted CtP context/provider to pass the list of available 'schemes'
- `<CtPSection />` component get the list of `schemes` that are available, and render them (before it was hardcoded)
- `<CtpCardsList />` and `<CtPSingleCard />` uses the scheme logo in case the Card doesn't have `artUri` property

## Note
- This PR is part of the umbrella branch `feature/click-to-pay`


packages/lib/src/components/Card/components/ClickToPay/components/CtPSection/CtPSection.scss
```diff
@@ -27,16 +30,16 @@
...
-.adyen-checkout-ctp__section-header-divider {
-    background-color: black;
-    width: 1px;
-    height: 15px;
-    border-radius: 1px;
+    object-fit: none;
     margin-right: 6px;
+
+    &-mc {
+        width: 27px;
```
m1aw, line 37: The logos are SVG right? Why don't we control their size in the SVG?

ribeiroguilherme: We don't have such control as the SVG's are external resources hosted in the checkoutshopper. In this case here, we are controlling size of the  html element which wraps the SVG's

m1aw (approved):

PR #1667, opened 2022-08-01
Adds Portugal Portuguese localisation file

m1aw:
## Summary

Adds Portugal Portuguese pt-PT localisation file

## Tested scenarios

**Fixed issue**:


ribeiroguilherme (commented):
Don't we need to dynamic load it [here](https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/language/locales/index.ts) ?

m1aw:
Yes we do. I forgot. Thank you!

ribeiroguilherme (approved):

PR #1679, opened 2022-08-08
Fix bug where BankTransfer comp wasn't shown if showPayButton set to false

sponglord:
…false

## Summary
Fix bug where BankTransfer comp wasn't shown if showPayButton set to false

## Tested scenarios
BankTransfer comp still shows if `showPayButton` set to false

**Fixed issue**:  COWEB-1116

### DIFF
--- a/packages/lib/src/components/BankTransfer/BankTransfer.tsx
+++ b/packages/lib/src/components/BankTransfer/BankTransfer.tsx
@@ -58,8 +58,6 @@ export class BankTransferElement extends UIElement<BankTransferProps> {
                 )}
             </CoreProvider>
         );
-
-        return null;
     }
 }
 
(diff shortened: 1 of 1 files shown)


marcperez (approved):

packages/lib/src/components/BankTransfer/BankTransfer.tsx
```diff
@@ -50,16 +50,14 @@ export class BankTransferElement extends UIElement<BankTransferProps> {
...
+                {this.props.showPayButton && (
                     <RedirectButton {...this.props} name={this.displayName} onSubmit={this.submit} payButton={this.payButton} />
-                </CoreProvider>
-            );
-        }
+                )}
+            </CoreProvider>
+        );
 
         return null;
```
marcperez, line 62: We might want to remove this one since it's unreachable (see Sonarcloud)

sponglord: sure!

PR #1681, opened 2022-08-08
Add proper typing for `AdyenCheckout` function

Spea:
## Summary

We are in the process of upgrading from `3.23.0` to `5.21.0` and were a bit surprised that there was no more typing when creating the `AdyenCheckout` object. 

Adding the typing to the function makes it easier to see what  options are actually available.

### DIFF
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -9,7 +9,7 @@ import { CoreOptions } from 'src/core/types';
 import Checkout from './core';
 /* eslint-enable */
 
-async function AdyenCheckout(props: CoreOptions) {
+async function AdyenCheckout(props: CoreOptions): Promise<Checkout> {
     const checkout = new Checkout(props);
     return await checkout.initialize();
 }
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -5,10 +5,11 @@ if (process.env.NODE_ENV === 'development') {
     // require('preact/debug');
 }
 
+import { CoreOptions } from 'src/core/types';
 import Checkout from './core';
 /* eslint-enable */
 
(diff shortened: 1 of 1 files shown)


packages/lib/src/index.ts
```diff
@@ -5,10 +5,11 @@ if (process.env.NODE_ENV === 'development') {
     // require('preact/debug');
 }
 
+import {CoreOptions} from 'src/core/types';
 import Checkout from './core';
 /* eslint-enable */
 
-async function AdyenCheckout(props) {
+async function AdyenCheckout(props: CoreOptions) {
```
ribeiroguilherme, line 12: Could you also add the return type as well?

```suggestion
async function AdyenCheckout(props: CoreOptions): Promise<Checkout> {
```

packages/lib/src/index.ts
```diff
@@ -5,10 +5,11 @@ if (process.env.NODE_ENV === 'development') {
     // require('preact/debug');
 }
 
+import {CoreOptions} from 'src/core/types';
```
ribeiroguilherme, line 8: Likely prettier is not setup on your env :)

```suggestion
import { CoreOptions } from 'src/core/types';
```

Spea: You are right, I thought it was configured as our prettier config does remove the spaces between the braces. Anyhow, fixed it.

packages/lib/src/index.ts
Spea: Totally! Added it via 75ea617

ribeiroguilherme (approved):

packages/lib/src/index.ts
ribeiroguilherme: Great, thanks!

PR #1683, opened 2022-08-10
Click to Pay - Loading config from Backend and UI adjustments

ribeiroguilherme:
## Summary

This PR covers loading the CtP configuration from backend, and then initialize the CtP components. It also includes minor CSS/UI adjustments

How the CtP initialization works in a nutshell:
- The merchant needs to explicitly set `useClickToPay` on Card configuration in order to enable it 
- `/paymentMethods` returns as part of the `scheme` payment method configuration object the CtP initialization values used by the scheme SDK's. If values are provided, then CtP is initialized, otherwise it is ignored and we show the normal Card input fields.

How merchant can pass the shopper email data:
- If using 'sessions' , when creating the session, the merchant can pass the `shopperEmail` as part of the setup call. In this way, no extra configuration is needed when creating the Component
- If using 'manual' flow, the merchant needs to provide the shopper email in the card component configuration object. Ex:
```js
    card: {
        useClickToPay: true,
        clickToPayConfiguration: {
            shopperIdentityValue: 'shopper.email@domain.com'
        }
    },
}
```

Other highlights:
- `<InputBase />` now has the `onKeyUp` callback which is optional. In this way ,we can pass callback to track which key stroke was pressed.
- `<CtpResendOtpLink />` now shows 'Code resent' for 2 seconds before starting the countdown 

## Note
- This PR is part of the umbrella branch feature/click-to-pay, meaning that it is not the final PR.


packages/lib/src/components/Card/Card.tsx
```diff
@@ -85,7 +87,12 @@ export class CardElement extends UIElement<CardElementProps> {
                 showPanel
             },
             // installmentOptions of a session should be used before falling back to the merchant configuration
-            installmentOptions: props.session?.configuration?.installmentOptions || props.installmentOptions ,
+            installmentOptions: props.session?.configuration?.installmentOptions || props.installmentOptions,
+            clickToPayConfiguration: {
+                ...props.clickToPayConfiguration,
+                shopperIdentityValue: props.clickToPayConfiguration?.shopperIdentityValue || props?._parentInstance?.options?.session?.shopperEmail,
```
sponglord, line 93: If a `shopperEmail` has come from the `/sessions` endpoint do we want the merchant to be able to overwrite it with their manual config?

sponglord (approved):

packages/lib/src/components/Card/Card.tsx
ribeiroguilherme: I thought about this when developing - I believe that if the merchant explicitly set the shopperEmail in the configuration object, then it should be overridden. 

Ideally the merchant that is integrating using `sessions` just need to pass the `shopperEmail` to the session creation, and add the `useClickToPay` flag to make it all work - it doesn't make sense to add the `shopperEmail` to the CtP config when using sessions, but if so, then we will use that value

PR #1686, opened 2022-08-12
Click to Pay

ribeiroguilherme:
## Summary

This is the umbrella branch which contains all commits/PR's related to Click to Pay. 

How the Click to Pay initialization works in a nutshell:

- The merchant account needs to be configured with the initialization values for Click to Pay SDK's. After doing that, `/paymentMethods` API response will return the Click to Pay initialization values used by the 3rd party SDK's, and the Card component will display the proper Component.
- In order to configure the UI to show it, the merchant needs to explicitly set `useClickToPay` on Card configuration to enable it. Ex:
```js
const adyenCheckout = checkout.create('card', {
    useClickToPay: true,
    ...
}).mount('#card-area');
```
Or, if using Drop-in:
```js
        card: {
            useClickToPay: true,
        },
    },
    ...
}

const adyenCheckout = checkout.create('dropin').mount('#dropin-area')
```
There are two possibilities for the shopper to use this payment: the shopper has the Click to Pay cookies stored in his browser, or the shopper has a valid email which is registered in the Click to Pay system. If none of these cases are valid, then the default Card input fields is rendered.

In order to authenticate the shopper email in the Click to Pay system, the merchant has two different ways of passing the shopper email to Adyen Checkout:
- If using 'sessions' , when creating the session, the merchant can pass the [shopperEmail](https://docs.adyen.com/api-explorer/#/CheckoutService/v68/post/sessions__reqParam_shopperEmail) as part of the setup call
- If using 'manual' flow, the merchant needs to provide the shopper email in the card component configuration object. Ex:
```js
    card: {
        useClickToPay: true,
        clickToPayConfiguration: {
            shopperIdentityValue: 'shopper.email@domain.com'
        }
    },
```
If the shopper is identified (either by cookies or by email), the Click to Pay component is kept in the screen and the shopper can proceed with the payment.


sponglord:
I presume `useClickToPay: true,` also works if you add it to the `paymentMethodsConfiguration` object on Checkout?
i.e.
```
            card: {
                useClickToPay: true,
...
}

sponglord (approved):

Issue #1697, opened 2022-08-17
[5.22.0] Invalid import in index.d.ts 


mvannes:
**Describe the bug**
Introduced in https://github.com/Adyen/adyen-web/commit/c5ecf2d10c4f282d071c0df2849abd8d2e624cd3 it seems like the import is wrong here. 
It says to import from `src/` , but looking at the index.d.ts location, `src` is not an actual location once build.
Instead, it should probably import like the Checkout does, so `./core/types/` .

This is making some typescript builds with the latest version of this package invalid.

**To Reproduce**
Update to 5.22.0 version of `@adyen/adyen-web`  and try to build with the AdyenCheckout imported into a project.

**Expected behavior**
It shouldn't error on resolving the adyen-web's internal imports.

ribeiroguilherme:
Hi @mvannes ,

Good catch! Thanks for the heads up and for the PR. We merged it into master and the fix is going to be available on our next release that should take place earlier next week. 👌  I will close this issue once we release it.

mvannes:
Thanks! Looking forward to it!

ribeiroguilherme:
Good news @mvannes - the version 5.23.0 is available with the fix 😄  Thanks for the reaching out and for the fix. Cheers!

PR #1698, opened 2022-08-17
Import from relative path instead of src/ dir for AdyenCheckout's Cor…

mvannes:
…eOptions parameter

This import breaks when creating a dist/ version for typescript.
Instead, import as a relative path like is done for the Checkout itself.

fixes https://github.com/Adyen/adyen-web/issues/1697


ribeiroguilherme (approved):

Spea:
@mvannes Thank you for fixing this. I just realized my mistake from #1681 and wanted to fix it but then saw that you already took care of it

PR #1700, opened 2022-08-19
Connecting email validation to fields

sponglord:
## Summary
The shopper email fields in the `BankTransfer` & `Boleto` components were not validating.
This PR connects those fields to the same email validation rule as used in the `PersonalDetails` component

## Tested scenarios
Fields now validate using a email regEx
All tests pass


packages/lib/src/components/BankTransfer/components/BankTransferInput/BankTransferInput.tsx
```diff
@@ -26,6 +30,8 @@ function BankTransferInput(props) {
 
     useEffect(() => {
         props.onChange({ data, errors, valid, isValid });
+        console.log('### BankTransferInput::data:: ', data);
+        console.log('### BankTransferInput::errors:: ', errors);
```
ribeiroguilherme, line 34: Should they be removed?

ribeiroguilherme (approved):

packages/lib/src/components/BankTransfer/components/BankTransferInput/BankTransferInput.tsx
sponglord: yup!

PR #1701, opened 2022-08-19
Pay button shows Croatian kuna and converted EUR amount

sponglord:
## Summary
From the ticket (COWEB-1139): 
> Add secondaryAmount config to Drop-in/Components to allow merchants to display dual prices on the Pay button (HRK to EUR migration) 

TL;DR - since Croatia is converting to the Euro they need to display 2 prices on the Pay button from 05/09/2022 to end 2023

The merchant can now pass in a `secondaryAmount` config object in the form:
```
{ 
currency: 'HRK', 
value: {convertedValue: 1 EUR = 7.53450 kuna}, 
currencyDisplay: 'code' 
}
```
This represents the secondary currency and the value (converted from the original amount).
It is up to the merchant to perform this calculation themselves.
A third property `currencyDisplay` gives the merchant _some_ control over how that converted value is then displayed (e.g. whether a currency symbol is used or whether the currency type appears before the value or after; see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#currencydisplay)

Although exactly how this secondary currency is displayed ultimately depends on the rules defined within the browser for `.toLocaleString` for the primary language (`shopperLocale`)

## Tested scenarios
When countryCode (& currency) is Croatia (HR) we show the HRK amount first and then the converted EUR amount
When countryCode (& currency) is European (NL)  we show the EUR amount first and then the converted HRK amount

Added unit tests
All tests pass

**Fixed issue**:  COWEB-1139

### DIFF
--- a/packages/lib/src/components/internal/PayButton/PayButton.tsx
+++ b/packages/lib/src/components/internal/PayButton/PayButton.tsx
@@ -11,7 +11,7 @@ export interface PayButtonProps {
     classNameModifiers?: string[];
     label?: string;
     amount: PaymentAmountExtended;
-    secondaryAmount: PaymentAmountExtended;
+    secondaryAmount?: PaymentAmountExtended;
     status?: string;
     disabled?: boolean;
     icon?: string;
--- a/packages/lib/src/components/internal/PayButton/PayButton.tsx
+++ b/packages/lib/src/components/internal/PayButton/PayButton.tsx
@@ -18,7 +18,7 @@ export interface PayButtonProps {
     onClick?(): void;
 }
 
-const payAmountLabel = (i18n: Language, amount) => {
+const payAmountLabel = (i18n: Language, amount: PaymentAmountExtended) => {
     const payStr = `${i18n.get('payButton')}`;
 
     const primaryAmount =
@@ -29,7 +29,7 @@ const payAmountLabel = (i18n: Language, amount) => {
     return `${payStr} ${primaryAmount}`;
 };
 
-const secondaryAmountLabel = (i18n: Language, secondaryAmount) => {
+const secondaryAmountLabel = (i18n: Language, secondaryAmount: PaymentAmountExtended) => {
     const convertedSecondaryAmount =
         secondaryAmount && !!secondaryAmount?.value && !!secondaryAmount?.currency
             ? i18n.amount(secondaryAmount.value, secondaryAmount.currency, { currencyDisplay: secondaryAmount.currencyDisplay || 'symbol' })
--- a/packages/lib/src/components/internal/PayButton/PayButton.tsx
+++ b/packages/lib/src/components/internal/PayButton/PayButton.tsx
@@ -3,6 +3,7 @@ import Button from '../Button';
 import useCoreContext from '../../../core/Context/useCoreContext';
 import { PaymentAmountExtended } from '../../../types';
 import Language from '../../../language/Language';
+import SecondaryButtonLabel from './components/SecondaryButtonLabel';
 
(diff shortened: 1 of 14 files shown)


packages/lib/src/components/internal/PayButton/PayButton.tsx
```diff
@@ -10,27 +10,50 @@ export interface PayButtonProps {
...
 
+    const secondaryLabel = secondaryAmount && Object.keys(secondaryAmount).length ? secondaryAmountLabel(i18n, secondaryAmount) : null;
+
     return (
         <Button
             {...props}
             disabled={props.disabled || props.status === 'loading'}
             classNameModifiers={[...classNameModifiers, 'pay']}
             label={label || defaultLabel}
+            secondaryLabel={secondaryLabel}
```
ribeiroguilherme, line 56: Just an idea: did you also consider passing `secondaryLabel` as a child component for the Button instead of a prop? 
I think that in this way we can improve the reusability of the Button, as we can pass any sort of custom Components/Values to it.

Then it would be up to the `PayButton` component to pass the secondary label with its specific CSS format, and `Button` doesn't care about that

Ex:
```jsx
<Button...> {secondaryLabel && <SecondaryAmountLabel ... /> </Button>
```

packages/lib/src/components/internal/PayButton/PayButton.tsx
```diff
@@ -10,27 +10,50 @@ export interface PayButtonProps {
...
-const PayButton = ({ amount, classNameModifiers = [], label, ...props }: PayButtonProps) => {
+    const primaryAmount =
+        !!amount?.value && !!amount?.currency
+            ? i18n.amount(amount.value, amount.currency, { currencyDisplay: amount.currencyDisplay || 'symbol' })
+            : '';
+
+    return `${payStr} ${primaryAmount}`;
+};
+
+const secondaryAmountLabel = (i18n: Language, secondaryAmount) => {
```
ribeiroguilherme, line 32: Can you add the type `PaymentAmountExtended`  for `secondaryAmount`?

packages/lib/src/components/internal/PayButton/PayButton.tsx
```diff
@@ -10,27 +10,50 @@ export interface PayButtonProps {
...
+    secondaryAmount?: PaymentAmountExtended;
     status?: string;
     disabled?: boolean;
     icon?: string;
     onClick?(): void;
 }
 
-const payAmountLabel = (i18n: Language, amount) =>
-    `${i18n.get('payButton')} ${!!amount?.value && !!amount?.currency ? i18n.amount(amount.value, amount.currency) : ''}`;
+const payAmountLabel = (i18n: Language, amount) => {
```
ribeiroguilherme, line 21: Can you add the type `PaymentAmountExtended` for `amount`?

ribeiroguilherme (approved):

packages/lib/src/components/internal/PayButton/PayButton.tsx
sponglord: can do!

packages/lib/src/components/internal/PayButton/PayButton.tsx
sponglord: can do!

packages/lib/src/components/internal/PayButton/PayButton.tsx
sponglord: You're right - it does make it more powerful... I'll make the change

Issue #1638, opened 2022-06-22
Component can not be mounted, unmounted and mounted again


OliverHolden:
We are using adyen-web 5.16.2 and we are integrating Components for Web (https://docs.adyen.com/online-payments/web-components).

We have our own wrapper component rendering the payment option title and logo (in a label belonging to a radio button) and then we wish to mount the component belonging to that Adyen payment option when that payment option gets selected, and de render the component on deselection. 

The issue comes when switching back and forth between payment options. When we try and mount the component on re-selection we are getting the "Component is already mounted." even if we unmount the component on deselection/derender of our wrapper component.

I have tracked this back to line 80 and 81 of `BaseElement`, where if `this._node` is already defined it throws the "Component is already mounted." which is the error we are catching.

![image](https://user-images.githubusercontent.com/22643984/174973000-00ea631f-a075-4a45-b461-7b7350006f93.png)
 
and then I can see in the unmount function nothing is done to clear the value of `this._node` and it seems to only derender the component rather than actually unmount it:

![image](https://user-images.githubusercontent.com/22643984/174973335-377f4c07-8bfa-4117-be06-93c5b5a62001.png)

We have tried checking the value of `component_node` and then calling `component.remount()` if it is already defined and `component.mount()` if not, however, we still have issues when our wrapper component de renders.

As a workaround, before calling `component.mount()` we are setting the `componet._node` value to null and then mounting. This then behaves as we would expect `component.mount()` followed by a `component.unmount()` then followed by a component.mount() to behave, but we are unsure of the side effects of doing this. It makes no difference if the component used is the original component or the component returned by the `.mount()` function.

Is there a method that we are missing? or a better way to achieve this desired behaviour?

sponglord:
Hi @OliverHolden - thanks for reporting this issue.

I've been able to achieve the results you are after by using the `remount` function.

So... first time round you `mount` the component: `component.mount('.card-field');`
Then you `unmount` it: `component.unmount();`
Then for 2nd and subsequent "mounts" you use `remount`: `component.remount();`

All internal references stay intact, which I've tested by mounting, unmounting & remounting a card component and then going on to make a 3DS2 payment which leads to a call to `component.handleAction` - this relies on a maintained internal reference to the component in order to render the subsequent 3DS2 components.

Perhaps you could try the above?

ribeiroguilherme:
Hi @OliverHolden ,

[...]

PR #1702, opened 2022-08-19
Allow option to nullify this._node when unmounting

sponglord:
## Summary
Merchant issues like the one mentioned (#1638), and others; which are related to repeatedly mounting, unmounting and then (re)mounting components have all led to the conclusion that it would be useful when calling `unmount` on a component to be able to also set the reference to `component._node` to `null` without the merchant having to do it manually.

This PR allows the merchant to specify `true` when unmounting a component i.e. `component.unmount(true)` and this will cause the `BaseElement` to set its reference to `this._node` to `null`.

This seems to work to fix issues related to (re)mounting a component which then fails with the error `'Component is already mounted.'`

## Tested scenarios
All tests pass

**Relates to issue**:  #1638

### DIFF
--- a/packages/lib/src/components/BaseElement.ts
+++ b/packages/lib/src/components/BaseElement.ts
@@ -126,10 +126,14 @@ class BaseElement<P extends BaseElementProps> {
 
     /**
      * Unmounts a payment element from the DOM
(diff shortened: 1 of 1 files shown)


sponglord:
Will be addressed in a new, forthcoming, PR that will enhance the `mount` function so it also covers `remount`-ing

PR #1708, opened 2022-08-23
More unit tests for PayButton re. secondaryAmount

sponglord:
## Summary
Added unit tests for the PayButton to check it is only rendering the secondaryAmount when it should

## Tested scenarios
They pass!


sponglord:
Clash of merges - I'll remake this PR

Issue #1369, opened 2021-11-17
Web component: callback for loading error


JesusNaturitas:
Hi.

I'm looking for a way to show an error message when card form fails to load on your side.

A onConfigFailure callback (opposite to existing onConfigSuccess callback) would do the trick. Is there any similar callback?

Thanks a lot!

ribeiroguilherme:
Hi @JesusNaturitas ,

Do you mean the whole card form, or just the secured fields? Can you please clarify the scenario that you are trying to achieve?

It can help If you share a code snippet/screenshots of what is happening or what you expect to happen.

Jazilch:
👋  I am also looking for something similar. I am using a custom card integration and I have seen a few times that it appears adyen.js loads successfully in the network tab but then an error occurs when configuring adyen. I had a har file sent to me from the customer and this is why I saw that the adyen script loaded correctly but the secured fields never configured and the customer wasn't able to use the form. 

I am using the `onConfigSuccess` callback to see when the payment form is correctly configured but it would be great if there was also a `onConfigFailure` callback so that when these issues occur I could get a better idea of the error to help out my customers. 

This issue has been very tricky to debug and I haven't been able to repro myself the extra error reporting would be helpful to see why the form didn't configure correctly.

Jazilch:
@ribeiroguilherme Did you have any thoughts on my example above? Not sure if this gives you enough information about a use case for this callback

sponglord:
I believe that this issue has been addressed with PR #1714

PR #1714, opened 2022-08-29
Catching securedFields load and config failures

sponglord:
## Summary
Extra functionality around detecting if securedField's iframe fails to call its 'load' event or if they fail to configure
In both cases calls the `onError` callback defined by the merchant when they configure the main Checkout instance.(https://docs.adyen.com/online-payments/web-components/optional-configuration#events)

## Tested scenarios
- Manually created scenarios forcing securedFields to fail to load and to fail to configure - saw that these situations were detected and reported as errors to the `onError` callback
- All unit & e2e tests pass

**Relates to issue:** #1369

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/SFP/SecuredFieldsProviderHandlers.ts
+++ b/packages/lib/src/components/internal/SecuredFields/SFP/SecuredFieldsProviderHandlers.ts
@@ -21,6 +21,7 @@ import {
     CbObjOnLoad
 } from '../lib/types';
 import { existy } from '../lib/utilities/commonUtils';
+import AdyenCheckoutError from '../../../../core/Errors/AdyenCheckoutError';
 
 /**
  * Emits the onLoad event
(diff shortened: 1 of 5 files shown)


ribeiroguilherme (approved):

packages/lib/src/components/internal/SecuredFields/SFP/SecuredFieldsProviderHandlers.ts
```diff
@@ -23,36 +23,42 @@ import {
...
         if (this.state.status !== 'ready') {
             // Hide the spinner
-            this.setState({ status: 'invalidOriginError' });
-
-            // Now we catch clientKey & environment mismatch in core.ts - this error is indicative of a
-            // clientKey error: the requesting domain is not in the “Allowed origins” for the clientKey in the CA
-            this.props.onError({ error: 'invalidOriginError', fieldType: 'defaultError' });
+            this.setState({ status: 'csfConfigFailure' });
+            // Report the error
+            this.props.onError({ error: 'secured fields have failed to configure', fieldType: 'csfConfigFailure' });
```
ribeiroguilherme, line 49: Maybe it makes sense to throw a proper 'AdyenCheckoutError' object? 

```js
this.props.onErrror(new AdyenCheckoutError('ERROR', error));
```

PR #1721, opened 2022-09-06
Added Online Banking PL (P24) component

jobsturm:
## Summary
This adds support for Online Banking PL (P24)


packages/playground/src/pages/IssuerLists/IssuerLists.js
```diff
@@ -9,7 +9,7 @@ import '../../style.scss';
     const session = await createSession({
         amount: {
             value: 123,
-            currency: 'EUR'
```
ribeiroguilherme, line 12: We can leave 'EUR' as default - there is no point to change it

ribeiroguilherme (approved):

packages/lib/src/components/P24/index.ts
```diff
@@ -0,0 +1,7 @@
+import IssuerListContainer from '../helpers/IssuerListContainer';
+
+class P24Element extends IssuerListContainer {
```
ribeiroguilherme, line 3: Better to name the payment method to `OnlineBankingPL`, same as we do with the other online banking payment methods (`OnlineBankingCZ`, `OnlineBankingSK`..).  Also rename the folder that keeps it.

packages/lib/src/components/P24/index.ts
```diff
@@ -0,0 +1,7 @@
...
+    public static type = 'dotpay';
```
ribeiroguilherme, line 4: Are we sure that tx variant is `dotpay` ? It seems that the ticket description mentions `onlineBanking_PL`

packages/playground/src/pages/IssuerLists/IssuerLists.js
```diff
@@ -48,6 +48,9 @@ import '../../style.scss';
     // Dotpay
     window.dotpay = checkout.create('dotpay').mount('.dotpay-field');
 
+    // P24
+    window.dotpay = checkout.create('p24').mount('.p24-field');
```
ribeiroguilherme, line 52: This variable name conflicts with the one on line 49.

ribeiroguilherme (changes requested):

packages/lib/src/components/index.ts
```diff
@@ -131,6 +132,7 @@ const componentsMap = {
     molpay_ebanking_VN: MolPayEBankingVN,
     onlinebanking_IN: OnlineBankingINElement,
     openbanking_UK: OpenBankingUK,
+    p24: P24,
```
ribeiroguilherme, line 135: This mapping should be updated as well to use `onlineBanking_PL` key

ribeiroguilherme (dismissed):

packages/lib/src/components/UIElement.tsx
```diff
@@ -228,6 +228,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
      * Return the type of an element
      */
     get type(): string {
+        console.log(this.props.type || this.constructor['type']);
```
sponglord, line 231: Can we get rid of this ;-)

sponglord (approved):

PR #1723, opened 2022-09-07
[WIP] i18n: Adds ukrainian shopper locale

thomaslarsson:
## Summary
- Adds ukranian (ua_UK) as a shopper locale.

**Please note:** I do not speak Ukranian myself. All translations are provided by profesional translators, but context is hard when simply viewing json files. The translation is also missing a few, newer translation keys added since we copied the file from master. Hope to also have those added before merge. Would love input from native speakers before merge.

## Demo changes
- Adds support for changing shopperLocale via query params for the demo pages


ribeiroguilherme:
Hi @thomaslarsson,

We appreciate your collaboration but unfortunately we have our third-party service to generate the translations for us.

In the future if you'd like us to support another language you can just create an issue asking for it. We'll then check if it would benefit other merchants as well and if it does, we'll ask our third-party service to generate the translation files (so no need to generate the files yourself).

Also keep in mind that even if we don't support a language you can create your own [custom translations](https://docs.adyen.com/online-payments/web-drop-in/customization#creating-a-localization).

Thanks for the effort and feel free to contribute again in the future!

PR #1724, opened 2022-09-08
fix checkbox nanocss bug

m1aw:
## Summary

This fixes the bug where the checkbox tick was only showing one of the lines. I was unable to find when this bug started, and what was causing it, besides seems to be a bug in `nanocss`.

Might be useful in the future further investigation on this, because `nanocss` version has not been change for more than a year.

## Tested scenarios

- Check the CSS generated in the inspector
- Checked visually the checkbox 

**Fixed issue**:


sponglord (approved):

ribeiroguilherme (approved):
Did you test cross browser's by any chance?
It would be nice to link the nanocss issue link to this ticket - so we can keep track of it

PR #1729, opened 2022-09-12
Bumping securedField version

sponglord:
## Summary
Bumping securedField version and adding check for if securedFields iframe loads new content

## Tested scenarios
All tests pass

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/lib/CSF/extensions/createSecuredFields.ts
+++ b/packages/lib/src/components/internal/SecuredFields/lib/CSF/extensions/createSecuredFields.ts
@@ -12,6 +12,7 @@ import cardType from '../utils/cardType';
 import { SFSetupObject } from '../../securedField/AbstractSecuredField';
 import SecuredField from '../../securedField/SecuredField';
 import { CardObject, CbObjOnBrand, SFFeedbackObj, CbObjOnLoad, CVCPolicyType } from '../../types';
+import AdyenCheckoutError from '../../../../../../core/Errors/AdyenCheckoutError';
 
 /**
  * cvcPolicy - 'required' | 'optional' | 'hidden'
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/internal/SecuredFields/lib/CSF/extensions/createSecuredFields.ts
```diff
@@ -198,6 +198,15 @@ export function setupSecuredField(pItem: HTMLElement): Promise<any> {
...
                 if (window._b$dl) console.log('### createSecuredFields::onIframeLoaded:: this.state.iframeCount=', this.state.iframeCount);
 
+                // One of our existing securedFields has just loaded new content!
+                if (this.state.iframeCount > this.state.numIframes) {
+                    this.destroySecuredFields();
+                    throw new Error(
+                        `ERROR: One or more securedFields has just loaded new content. This should never happen. securedFields have been removed.
+                        iframe load count=${this.state.iframeCount}. Expected count:${this.state.numIframes}`
+                    );
+                }
```
ribeiroguilherme, line 208: Maybe you could use `AdyenCheckoutError` model here instead of the generic one?

ribeiroguilherme (approved):

packages/lib/src/components/internal/SecuredFields/lib/CSF/extensions/createSecuredFields.ts
sponglord: But hat's more for formatting an error to be passed to the merchant defined `onError` handler - what we want here is a proper big, breaking, console error

sponglord: But I guess you can throw them too... so, ok

PR #1730, opened 2022-09-12
Improving contrast on main colors

sponglord:
## Summary
Improving contrast on main colors re. a11y guidelines - namely the red, green & blue colours that we use for borders etc for focus, error & valid states

NOTE: follow up work needs to be done to change the colours in the error & valid icons - but that is a separate issue in a separate repo

**Fixed issue**:  COWEB-1140

### DIFF
--- a/packages/lib/src/components/Ach/components/AchInput/defaultStyles.ts
+++ b/packages/lib/src/components/Ach/components/AchInput/defaultStyles.ts
@@ -1,5 +1,5 @@
 export default {
     base: {
-        caretColor: '#0066FF'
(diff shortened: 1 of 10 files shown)


ribeiroguilherme (approved):
Is it possible to import the CSS variables in the `.scss` files in this PR? 
We should use/reuse the proper variables instead of the hexadecimal code.

sponglord:
Maybe - but that will have to be in another PR, 'cos I'd like to get these changes in now, and testing that all the values are still set everywhere if we switch to the css variables will take too long

PR #1732, opened 2022-09-13
UPI redesign

ribeiroguilherme:
## Summary
Redesign of the UPI component, which instead of two Pay buttons, we are using only one Pay button. The goal is to make this payment method compatible with the idea of the merchant provide his own custom pay button.

Highlights:
- UPI supports an optional configuration property called `defaultMode` , which accepts `vpa` or `qrCode` as value . This property defines which mode will be set by default when the payment method is opened. Its default value is `vpa` .
Ex:
```js
const upi = checkout.create('upi', { defaultMode: 'qrCode' }).mount('.upi-field');
```

- Introduced `SegmentedControl` component as part of our reusable components, which derives from [this UI element](https://developer.apple.com/design/human-interface-guidelines/components/selection-and-input/segmented-controls/#:~:text=A%20segmented%20control%20is%20a,the%20control%20as%20a%20whole) that is commonly used on mobile devices. 

## Tested scenarios
- Manual testing trying out both UPI flows (upi_qr / upi_collect )

**Fixed issue**: COWEB-1144


packages/lib/src/components/internal/SegmentedControl/SegmentedControl.scss
```diff
@@ -0,0 +1,55 @@
...
+
+  &--selected {
+    background: #E5F1FF;
+    border: 1.5px solid #0075FF;
+    color: #0075FF;
+    font-weight: 700;
+  }
+}
+
+.adyen-checkout__segmented-control--disabled > .adyen-checkout__segmented-control-segment {
```
m1aw, line 48: Nit pick, but could we nest these rules?

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -1,58 +1,104 @@
...
+    }
+};
+
+export default function UPIComponent({ defaultMode, onChange, onUpdateMode, payButton, showPayButton }: UPIComponentProps): h.JSX.Element {
     const { i18n, loadingContext } = useCoreContext();
     const inputRef = useRef<VpaInputHandlers>(null);
     const [status, setStatus] = useState<UIElementStatus>('ready');
-    const [isQrCodeFlow, setIsQrCodeFlow] = useState<boolean>(false);
+    const [mode, setMode] = useState<UpiMode>(defaultMode);
+    const isMobile = window.matchMedia('(max-width: 768px)').matches && /Android|iPhone|iPod/.test(navigator.userAgent);
```
m1aw, line 37: We should extract this function into utils. I'm sure it's going to be needed in some other place later.

m1aw (approved):

packages/lib/src/components/internal/SegmentedControl/SegmentedControl.scss
ribeiroguilherme: done

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
ribeiroguilherme: done!

PR #1738, opened 2022-09-14
Updating git workflows to use node 14

ribeiroguilherme:
## Summary
- Updating workflows to use node 14

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x]
+        node-version: [14.x, 16.x]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x]
```
sponglord, line 15: Is their any reason/advantage to know that it still builds in Node 12?

m1aw:
Should we also make run on the current LTS (16.x.x)?

.github/workflows/ci-test.yml
ribeiroguilherme: Not really as far as I know - this line was added 2 years ago where Node 12 had [active support and security support](https://endoflife.date/nodejs). I believe that is just legacy config.

ribeiroguilherme:
> Should we also make run on the current LTS (16.x.x)?

Makes sense, as we are internally planning to upgrade to Node 16 any time soon. I will add it to the CI pipeline at least.

ribeiroguilherme:
> > Should we also make run on the current LTS (16.x.x)?
> 
> Makes sense, as we are internally planning to upgrade to Node 16 any time soon. I will add it to the CI pipeline at least.

Ok it seems that adding Node16 breaks and the tests don't pass - we need to fix few of them that are throwing `UnhandledPromiseRejection` . Better to do this in another PR, and keep the scope of this one to upgrade to Node14.

m1aw (approved):

PR #1747, opened 2022-09-19
FIX Added array as fallback when no issuers are found in OnlineBankingPL

jobsturm:


sponglord:
Not necessary after #1748

PR #1750, opened 2022-09-21
Storing the checkoutAttemptId on sessionStorage

ribeiroguilherme:
## Summary
This PR handles the storage and usage of the `checkoutAttemptId` on the `sessionStorage`. This change aims to improve the calculation of the conversion rates, in scenarios like the shopper reloads the checkout page or comes back from a redirect payment method.

Highlights on how it works:

- If `checkoutAttemptId` is passed during the initialization of `AdyenCheckout`, then we don't handle the storage of the ID - the PBL app or the merchant itself is handling it in this scenario
- In case that there is no `checkoutAttemptId` passed, the SDK will request it and save on the sessionStorage along with a timestamp of 15 minutes - so if a redirect happens, or if the shopper decides to move out from the checkout and add another extra item to the cart, we would still use the same checkoutAttemptId if the timespan is within the 15 minutes.

## Tested scenarios
- Passing `checkoutAttemptId` to `AdyenCheckout` doesn't trigger a request to get new ID, neither use the one on the session storage
- After saving the value in the session storage and refreshing the page, the UI will sill use the same ID
- After 15 min passed, the UI requests another ID

**Fixed issue**:  COWEB-1099

### DIFF
--- a/packages/e2e/app/src/pages/IssuerLists/IssuerLists.js
+++ b/packages/e2e/app/src/pages/IssuerLists/IssuerLists.js
@@ -12,6 +12,9 @@ const initCheckout = async () => {
     });
 
     const checkout = await AdyenCheckout({
+        analytics: {
+            enabled: false
+        },
         amount,
         paymentMethodsResponse,
         clientKey: process.env.__CLIENT_KEY__,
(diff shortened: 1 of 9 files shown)


sponglord:
As discussed - is it possible to write any tests for this?

sponglord (approved):

PR #1751, opened 2022-09-21
Adds PayByBank support

m1aw:
## Summary

The objective of this ticket is to add support for Open Banking / PayByBank payment method.
Architecturally this was solved by adding two new functionalities to IssuerListContainer. One is the ability to handle a scenario where no issuers are returned by `paymentMethods/`. In this first scenario display a single button that works has a redirect payment method.
The second new functionality is the ability to manipulate the issuer list, into creating a brand list to display in the PaymentMethodItem on Dropin.

## Tested scenarios

**Fixed issue**:


packages/lib/src/components/index.ts
```diff
@@ -135,7 +135,7 @@ const componentsMap = {
     molpay_ebanking_VN: MolPayEBankingVN,
     onlinebanking_IN: OnlineBankingINElement,
     onlineBanking_PL: OnlinebankingPL,
-    openbanking_UK: OpenBankingUK,
+    paybybank: PayByBank,
```
sponglord, line 138: Has the txVariant name changed?

Issue #1712, opened 2022-08-26
Swish payment method always triggers an error on mobile


rvdsteege:
**Describe the bug**
When an error occurs in drop-in, our [Pronamic Pay](https://github.com/pronamic/wp-pronamic-pay) plugin for WordPress redirects to a _failed payment_ page. The payment status then gets updated accordingly and the error is logged.
- https://github.com/pronamic/wp-pronamic-pay-adyen/blob/4.1.0/js/src/checkout-drop-in.js#L19-L38

For the Swish payment method, this seems to result in an unreasonable amount of failed payments (`60%` according to client) with the following error message:

```
swish App was not found
```

This error seems to be triggered by `redirectToApp()` of the QR loader:

https://github.com/Adyen/adyen-web/blob/68e1573e39ff925f3cbcb13799816269ea6c57a8/packages/lib/src/components/internal/QRLoader/QRLoader.tsx#L70-L77

We're thinking the really short timeout of `25` ms might be the culprit. In a test similar to `redirectToApp()`, it seems that even the simplest redirect will trigger the error, even though the redirect works perfectly fine:

```html
<script>
setTimeout( function() {
	console.log( "Unable to redirect" );
}, 25 );

window.location.assign( 'https://www.adyen.com/' );
</script>
```

In the case of Swish, the URL being redirected to uses the custom protocol `swish://` to open their app. This means that the browser might also asks for confirmation before opening the Swish app, which will take a little longer to confirm. It is impossible to confirm that this fast.

**To Reproduce**
1. on mobile, select Swish in drop-in
2. click the button to continue payment
3. the `onError` method of the checkout configuration is called with a `AdyenCheckoutError`

**Expected behavior**
The drop-in should not trigger an error if the Swish app can be opened. Also, if we would want to leave the QR code shown on mobile when the app is not installed (for example to use an other device to scan the QR code), there is no way to distinguish between a real error (like when the payment expires after 3 minutes in https://github.com/Adyen/adyen-web/blob/v5.23.0/packages/lib/src/components/internal/QRLoader/QRLoader.tsx#L94-L98) and the case that the app could not be opened. If the PayPal popup is cancelled for example, we do not redirect to our failed payment page:
- https://github.com/pronamic/wp-pronamic-pay-adyen/blob/4.1.0/js/src/checkout-drop-in.js#L28-L30

**Smartphone (please complete the following information):**
I have tested on at least the following device, but assume it affects others too.
 - Device: iPhone
[...]

PR #1755, opened 2022-09-22
Increasing deep link timeout for QrCode and Await components

ribeiroguilherme:
## Summary

Apparently some devices don't manage to open the respective payment method apps in time, therefore the library throws an Error. There is also differences between Android and iPhone, which seems that Android requires a higher timeout than iPhone.

To be on the safe side, this PR increases the deep link timeout from 25ms to 1 second for the QrCode and Await components.  If the payment method app doesn't open during this interval, then the fallback polling mechanism starts.


remcotolsma:
I don't think this should be merged in, please consider the following comments in review:

- https://github.com/Adyen/adyen-web/issues/1712#issuecomment-1254928641
- https://github.com/Adyen/adyen-web/issues/1712#issuecomment-1255140866

ribeiroguilherme:
Parking this PR for now, as we need to investigate internally a better way to handle the scenario where the payment app is/isn't available in the shoppers device.

PR #1764, opened 2022-09-26
Removing 'onError' call when redirectToApp happens in QRLoader/Await components

DavidVujic:
## Summary

Motivation:
currently, opening an app - in our case the Swish app - will always trigger an AdyenCheckoutError.

This behavior is described in issue #1712.

This PR aims to solve that by letting the (mobile) browser optimistically open the app (i.e. Swish). I haven't found a way to catch any failures. However, on iOS, it will display a popup (see screenshots) notifying the user about the missing app, and will not trigger any script errors in the browser.

## Tested scenarios

With a iPhone connected to a Desktop Mac OS X:
navigated a web page in the iPhone browser, run scripts from the desktop Safari (web developer tools, connect to device).

```javascript
// the url is grabbed from the Swish developer guide
let url = "swish://paymentrequest?token=c28a4061470f4af48973bd2a4642b4fa&callbackurl=merchant%253A%252F%252F";

addEventListener("myFakeEvent", () => window.location.assign(url));

dispatchEvent(new Event("myFakeEvent"));
```

**Fixed issue**: #1712   

__Without the Swish app installed (message in Swedish)__
without_the_swish_app_installed

__With the app installed__
with_swish_app_installed

### DIFF
--- a/packages/lib/src/components/internal/Await/Await.tsx
+++ b/packages/lib/src/components/internal/Await/Await.tsx
@@ -87,11 +87,7 @@ function Await(props: AwaitComponentProps) {
     };
 
     const redirectToApp = (url, fallback = (): void => {}): void => {
-        setTimeout((): void => {
-            // Redirect to the APP failed
-            props.onError(new AdyenCheckoutError('ERROR', `${props.type} App was not found`));
-            fallback();
-        }, 25);
+        setTimeout(fallback, 25);
         window.location.assign(url);
     };
 
--- a/packages/lib/src/components/internal/Await/Await.tsx
+++ b/packages/lib/src/components/internal/Await/Await.tsx
@@ -87,7 +87,7 @@ function Await(props: AwaitComponentProps) {
     };
 
     const redirectToApp = (url, fallback = (): void => {}): void => {
-        setTimeout(fallback, 25);
+        setTimeout(fallback, 1000);
         window.location.assign(url);
     };
 
--- a/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
+++ b/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
@@ -68,11 +68,7 @@ class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
     }
 
     public redirectToApp = (url, fallback = () => {}) => {
-        setTimeout(() => {
-            // Redirect to the APP failed
-            this.props.onError(new AdyenCheckoutError('ERROR', `${this.props.type} App was not found`));
-            fallback();
-        }, 25);
+        setTimeout(fallback, 25);
         window.location.assign(url);
     };
 
--- a/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
+++ b/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
@@ -68,7 +68,7 @@ class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
     }
 
     public redirectToApp = (url, fallback = () => {}) => {
-        setTimeout(fallback, 25);
+        setTimeout(fallback, 1000);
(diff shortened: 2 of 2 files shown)


DavidVujic:
There's another place in this repo running a very much similar function: the file `components/internal/Await/Await.tsx`

I haven't enough knowledge about the structure in this repo, but think the Swish related issues are in `QRloader`, i.e. the code changed in this PR.

ribeiroguilherme (commented):
Thanks for the PR @DavidVujic !

We discussed internally, and it is better to keep the timeout for now and increase its value. The culprit here is the `onError` being triggered, which seems pointless for this scenario.

Could you adjust your PR to do something similar as what we did [here](https://github.com/Adyen/adyen-web/pull/1755) ? but then removing the `this.props.onError` for both `QRLoader` and `Await`

DavidVujic:
> Thanks for the PR @DavidVujic !
> 
> We discussed internally, and it is better to keep the timeout for now and increase its value. The culprit here is the `onError` being triggered, which seems pointless for this scenario.
> 
> Could you adjust your PR to do something similar as what we did [here](https://github.com/Adyen/adyen-web/pull/1755) ? but then removing the `this.props.onError` for both `QRLoader` and `Await`

Ok, I think I understand! Do you mean that you want the `fallback()` function to be triggered in a `setTimeout`, i.e. only remove the on-error call?

DavidVujic:
@ribeiroguilherme I have updated this Pull Request with:
* the `fallback()` function now is added to the eventloop through `setTimeout` as before, but without throwing an error. 
* increased the timeout to 1000 ms according to your suggestions in #1755 
* changed both QRLoader and Await to use a new shared `redirect` module.

ribeiroguilherme:
> Ok, I think I understand! Do you mean that you want the fallback() function to be triggered in a setTimeout, i.e. only remove the on-error call?

Yes - just increase the timeout and remove the onError calls on both Await and QRLoader 👍 

There is no need to extract the `redirect` function to a separated file - 'redirect' is a generic word and could be applied for different scenarios. Let's keep it for now in the QRLoader and Await files :)

DavidVujic:
> > Ok, I think I understand! Do you mean that you want the fallback() function to be triggered in a setTimeout, i.e. only remove the on-error call?
> 
> Yes - just increase the timeout and remove the onError calls on both Await and QRLoader 👍
> 
> There is no need to extract the `redirect` function to a separated file - 'redirect' is a generic word and could be applied for different scenarios. Let's keep it for now in the QRLoader and Await files :)

Oh, I just moved it to a separate module. No worries, I can revert it.

DavidVujic:
Done! @ribeiroguilherme

DavidVujic:
We are about to go live all-in within a day or two (currently A/B testing) and this is a big issue for our users trying to purchase from their phones with Swish. Would it be possible for you to fast-track this, review and make a decision on merging or not?

ribeiroguilherme (approved):

ribeiroguilherme:
Thanks for your contribution @DavidVujic 👌 

> We are about to go live all-in within a day or two (currently A/B testing) and this is a big issue for our users trying to purchase from their phones with Swish. Would it be possible for you to fast-track this, review and make a decision on merging or not?

It will be merged into master, but our release cycle takes place early next week (Monday afternoon or Tuesday morning). If you need to go live this week, an option is to do the workaround proposed [here](https://github.com/Adyen/adyen-web/issues/1712#issuecomment-1255218535), and then upgrade the library next week.

DavidVujic:
> It will be merged into master, but our release cycle takes place early next week (Monday afternoon or Tuesday morning). If you need to go live this week, an option is to do the workaround proposed [here](https://github.com/Adyen/adyen-web/issues/1712#issuecomment-1255218535), and then upgrade the library next week.

Great! 🎉

Issue #1769, opened 2022-09-28
Missing translation for QR Component


markusahlstrand:
**Describe the bug**
The QR Component uses the i18n key `or` which is missing at least from the Swedish locale file.

**Expected behavior**
When using a swedish browser the to `or` should be translated to `eller`

**Screenshots**
<img width="441" alt="Skärmavbild 2022-09-27 kl  17 14 34" src="https://user-images.githubusercontent.com/89469/192770485-f9d57e00-486c-4ce8-b9cb-685351bd2d02.png">

**Additional context**
This is where the i18n tag is used:
https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/components/internal/QRLoader/QRLoader.tsx#L228

This is the locale file for swedish:
https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/language/locales/sv-SE.json

ribeiroguilherme:
Thanks for reaching out @markusahlstrand . I created an PR fixing this issue, and we expect it to be released early next week. I will close this github issue once the fix is live. Cheers.

ribeiroguilherme:
The fix is live, therefore closing this issue. Thanks for reaching out!

PR #1771, opened 2022-09-30
Fix for missing translation for QRLoader/Await

ribeiroguilherme:
## Summary
`QRLoader` and `Await` components were using invalid translation key. 

This PR reuses the `<ContentSeparator />` component which has the proper translation in place for 'qrCodeOrApp' scenario

## Tested scenarios
- Manual testing checking that proper translation shows up

**Fixed issue**: #1769

### DIFF
--- a/packages/lib/src/components/internal/Await/Await.scss
+++ b/packages/lib/src/components/internal/Await/Await.scss
@@ -81,36 +81,6 @@
     display: none;
 }
 
-.adyen-checkout__await__separator__label {
-    position: relative;
-    font-size: 13px;
-    color: $color-gray-darker;
-    overflow: hidden;
-    text-align: center;
-    z-index: 1;
-    display: block;
-}
-
-.adyen-checkout__await__separator__label:before,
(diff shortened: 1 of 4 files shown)


sponglord:
A lot of styling has been removed from both the `Await.scss` and the `QRLoader.scss` to be replaced by the much simpler set of rules in the `ContentSeparator`. All of which is good if it simplifies things - but I just wanted to check that everything still _looks_ as it should?

sponglord:
And... where's the translation?

sponglord (approved):

ribeiroguilherme:
> A lot of styling has been removed from both the `Await.scss` and the `QRLoader.scss` to be replaced by the much simpler set of rules in the `ContentSeparator`. All of which is good if it simplifies things - but I just wanted to check that everything still _looks_ as it should?

Looks better actually :)

> And... where's the translation? Or was the "qrCodeOrApp" always present in the translation files?

'qrCodeOrApp' is present in the translation files. It is the default value in the `ContentSeparator` component

PR #1772, opened 2022-09-30
Feature/expanding core mount fny

sponglord:
## Summary
Currently in order to mount/unmount and then mount a component again the user has to call:
`mount` -> `unmount` and then `remount` for any second and subsequent "mounting"

Following our discussions around this, this PR expands the core `mount` functionality to also incorporate a "remount" operation. 
So now the user would call: `mount` -> `unmount` -> `mount`

The public `remount` function has been left in for legacy reasons but is no longer expected to be used

## Tested scenarios
Extensive manual testing has taken place both of mounting/unmounting/mounting in a "regular" environment where the node on which to be mounted is passed as a selector e.g. `'.card-field'`;
and also in a (re-rendering) React environment where the node is passed as a React `ref` 
The tests were conducted for both the Credit card comp and the PayPal component.

Following these latter tests a new element has been added to the Cards Playground: "Card in React" which adds a React environment in which we implement our Card component and which allows testing of the mounting & unmounting of the Card component

**Relates to issue**:  #1638

### DIFF
--- a/packages/lib/src/components/BaseElement.ts
+++ b/packages/lib/src/components/BaseElement.ts
@@ -78,21 +78,22 @@ class BaseElement<P extends BaseElementProps> {
         }
 
         if (this._node) {
-            throw new Error('Component is already mounted.');
+            this.unmount(); // new, if this._node exists then we are "remounting" so we first need to unmount if it's not already been done
+        } else {
+            // Set up analytics, once
+            if (this.props.modules && this.props.modules.analytics && !this.props.isDropin) {
+                this.props.modules.analytics.send({
+                    containerWidth: this._node && this._node.offsetWidth,
+                    component: this.constructor['analyticsType'] ?? this.constructor['type'],
+                    flavor: 'components'
+                });
+            }
         }
 
         this._node = node;
         this._component = this.render();
 
         render(this._component, node);
-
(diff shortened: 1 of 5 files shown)


packages/lib/src/components/BaseElement.ts
```diff
@@ -106,11 +107,13 @@ class BaseElement<P extends BaseElementProps> {
...
 
-        return this.unmount().remount();
+        return this.unmount().mount(this._node); // for new mount fny
     }
 
     /**
-     * Unmounts an element and mounts it again on the same node
+     * Unmounts an element and mounts it again on the same node i.e. allows mount w/o having to pass a node.
+     * Should be "private" & undocumented (although being a public function is useful for testing).
+     * Left in for legacy reasons
```
ribeiroguilherme, line 116: @sponglord can you recall me for which scenario is it useful for testing? Does it make sense to add to our SDK v6 plans to completely remove this method?

ribeiroguilherme (approved):

packages/lib/src/components/BaseElement.ts
sponglord: remount is useful for testing (according to Marc).
And yes - it should be removed for v6

PR #1806, opened 2022-10-20
Change level to silent

aefimov:
## Summary

In our case for Blik integration sometimes (rare but it exists) we got message in JS console log (we used some mechanism to wrap up that logging and put into our Kibana, it logged with warning level and visible to alerts):
```
Call to https://checkoutshopper-live.adyen.com/checkoutshopper/services/PaymentInitiation/v1/status?
clientKey=live_<reducted> failed. Error= TypeError: Load failed
```

## Tested scenarios

 - Initialise component on slow or bad network, see on JS console

**Fixed issue**:

### DIFF
--- a/packages/lib/src/core/Services/payment-status.ts
+++ b/packages/lib/src/core/Services/payment-status.ts
@@ -12,10 +12,9 @@ export default function checkPaymentStatus(paymentData, clientKey, loadingContex
         throw new Error('Could not check the payment status');
     }
 
-    const options = {
+    return httpPost({
         loadingContext,
-        path: `services/PaymentInitiation/v1/status?clientKey=${clientKey}`
-    };
-
-    return httpPost(options, { paymentData });
+        path: `services/PaymentInitiation/v1/status?clientKey=${clientKey}`,
+        errorLevel: 'silent'
+    }, { paymentData });
 }


packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -67,7 +67,7 @@ function Await(props: AwaitComponentProps) {
             .catch(({ message, ...response }) => ({
                 type: 'network-error',
                 props: {
-                    ...(message && { message: this.props.i18n.get(message) }),
+                    ...(message && { message: i18n.get(message) }),
```
ribeiroguilherme, line 70: Nice catch!

packages/lib/src/core/Services/payment-status.ts
```diff
@@ -12,10 +12,9 @@ export default function checkPaymentStatus(paymentData, clientKey, loadingContex
...
 
-    const options = {
+    return httpPost({
         loadingContext,
-        path: `services/PaymentInitiation/v1/status?clientKey=${clientKey}`
-    };
-
-    return httpPost(options, { paymentData });
+        path: `services/PaymentInitiation/v1/status?clientKey=${clientKey}`,
+        errorLevel: 'fatal'
```
ribeiroguilherme, line 18: Adding the `errorLevel: 'fatal'` here would avoid polluting the console with the console.warn - but I am not fully sure that this is the best approach. 

It would end up throwing an `AdyenCheckoutError('NETWORK_ERROR', message')`, but this object won't be passed to the `onError` - it would be ignored [here](https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/components/internal/Await/Await.tsx#L84) . 

Maybe a better approach is to use `errorLevel: 'silent'` , and let the Component retry the request (this way, there is no logging). If the timer of the Await runs out, then the `onError` will be triggered and your code can react upon that. What do you think of this approach?

Btw, do you know what is the reason for the request to the 'status' endpoint to fail?  Maybe we can try to find the root cause and fix that.

aefimov: >Btw, do you know what is the reason for the request to the 'status' endpoint to fail? Maybe we can try to find the root cause and fix that.

aefimov: Maybe we need to change status type from `network-error` to `error`, then `onError` will happen.
Also, checkPaymentStatus can raise Error for that code:
```
    if (!paymentData || !clientKey) {
        throw new Error('Could not check the payment status');
    }
```
So it in any case not a `network-error`. Wdyt?

ribeiroguilherme: I'd say that start triggering the `onError` when there is a network problem can be a breaking change in this case. 

Merchants implement different use-cases when this callback is triggered (Some merchants redirect shoppers, other merchants display pop-ups, etc). We can't simply start triggering it if there is a network error. And, if such an error happens in this specific scenario, the frontend can ignore it and retry the request few seconds later due to the polling mechanism.

It seems that the feasible approach here is to use `errorLevel: 'silent'` and let the onError be triggered when the Component times out.

ribeiroguilherme: Can you rebase on master? The issue with the i18n is already [fixed and released](https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/components/internal/Await/Await.tsx#L83)

aefimov: Then it fixed? WE start getting another one error messages :)

aefimov: Now we get:

aefimov: Now we get:
```
Cannot read properties of undefined (reading 'get')
```

packages/lib/src/components/internal/Await/Await.tsx
aefimov: Rebased

packages/lib/src/core/Services/payment-status.ts
aefimov: Rebased anyway

aefimov: We have no more messages from our side yet. We also migrated to latest `adyen-web`. This `silent` thingy is worth now?

ribeiroguilherme: Hey @aefimov . Thanks for getting back with the feedback. I'm glad that there are no error messages anymore.

I'd say that we do not need the `silent` change anymore - there is no one complaining about it, so let's keep as it is :)

aefimov: Ok, closing PR, then! Thanks for help!

PR #1814, opened 2022-10-27
Fixes AmazonPay settings loading region from wrong defaultProp

m1aw:
## Summary

There was a bug found where amazon would always load locale `en_GB` even when in the US. This would cause the script to fail when instantiating if no `region: US` prop would be passed.
This property has been deprecated and only `configuration.region` should be used.

Further more a generalised bug was detected where almost none of the components default pros are actually typed.

In this PR was also cleanup an unused property `design`. Since it cannot be found anywhere in amazon documentation.

## Tested scenarios

Tested AmazonPay loading in the US and AmazonPay loading in EU.

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/AmazonPay/defaultProps.ts
+++ b/packages/lib/src/components/AmazonPay/defaultProps.ts
@@ -1,11 +1,12 @@
-export default {
+import { AmazonPayElementProps } from './types';
+
+const defautProps: Partial<AmazonPayElementProps> = {
     cancelUrl: typeof window !== 'undefined' ? window.location.href : '',
     configuration: {},
     environment: 'TEST',
     locale: 'en_GB',
     placement: 'Cart',
     productType: 'PayAndShip',
-    region: 'EU',
     returnUrl: typeof window !== 'undefined' ? window.location.href : '',
     showOrderButton: true,
     showChangePaymentDetailsButton: false,
@@ -14,3 +15,5 @@ export default {
     onClick: resolve => resolve(),
     onSignOut: resolve => resolve()
 };
+
+export default defautProps;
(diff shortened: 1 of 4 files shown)


packages/lib/src/components/AmazonPay/defaultProps.ts
```diff
@@ -1,16 +1,19 @@
-export default {
+import {AmazonPayElementProps} from "./types";
```
ribeiroguilherme, line 1: Seems like prettier is disabled? It didn't add the spaces between the brackets

packages/lib/src/components/AmazonPay/defaultProps.ts
```diff
@@ -1,16 +1,19 @@
...
+
+const defautProps: Partial<AmazonPayElementProps> = {
     cancelUrl: typeof window !== 'undefined' ? window.location.href : '',
     configuration: {},
     environment: 'TEST',
     locale: 'en_GB',
     placement: 'Cart',
     productType: 'PayAndShip',
-    region: 'EU',
     returnUrl: typeof window !== 'undefined' ? window.location.href : '',
```
ribeiroguilherme, line 10: Does it make sense to give deprecation message that this property shouldn't be used (in case it is added in the AmazonPay config)? 

I can imagine merchants using it, then upgrading the library and out of the sudden it stops working. Is this a valid use case?

ribeiroguilherme (commented):
Tests need to be fixed

packages/lib/src/components/AmazonPay/defaultProps.ts
m1aw: There is no real reason for them to be using this prop. First there where no merchants using AmazonPay in the US since it lunched recently. Plus it has been broken since launch since it's not the prop that is sent to AmazonPay SDK.

ribeiroguilherme (approved):

PR #1821, opened 2022-11-01
Feature: PromptPay Thailand for Web

jobsturm:
## Summary
Added the QR code needed for PromptPay.

## Tested scenarios
Couldn't test past pending yet.


packages/lib/src/components/PromptPay/PromptPay.test.ts
```diff
@@ -0,0 +1,26 @@
+import PromptPay from './PromptPay';
+
+describe('PromptPay', () => {
+    describe('formatProps', () => {});
```
ribeiroguilherme, line 4: Useless describe block. It can be deleted

packages/lib/src/components/PromptPay/PromptPay.test.ts
```diff
@@ -0,0 +1,24 @@
+import PromptPay from './PromptPay';
+
+describe('formatProps', () => {});
```
ribeiroguilherme, line 3: I meant about this describe block 'formatProps' - there is no point to have it in this test, it is not doing anything

ribeiroguilherme (approved):

PR #1822, opened 2022-11-01
Added QR code component for Paynow

jobsturm:
## Summary
Added the QR code component for Paynow, couldn't test the full flow yet.

## Tested scenarios
Tested that the QR code shows and fails when the countdown runs out.

### DIFF
--- a/packages/lib/src/components/PayNow/PayNow.test.ts
+++ b/packages/lib/src/components/PayNow/PayNow.test.ts
@@ -1,0 +1,26 @@
+import PayNow from './PayNow';
+
+describe('PayNow', () => {
+    describe('formatProps', () => {});
+
+    describe('isValid', () => {
+        test('should be always true', () => {
+            const paynow = new PayNow({});
+            expect(paynow.isValid).toBe(true);
+        });
+    });
+
+    describe('get data', () => {
+        test('always returns a type', () => {
+            const paynow = new PayNow({});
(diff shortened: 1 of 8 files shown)


packages/lib/src/components/PayNow/PayNow.test.ts
```diff
@@ -0,0 +1,26 @@
+import PayNow from './PayNow';
+
+describe('PayNow', () => {
+    describe('formatProps', () => {});
```
ribeiroguilherme, line 4: Useless describe block. It can be deleted

packages/lib/src/components/PayNow/PayNow.ts
```diff
@@ -0,0 +1,16 @@
+import QRLoaderContainer from '../helpers/QRLoaderContainer';
+import { delay, countdownTime } from './config';
+
+class WeChatPayElement extends QRLoaderContainer {
```
ribeiroguilherme, line 4: `WeChatPayElement` ?

packages/lib/src/components/PayNow/PayNow.test.ts
```diff
@@ -0,0 +1,25 @@
+import PayNow from './PayNow';
+
+describe('formatProps', () => {});
```
ribeiroguilherme, line 3: I meant about this describe block 'formatProps' - there is no point to have it in this test, it is not doing anything

ribeiroguilherme (approved):

PR #1824, opened 2022-11-04
FIX: Adjust credit card brand animation in card component

jobsturm:
## Summary
See title. The credit card brands now slide out when the first two numbers of the card are recognised as a brand.

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.test.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.test.tsx
@@ -78,7 +78,7 @@ describe('CardInput - Brands beneath Card Number field', () => {
         expect(wrapper.find('.adyen-checkout__card__brands__brand-wrapper--disabled')).toHaveLength(0);
     });
 
-    test('should disable the brands icons that are not detected', () => {
+    test('should hide all brand icons when brand is detected', () => {
         const detectedBrand = 'visa';
         const brandsIcons = [
             { name: 'visa', icon: 'visa.png' },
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/Card/components/CardInput/components/AvailableBrands/AvailableBrands.scss
```diff
@@ -6,23 +6,28 @@
   flex-shrink: 1;
   flex-wrap: wrap;
   gap: 4px;
+  height: 16px;
+  overflow: hidden;
   margin-top: -8px;
   margin-bottom: 16px;
+  will-change: height;
```
ribeiroguilherme, line 13: First time seeing this property. Is this doing something in the code? MDN says that it should be used as last resort if we face performance issues.

packages/lib/src/components/Card/components/CardInput/CardInput.test.tsx
```diff
@@ -78,19 +78,27 @@ describe('CardInput - Brands beneath Card Number field', () => {
...
+    test('should show all brand icons when no brand is detected', () => {
+        const detectedBrand = 'card';
+        const brandsIcons = [
+            { name: 'visa', icon: 'visa.png' },
+            { name: 'mc', icon: 'mc.png' },
+            { name: 'amex', icon: 'amex.png' }
+        ];
+        const wrapper = mount(<CardInput i18n={i18n} brand={detectedBrand} brandsIcons={brandsIcons} />);
+        const brandWrapper = wrapper.find('.checkout__card__brands');
+        expect(wrapper.find('.adyen-checkout__card__brands--hidden')).toHaveLength(0);
```
ribeiroguilherme, line 101: Testing if the code has a classname does not mean that it works as expected - the CSS could be using the wrong properties. 
In this case you should check if the images were rendered in the DOM or not.

jobsturm: I think they will still be in the DOM, since they're only animated away. I could check if their height is 0px, is that valid too?

packages/lib/src/components/Card/components/CardInput/components/AvailableBrands/AvailableBrands.scss
jobsturm: I use this on animated properties that will shift the layout, it tells the browser it will change so it can anticipate this change. But you're right, I didn't check the MDN article in a long time and after rereading it, the `will-change` property is definitely not warranted here.

packages/lib/src/components/Card/components/CardInput/CardInput.test.tsx
ribeiroguilherme: It was agreed to follow up on this in another PR

ribeiroguilherme (approved):

PR #1825, opened 2022-11-04
fixes payment errors on order flow not triggering any callback

m1aw:
## Summary

In this change we fix a bug where no callback was called when a `payment/` error was received. 
This would happen because of the order object in the response would trigger `handleOrder`, which would just update the component state but not trigger anything.
This change moves the responsibility of handling an order to the payment method. This means, that `Giftcard` is the only payment method capable of triggering `onOrderCreated`. A fallback is created in `UIElement` so handle a response with "order" has `onPaymentCompleted`. This last scenario should only happen when a payment fails since other payment methods always pay full remaining amount. 

## Tested scenarios

  | Dropin | Comp | Custom Button
-- | -- | -- | --
Giftcard -> Authorization | ✅ | ✅ | ✅
Giftcard -> Success | ✅ | ✅ | ✅
Giftcard -> Fail | ✅ | ✅ | ✅
Card -> Success | ✅ | ✅ |  
Card -> Fail | ✅ | ✅ |  
Giftcard -> Card -> Fail | ✅ | ✅ | ✅
Giftcard -> Card -> Success | ✅ | ✅ | ✅
Giftcard -> Giftcard -> Card -> Fail | ✅ | ✅ |  
Giftcard -> Giftcard -> Card -> Success | ✅ | ✅ |  

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Dropin/Dropin.tsx
+++ b/packages/lib/src/components/Dropin/Dropin.tsx
@@ -144,8 +144,8 @@ class DropinElement extends UIElement<DropinElementProps> {
      * handleOrder is implemented so we don't trigger a callback like in the components
      * @param response - PaymentResponse
      */
-    protected handleOrder = (response: PaymentResponse): void => {
-        this.updateParent({ order: response.order });
+    protected handleOrder = ({ order }: PaymentResponse): void => {
+        this.updateParent({ order });
     };
 
     closeActivePaymentMethod() {
--- a/packages/lib/src/components/Dropin/Dropin.tsx
+++ b/packages/lib/src/components/Dropin/Dropin.tsx
@@ -9,6 +9,7 @@ import { getCommonProps } from './components/utils';
 import { createElements, createStoredElements } from './elements';
 import createInstantPaymentElements from './elements/createInstantPaymentElements';
 import { hasOwnProperty } from '../../utils/hasOwnProperty';
+import { PaymentResponse } from '../types';
 
 const SUPPORTED_INSTANT_PAYMENTS = ['paywithgoogle', 'applepay'];
 
@@ -138,6 +139,14 @@ class DropinElement extends UIElement<DropinElementProps> {
 
         return null;
     }
+
+    /**
+     * handleOrder is implemented so we don't trigger a callback like in the components
+     * @param response - PaymentResponse
+     */
+    protected handleOrder = (response: PaymentResponse): void => {
+        this.updateParent({ order: response.order });
+    };
 
     closeActivePaymentMethod() {
         this.dropinRef.closeActivePaymentMethod();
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -74,8 +75,9 @@ export class GiftcardElement extends UIElement {
         }
     };
 
-    protected handleOrder = (order: Order) => {
-        this.elementRef._parentInstance.update({ order });
+    protected handleOrder = (response: PaymentResponse) => {
+        const order: Order = response.order;
+        this._parentInstance.update({ order });
```
sponglord, line 80: is `elementRef` not still needed?

m1aw: It's not needed since the instance is elementRef is also the element itself right?
I changed because I added type to `public elementRef: UIElement;` in UIElement, but that a type check error.

sponglord: `this.elementRef._parentInstance` & `this._parentInstance` are the same thing for the standalone Giftcard comp.
But are they the same for the Dropin?

sponglord: Update: I think they are the same

sponglord:
I take it you still have Playground fixes to add?

packages/lib/src/components/Giftcard/Giftcard.tsx
m1aw: Let me double check this. I didn't fully test in dropin and I'm now realising there might be some problems.

packages/lib/src/components/UIElement.tsx
```diff
@@ -172,11 +183,13 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
         return null;
     }
 
-    protected handleOrder = (order: Order): void => {
-        this.elementRef._parentInstance.update({ order });
+    protected handleOrder = (response: PaymentResponse): void => {
+        this.elementRef._parentInstance.update({ order: response.order });
```
ribeiroguilherme, line 187: This should use the `updateParent` function

packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -139,6 +140,14 @@ class DropinElement extends UIElement<DropinElementProps> {
         return null;
     }
 
+    /**
+     * handleOrder is implemented so we don't trigger a callback like in the components
+     * @param response - PaymentResponse
+     */
+    protected handleOrder = (response: PaymentResponse): void => {
```
jobsturm, line 147: I think order could be destructured from response since nothing from response is being used.

packages/lib/src/components/Giftcard/Giftcard.tsx
```diff
@@ -74,8 +75,9 @@ export class GiftcardElement extends UIElement {
         }
     };
 
-    protected handleOrder = (order: Order) => {
-        this.elementRef._parentInstance.update({ order });
+    protected handleOrder = (response: PaymentResponse) => {
+        const order: Order = response.order;
```
jobsturm, line 79: Also should probably be destructured.

sponglord (approved):

PR #1831, opened 2022-11-07
Blik (Await) Don't call /details endpoint when payload is missing

sponglord:
## Summary
Don't make "final state" calls (leading to a call to `onAdditionalDetails`) when `payload` is missing. 
Instead show an error state in the component.

## Tested scenarios
Mocked responses with `resultCode:"authorised"` and an empty &/or missing 'payload' to ensure an error is shown and no `/details` call is made.
Mocked responses with `resultCode:"pending"` to confirm that the regular polling to the `/status` endpoint continues until the timeout (there were some concerns, expressed in the ticket) that it switched off after three attempts

**Relates to issue**:  DEV-54329

### DIFF
--- a/packages/lib/src/components/internal/Await/Await.tsx
+++ b/packages/lib/src/components/internal/Await/Await.tsx
@@ -80,7 +80,7 @@ function Await(props: AwaitComponentProps) {
             .catch(({ message, ...response }) => ({
                 type: 'network-error',
                 props: {
-                    ...(message && { message: props.getI18n(message) }),
+                    ...(message && { message: i18n.get(message) }),
                     ...response
                 }
             }))
--- a/packages/lib/src/components/internal/Await/Await.tsx
+++ b/packages/lib/src/components/internal/Await/Await.tsx
@@ -35,28 +35,42 @@ function Await(props: AwaitComponentProps) {
     };
 
     const onComplete = (status: StatusObject): void => {
-        setCompleted(true);
-        const state = {
-            data: {
-                details: { payload: status.props.payload },
-                paymentData: props.paymentData
-            }
-        };
+        // Only make details call if we have a payload
+        if (status.props.payload) {
+            setCompleted(true);
+            const state = {
+                data: {
+                    details: { payload: status.props.payload },
+                    paymentData: props.paymentData
+                }
+            };
 
-        props.onComplete(state, this);
+            return props.onComplete(state, this);
+        }
+
+        // Show error state & call merchant defined error callback if we do not have a payload
+        setExpired(true);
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -67,7 +80,7 @@ function Await(props: AwaitComponentProps) {
             .catch(({ message, ...response }) => ({
                 type: 'network-error',
                 props: {
-                    ...(message && { message: this.props.i18n.get(message) }),
+                    ...(message && { message: props.getI18n(message) }),
```
m1aw, line 83: Why was this changed?

sponglord: Because we don't actually use it.
And if we were to use it I thought we should just pass in a function that retrieves a translation rather than the translation "machine" itself (i.e. `i18n`) - which means the Await comp then needs to know about the existence of things like a `get` function

sponglord: My mistake - I didn't think it ever received a `i18n` prop - but I see it retrieves it directly from `CoreContext`.
I'll change it

packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -127,9 +140,13 @@ function Await(props: AwaitComponentProps) {
...
             };
 
+            // Reset 'loading' to ensure that it is changes to this value that are the main "engine" that drives this useEffect
+            setLoading(true);
+
+            // Create (another) interval to poll for a result
             setStoredTimeout(setTimeout(statusInterval, delay));
         }
-    }, [loading, timePassed, expired, completed]);
+    }, [loading, expired, completed]);
```
ribeiroguilherme, line 149: Are you sure we don't need the `timePassed` here? We rely on it inside this useEffect , and its value gets updated often

sponglord: But I'm going to set it back - 'cos using `loading` as the drive has an odd effect on the UI

packages/lib/src/components/internal/Await/types.ts
```diff
@@ -27,4 +27,5 @@ export interface AwaitComponentProps {
     messageText: string;
     awaitText: string;
     ref: any;
+    getI18n?: (key: string, options?) => string; // Should equate to the Language.get method
```
ribeiroguilherme, line 30: Is this needed? I don't see it being used

ribeiroguilherme (approved):

packages/lib/src/components/internal/Await/types.ts
sponglord: you're right it got removed in a recent change to this PR

PR #1854, opened 2022-11-21
Removing dependabot configuration

ribeiroguilherme:
## Summary
Removing dependabot configuration

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,0 @@
-version: 2
-updates:
-- package-ecosystem: npm
-  directory: "/"
-  schedule:
-    interval: daily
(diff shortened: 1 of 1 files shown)


sponglord:
Will re be able to retrieve this file when we need it again? (Apparently the plan is to return to dependabot once they have a `stabilityDays` flag)

ribeiroguilherme:
> Will re be able to retrieve this file when we need it again? (Apparently the plan is to return to dependabot once they have a `stabilityDays` flag)

It will be in the git history, so we can always retrieve it.  But the file and its configuration are very small and straight forward, so should be ok to also create a new one when we enable dependabot again.

sponglord (approved):

PR #1858, opened 2022-11-24
[COWEB-1143] Move checkoutAttemptId field to paymentMethod object in state.data

ribeiroguilherme:
## Summary
Moving `checkoutAttemptId` to `paymentMethod` object in order to increase adoption of telemetry logic.

## Tested scenarios
- e2e test
- manual test

**Fixed issue**:  COWEB-1143

### DIFF
--- a/packages/lib/src/components/PayPal/Paypal.tsx
+++ b/packages/lib/src/components/PayPal/Paypal.tsx
@@ -57,7 +57,7 @@ class PaypalElement extends UIElement<PayPalElementProps> {
     }
 
     updateWithAction(action: PaymentAction) {
-        if (action.paymentMethodType !== this.data.paymentMethod.type) throw new Error('Invalid Action');
+        if (action.paymentMethodType !== this.type) throw new Error('Invalid Action');
 
         if (action.paymentData) {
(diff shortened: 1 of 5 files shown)


packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -57,7 +57,7 @@ class PaypalElement extends UIElement<PayPalElementProps> {
     }
 
     updateWithAction(action: PaymentAction) {
-        if (action.paymentMethodType !== this.data.paymentMethod.type) throw new Error('Invalid Action');
+        if (action.paymentMethodType !== this.type) throw new Error('Invalid Action');
```
sponglord, line 60: Was this line always incorrect in the old code?

sponglord (approved):

packages/lib/src/components/PayPal/Paypal.tsx
ribeiroguilherme: `this.data.paymentMethod.type` is the same as `this.type` . 

I had to update this because Typescript started complaining after adding the return type in the `get data()` method

PR #1861, opened 2022-11-27
Added DuitNow QR Code loader

jobsturm:
## Summary
Added DuitNow QR Code loader

## Tested scenarios
- [x] QR Code shows up in playground


packages/lib/src/components/DuitNow/DuitNow.ts
```diff
@@ -0,0 +1,16 @@
+import QRLoaderContainer from '../helpers/QRLoaderContainer';
+import { delay, countdownTime } from './config';
+
+class PromptPayElement extends QRLoaderContainer {
```
marcperez, line 4: This class name might need to be updated

PR #1862, opened 2022-11-27
Fix/added hint icon to expiry date

jobsturm:
## Summary
Added expiry date icon as hint to the expiry date fields

## Tested scenarios
- [x] Tested in the Playground
- [x] Tested when field is invalid
- [x] Tested when field is valid
- [x] Tested when field is empty

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx
@@ -59,9 +59,9 @@ export default function ExpirationDate(props: ExpirationDateProps) {
             />
             <div className="adyen-checkout__field__exp-date_hint_wrapper">
                 <img
-                    src="https://picsum.photos/24/16"
+                    src="https://checkoutshopper-test.adyen.com/checkoutshopper/images/components/expiry_date_hint.svg"
                     className="adyen-checkout__field__exp-date_hint"
-                    alt={'Expiry date hint'}
+                    alt={fieldLabel}
                 />
             </div>
         </Field>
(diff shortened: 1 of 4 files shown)


packages/lib/src/components/Card/components/CardInput/components/StoredCardFields.tsx
```diff
@@ -41,6 +41,13 @@ export default function StoredCardFields({
                     >
                         {expiryMonth} / {expiryYear}
                     </div>
+                    <div className="adyen-checkout__field__exp-date_hint_wrapper">
```
sponglord, line 44: Do we need the date icon in the Stored Card?

jobsturm: It isn't, removed it!

packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx
```diff
@@ -56,6 +57,19 @@ export default function ExpirationDate(props: ExpirationDateProps) {
...
+                    'adyen-checkout__field__exp-date_hint_wrapper--hidden': error || isValid,
+                }
+            )}>
+                <img
+                    src="https://checkoutshopper-test.adyen.com/checkoutshopper/images/components/expiry_date_hint.svg"
+                    className="adyen-checkout__field__exp-date_hint"
+                    alt={fieldLabel}
+                />
+            </div>
         </Field>
```
ribeiroguilherme, line 73: Why aren't we implementing the same way as the CVC hint image in this case?

sponglord: I believe it's to do with the fact that the cvc icon animates, so is an inline svg. Whilst the date icon does not - so we load it from the server

sponglord (approved):

PR #1864, opened 2022-11-28
Removed auto redirect from QR code loader

jobsturm:
## Summary
Removed the auto redirect from QR code loader on mobile. Instead the user will now have to use the "open app" button below the QR code.

## Tested scenarios
- [x] Tested in the playground.

### DIFF
--- a/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
+++ b/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
@@ -54,24 +54,8 @@ class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
     };
 
     componentDidMount() {
-        const { shouldRedirectOnMobile, url } = this.props;
-        const isMobile = window.matchMedia('(max-width: 768px)').matches && /Android|iPhone|iPod/.test(navigator.userAgent);
-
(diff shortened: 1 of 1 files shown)


jobsturm:
@sponglord Since the redirect function now only consists out of `window.location.assign`, do you think it would be advisable to change the button to an anchor tag?

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -54,22 +54,10 @@ class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
...
-        const startPolling = () => {
-            this.interval = setInterval(this.statusInterval, this.state.delay);
-        };
-
-        if (shouldRedirectOnMobile && url && isMobile) {
-            this.redirectToApp(url, startPolling);
-        } else {
-            startPolling();
-        }
+        this.interval = setInterval(this.statusInterval, this.state.delay);
```
ribeiroguilherme, line 57: Why not completely delete the prop `shouldRedirectOnMobile`? Seems like it is useless now after this change

jobsturm:
Closed, because it got stale. Reopening a new PR with this cherry-picked into it.

PR #1866, opened 2022-11-29
Fixes Pix not submiting when showPayButton was false

m1aw:
## Summary

Fixes a bug FOC-67919 where Pix wouldn't work unless showPayButton was set to true. There was a bug where the Input component that manages the state was hidden when the property was set to false.

## Tested scenarios

- Successful payment methods with `showPayButton: true/false` and `personalDetailsRequired: true/false`.

**Fixed issue**:  
FOC-67919

### DIFF
--- a/packages/lib/src/components/Pix/Pix.tsx
+++ b/packages/lib/src/components/Pix/Pix.tsx
@@ -45,23 +45,19 @@ class PixElement extends QRLoaderContainer<PixProps> {
             return this.renderQRCode();
         }
 
-        if (this.props.showPayButton) {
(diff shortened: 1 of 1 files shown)


marcperez (approved):

sponglord:
So what happens when `showPayButton` is set to false- since there is no longer a `if (this.props.showPayButton)` if-clause in the code anymore?

m1aw:
That logic is already part of PixInput, it was on the Pix component too because that's how it's handle in other QR codes, but it doesn't make sense here.

PR #1867, opened 2022-11-29
FIX: Restrict Blik input field to digits only

jobsturm:
## Summary
Changed the Blik input to accept digits only. Also made room for generalised Formatting types. And moved the Formatter types to `useForm` since the originate from here (discussed this with @ribeiroguilherme)

## Tested scenarios
- [x] Changed the Blik input in the playground
- [x] Wrote tests for the digit only formatter.


packages/lib/src/utils/useForm/useForm.ts
```diff
@@ -11,7 +11,7 @@ function useForm<FormSchema>(props: FormProps): Form<FormSchema> {
     /** Formats and validates a field */
     const processField = ({ key, value, mode }, fieldContext) => {
         // Find a formatting function either stored under 'key' or a level deeper under a 'formatter' property
-        const formatterFn = formatters?.[key]?.formatter ? formatters[key].formatter : formatters?.[key];
+        const formatterFn = (formatters as Formatter)?.[key]?.formatter ? (formatters as Formatter)[key].formatter : formatters?.[key];
```
ribeiroguilherme, line 14: Is there any way to avoid the `as` here? This is error prone as we are telling Typescript which type that variable is. It should be the other way around.

I believe you could assert the type of this object using type-guard function

packages/lib/src/utils/Formatters/types.ts
```diff
@@ -0,0 +1 @@
+export type FormatterFn = (value, context?) => string;
```
ribeiroguilherme, line 1: As you are working on this - do you have more insights about the types of the `value` and `context` objects? It would be great to add here, as I believe they are typed 'any' now

packages/lib/src/utils/Formatters/formatter.test.ts
```diff
@@ -0,0 +1,18 @@
+import { digitsOnlyFormatter } from "./formatters";
+
+function getAllCharactersArray() { 
+    const allCharacters:Array<string> = [];
+    for (let i=32; i<127; i++) { allCharacters.push(String.fromCharCode(i)); }
```
ribeiroguilherme, line 5: Does the generated chars array include symbols as well? So we can make sure that it indeed the regex will remove everything except digits

ribeiroguilherme (commented):
As a suggestion: it could be good to also write an e2e tests for this (Ex: checking that Blik doesn't accept non-digit characters, and that the payment doesn't go through), then you have better understanding how the e2e works, plus we also we cover this regression.

packages/lib/src/utils/Formatters/types.ts
jobsturm: I moves this from the Validator file (they were used for the `addressFormatters`, inferring from the code now I think `val` could be a string. Do you know of any more complex use cases for validators where it might need to accept an object? Context should just be an object since it is used to get state from the input field where the formatter is used.

packages/lib/src/utils/Formatters/formatter.test.ts
ribeiroguilherme: Ah ok, I thought it generated only the alphabet letters.

packages/lib/src/utils/Formatters/types.ts
ribeiroguilherme: I don't recall anymore the possible values of `val` , I just suggested as you are working on this. 

You can add the `string` type and check if there are any linter complains - if not, then we go ahead with that and if in the future turns out there are more types, we fix that.

packages/lib/src/components/Blik/components/BlikInput.tsx
```diff
@@ -56,6 +60,7 @@ function BlikInput(props: BlikInputProps) {
                     onInput: handleChangeFor('blikCode', 'input'),
                     onBlur: handleChangeFor('blikCode', 'blur'),
                     placeholder: '123456',
+                    inputMode: 'number',
```
m1aw, line 63: I think input mode should be `numeric` https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

m1aw: And we should actually check if this produces the correct result because some browsers use the `pattern` attribute.

sponglord: @m1aw - can you elaborate on this comment about the `pattern` attribute?
Foe securedFields we only set `inputmode="numeric"`

m1aw (approved):

PR #1869, opened 2022-11-30
Feature Prettier pre commit hook

m1aw:
## Summary

Runs prettier has pre-commit hook.

## Tested scenarios

**Fixed issue**:


packages/lib/.husky/pre-commit
```diff
@@ -0,0 +1,13 @@
...
+# helper if you are running repo inside a container
+if which vagrant; then
+  vagrant ssh -- -t <<HEREDOC
+  cd workspace/adyen-web/packages/lib;
+  echo "head -n -1 ~/.bashrc > temp.txt ; mv temp.txt ~/.bashrc && npx lint-staged;" >> ~/.bashrc;
+  /bin/bash -i
+HEREDOC
+else
+  npx lint-staged
+fi
```
ribeiroguilherme, line 13: Any reason why this is placed inside the 'lib' package? Why not put it and the husky dependency in the root level? It could potentially be used in other packages too

packages/lib/package.json
```diff
@@ -45,7 +45,8 @@
         "type-check-generate": "tsc --emitDeclarationOnly",
         "lint": "eslint 'src/**/*.{js,ts,tsx}' --quiet",
         "lint:fix": "npm run lint -- --fix",
-        "prepublishOnly": "npm run build"
+        "prepublishOnly": "npm run build",
+        "prepare": "cd ../.. && husky install packages/lib/.husky"
```
ribeiroguilherme, line 49: How does it work to setup the commit hooks in our machines with VM? Do we need to run this script/do any other setup or will it work out of the box?

packages/lib/.husky/pre-commit
m1aw: The main reason is that we only had prettier on the lib/ project.

packages/lib/package.json
m1aw: Yeah, it does. There should be no difference since you are running already inside of vagrant. Just the commits are outside.

PR #1873, opened 2022-12-05
Fix/personal details validation

sponglord:
## Summary
Fixes to how the `PersonalDetails` component validates:
- if validation rules were passed as a prop - they weren't being merged with the default rules to ensure all fields were still validated
- errors were being registered but not converted into translatable (and displayed) error messages
- `required` & `aria-required` attributes were not being set on the (required) input fields

## Tested scenarios
All tests pass
Also fixed some e2e tests that were failing because the constant used for the date had "expired"

**Fixed issue**:  Relates to some of the points raised in our a11y review

### DIFF
--- a/packages/e2e/tests/cards/errorPanel/errorPanel.kcpCard.visible.test.js
+++ b/packages/e2e/tests/cards/errorPanel/errorPanel.kcpCard.visible.test.js
@@ -12,7 +12,7 @@ const INVALID_TAX_NUMBER = LANG['creditCard.taxNumber.invalid'];
 
 const cardPage = new CardComponentPage();
 
-fixture.only`Testing kcpCard, with holderName, error panel`
+fixture`Testing kcpCard, with holderName, error panel`
     .beforeEach(async t => {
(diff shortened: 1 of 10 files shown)


packages/e2e/tests/cards/errorPanel/errorPanel.kcpCard.visible.test.js
```diff
@@ -12,7 +12,7 @@ const INVALID_TAX_NUMBER = LANG['creditCard.taxNumber.invalid'];
 
 const cardPage = new CardComponentPage();
 
-fixture`Testing kcpCard, with holderName, error panel`
```
ribeiroguilherme, line 15: I guess you forgot to remove the `only` :)

ribeiroguilherme (approved):

packages/e2e/tests/cards/errorPanel/errorPanel.kcpCard.visible.test.js
sponglord: guess so! Fixed

PR #1878, opened 2022-12-08
Click to Pay - Info modal and new loader

ribeiroguilherme:
## Summary
- Replaced previous CtP loader spinner with a GIF
- Added Info icon which, when shopper clicks on it, it opens a modal sharing what is Click to Pay. 
- *Highlight*: Introduced the `Modal` component which is used for rendering modals:
  - It is not using `dialog` HTML element as it is not compatible with IE11 and Safari < 15.4, but we can iterate on this next to fallback to `dialog` component if browser supports it
  - It uses two hooks: 
    - `useTrapFocus` which traps the focus inside a specific HTML element - this hook can be reused in other components too. 
    - `useModal` which encapsulates the logic of the modal (clicking outside of it, or closing it properly by setting the focus where it should be when the modal closes)

## Tested scenarios
- Tested modal by:
  - Checked that pressing ESC closes it
  - Checked focus is set on the informed element
  - Checked that, once the modal is closed, focus is set back to the expected element
  - Checked that focus is trapped inside the modal


packages/lib/src/components/Card/components/ClickToPay/components/CtPBrand/CtPBrand.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const pipeImageUrl = getImage({ loadingContext, imageFolder: 'components/' })('pipe');
+
+    return (
+        <div
+            className={classnames(
+                'adyen_checkout-ctp__brand-wrapper',
+                classNameModifiers.map(m => `adyen_checkout-ctp__brand-wrapper--${m}`)
+            )}
+        >
+            <Img className="adyen_checkout-ctp__brand-logo" src={ctpImageUrl} alt={'Click to Pay'} />
```
m1aw, line 28: Logo....

m1aw (approved):

PR #1879, opened 2022-12-12
binLookup - pass issuingCountryCode to onBinLookup callback

sponglord:
## Summary
Due to the regulations around recurring in India, merchants need to route payments on Indian issued cards via India local acquiring in India platform. 
To do this merchants need to first identify whether the shopper is using a card issued in India.
To this end we now pass the `issuingCountryCode` property to the merchant defined `onBinLookup` handler

## Tested scenarios
All tests pass

**Fixed issue**:  FOC-68174

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
+++ b/packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
@@ -121,7 +121,8 @@ export default parent => {
                                 // this.props.brands that matches the card
                                 // number that the shopper has typed
                                 supportedBrandsRaw: mappedResponse.supportedBrands, // full supportedBrands data (for customCard comp)
-                                brands: parent.props.brands || DEFAULT_CARD_GROUP_TYPES
+                                brands: parent.props.brands || DEFAULT_CARD_GROUP_TYPES,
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
```diff
@@ -121,7 +121,8 @@ export default parent => {
                                 // this.props.brands that matches the card
                                 // number that the shopper has typed
                                 supportedBrandsRaw: mappedResponse.supportedBrands, // full supportedBrands data (for customCard comp)
-                                brands: parent.props.brands || DEFAULT_CARD_GROUP_TYPES
+                                brands: parent.props.brands || DEFAULT_CARD_GROUP_TYPES,
+                                issuingCountryCode: data.issuingCountryCode
```
ribeiroguilherme, line 125: Do we need to update the `CbObjOnBinLookup` type or is this property already there?

ribeiroguilherme (approved):

packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
sponglord: done

PR #1918, opened 2022-12-22
Remove frontend sorting logic for binLookup

sponglord:
## Summary
Remove frontend sorting logic for binLookup now that this is done on the backend

## Tested scenarios
All tests pass


packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
```diff
@@ -4,7 +4,6 @@ import { DEFAULT_CARD_GROUP_TYPES } from '../lib/configuration/constants';
 import { getError } from '../../../../core/Errors/utils';
 import { ERROR_MSG_UNSUPPORTED_CARD_ENTERED } from '../../../../core/Errors/constants';
 import { BinLookupResponse, BinLookupResponseRaw } from '../../../Card/types';
-import { sortBrandsAccordingToRules } from './sortBinLookupBrands';
```
ribeiroguilherme, line 7: This function can be deleted, right?

ribeiroguilherme (approved):

packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
sponglord: I'm keeping it there for the time being until the backend changes have been fully confirmed to work (no merchant complaints!)

PR #1920, opened 2022-12-22
Adding field specific errors for fields in the PersonalDetails component

sponglord:
## Summary
Adding field specific errors for first & last name fields in the PersonalDetails component now that the translations are available. This brings the errors on these fields into line with a11y guidelines.

## Tested scenarios
Translated field specific errors are used for these fields

**Relates to issue**:  FOC-66770


ribeiroguilherme (approved):
Do you know if we are using same error messages on the mobile SDK's as well?

sponglord:
> Do you know if we are using same error messages on the mobile SDK's as well?

Not to my knowledge, no

PR #1922, opened 2022-12-23
Forwarding Component props when creating Component from Action

ribeiroguilherme:
## Summary
There are two issues when the props of the Component aren't properly passed forward when the Component needs to handle an action.

**First issue (Drop-in):**
When creating with `paymentMethodsConfiguration`:
```js
    ...,
       pix: { countdownTime: 5 }, 
   },
});
```
When handling the action from payments call, a new Pix Component is created to handle the QR code flow. This new Component does not pick its configured property. Only if the merchant defines the code as:

```js
    ...,
       pix: { countdownTime: 5 }, 
       qrCode: { countdownTime: 5 },
   },
});
```

**Second issue (standalone component):**

When creating Pix component passing custom config:
```js
checkout.create('pix', { countdownTime: 5 });
````
Same problem happens as described above - as soon as the Component performs the payment request and receives an `action` response back in order to create the QR code flow, the 'new' Pix Component doesn't pick up the custom config passed once the Component was created. It uses the default value set on the Component which is `coutdownTime: 15 `

**Solution**: 

Before calling the `createFromAction` method, the Element build up the props that will be passed to the new Component. 
In this PR, the `elementRef.props` are also included in the group of props.

## Tested scenarios
- Unit tests
- E2E tests

**Fixed issue**:  COWEB-1155

### DIFF
--- a/packages/lib/src/components/Pix/Pix.tsx
+++ b/packages/lib/src/components/Pix/Pix.tsx
@@ -42,6 +42,8 @@ class PixElement extends QRLoaderContainer<PixProps> {
     }
 
     render() {
+        console.log(this.props.countdownTime);
+
         if (this.props.paymentData) {
             return this.renderQRCode();
         }
--- a/packages/lib/src/components/Pix/Pix.tsx
+++ b/packages/lib/src/components/Pix/Pix.tsx
@@ -41,8 +41,6 @@ class PixElement extends QRLoaderContainer<PixProps> {
     }
 
     render() {
-        console.log(this.props);
-
         if (this.props.paymentData) {
             return this.renderQRCode();
         }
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: packages/lib/src/components/Pix/Pix.tsx [hunk 1]
```diff
@@ -1,33 +1,34 @@
...
 import PixInput from './PixInput';
 import { cleanCPFCNPJ } from '../internal/SocialSecurityNumberBrazil/utils';
-import { PixProps } from './types';
+import { PixElementData, PixProps } from './types';
 
 class PixElement extends QRLoaderContainer<PixProps> {
     public static type = 'pix';
 
     public static defaultProps = {
+        showPayButton: false,
```
m1aw, line 12: Why was this changed? No other payment method defaults to `showPayButton: false`.

m1aw: You are right, I always thought Components had the same default has Dropin. I think we can keep it this way, no reason to change it now.

m1aw (approved):

PR #1924, opened 2022-12-27
Validating postal code when billingAddressMode is partial

ribeiroguilherme:
## Summary
When setting `billingAddressMode` to `partial` , the postal code wasn't being validated as there was no country set.

In this PR, we add support for validating the postal code field for such case. In order to trigger the validation, the country needs to be set in the Component configuration. Example:

```js
const card = checkout.create('card', {
    billingAddressRequired: true,
    billingAddressMode: 'partial',
    data: {
        billingAddress: {
            country: 'US'
        }
    }
});
```

## Tested scenarios
- Added e2e tests

**Fixed issue**:  COWEB-1138

### DIFF
--- a/packages/e2e/tests/cards/avs/avs.partial.test.js
+++ b/packages/e2e/tests/cards/avs/avs.partial.test.js
@@ -1,0 +1,45 @@
+import { CARDS_URL } from '../../pages';
+import { start } from '../../utils/commonUtils';
+import { REGULAR_TEST_CARD } from '../utils/constants';
+import CardComponentPage from '../../_models/CardComponent.page';
+import { CLIENTSCRIPT_PARTIAL_AVS_WITH_COUNTRY, CLIENTSCRIPT_PARTIAL_AVS_WITHOUT_COUNTRY } from './avs.partial.clientScripts';
+
+const TEST_SPEED = 1;
+const INVALID_POSTALCODE = 'aaaaaaaaaa';
+
+let cardPage = null;
+
+fixture`Card with Partial AVS`.page(CARDS_URL).beforeEach(() => {
+    cardPage = new CardComponentPage();
+});
+
+test('should validate Postal Code if property data.billingAddress.country is provided', async t => {
+    // Start, allow time for iframes to load
+    await start(t, 2000, TEST_SPEED);
+
+    await cardPage.cardUtils.fillCardNumber(t, REGULAR_TEST_CARD);
+    await cardPage.cardUtils.fillDateAndCVC(t);
+
+    // Fill in wrong Postal code format
+    await t.typeText(cardPage.postalCodeInput, INVALID_POSTALCODE);
+    await t.click(cardPage.payButton);
+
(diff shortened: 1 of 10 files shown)


packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -75,6 +76,8 @@ const CardInput: FunctionalComponent<CardInputProps> = props => {
     const showBillingAddress = props.billingAddressMode !== AddressModeOptions.none && props.billingAddressRequired;
 
     const partialAddressSchema = handlePartialAddressMode(props.billingAddressMode);
+    // Keeps the value of the country set initially by the merchant, before the Address Component mutates it
+    const partialAddressCountry = useRef<string>(partialAddressSchema && props.data?.billingAddress?.country);
```
m1aw, line 80: I think we should default to `US` in case of partial address mode. This rule was designed for `US` merchants in mind and I think this makes the change safer.

packages/e2e/tests/cards/avs/avs.partial.test.js
```diff
@@ -0,0 +1,41 @@
...
+    await start(t, 2000, TEST_SPEED);
+
+    await cardPage.cardUtils.fillCardNumber(t, REGULAR_TEST_CARD);
+    await cardPage.cardUtils.fillDateAndCVC(t);
+
+    await t.typeText(cardPage.postalCodeInput, INVALID_POSTALCODE);
+    await t.click(cardPage.payButton);
+
+    // Check the value of the alert text
+    const history = await t.getNativeDialogHistory();
```
m1aw, line 39: I'm not sure if this a good default?
Shouldn't we maybe throw and error that is not configured or default to some value?

packages/lib/src/components/Card/components/CardInput/CardInput.tsx
ribeiroguilherme: Not really - if you check the scope of the internal ticket, it was reported by one merchant from Brazil. In the comments you can also see that, if no country is set, then we don't apply any validation.

packages/e2e/tests/cards/avs/avs.partial.test.js
ribeiroguilherme: I didn't really understand what you meant by 'good default'. Can you elaborate?

If the merchant didn't configure the partial billing address validation by setting the country, then the UI will accept anything.

m1aw: I think we should throw some kind of error. Otherwise most merchants won't notice there's validation since it's not enabled by default.

ribeiroguilherme: Starting throwing an error would actually break things. We can't throw an error if we don't know the country, and if we assume that the country would be `US` by default (as partial address is mainly used there), it would break integrations done by brazilian merchants for example.

There is a ticket for the docs team to document this and explain how the validation is be done with partial address mode, to avoid confusion.

m1aw (approved):

PR #1925, opened 2022-12-28
Adding buttonRootNode property to GooglePay

ribeiroguilherme:
## Summary

Adding support for [buttonRootNode](https://developers.google.com/pay/api/web/reference/request-objects#ButtonOptions) property for GooglePay.

**Fixed issue**:  COWEB-1167


ribeiroguilherme:
SonarCloud scan failed as it complained about using `HTMLDocument` type which is deprecated, but I will keep as it is as the GooglePay types still uses it.

m1aw (approved):

PR #1928, opened 2023-01-05
Stop calling onError during 3DS2 fingerprinting

sponglord:
## Summary
When performing a 3DS2 fingerprint it is possible for the process to timeout - in which case we generate an error and call the merchant defined `onError` callback.
This error is not _fatal_ and the 3DS2 process can continue.
However it seems some merchants are treating it as fatal and from their `onError` callback abort the whole Checkout process.
To combat this, this PR no longer calls the  `onError` callback for the "timeout" scenario

## Tested scenarios
`onError` no longer called when fingerprinting times-out
All e2e tests pass

**Relates to issue**:  FOC-68756

### DIFF
--- a/packages/lib/src/components/ThreeDS2/components/DeviceFingerprint/PrepareFingerprint3DS2.tsx
+++ b/packages/lib/src/components/ThreeDS2/components/DeviceFingerprint/PrepareFingerprint3DS2.tsx
@@ -36,6 +36,7 @@ class PrepareFingerprint3DS2 extends Component<PrepareFingerprint3DS2Props, Prep
         // If no fingerPrintData or no threeDSMethodURL - don't render component. Instead exit with threeDSCompInd: 'U'
         if (!this.state.fingerPrintData || !this.state.fingerPrintData.threeDSMethodURL) {
             this.setStatusComplete({ threeDSCompInd: 'U' });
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/ThreeDS2/components/DeviceFingerprint/PrepareFingerprint3DS2.tsx
```diff
@@ -21,7 +21,12 @@ class PrepareFingerprint3DS2 extends Component<PrepareFingerprint3DS2Props, Prep
             };
         } else {
             this.state = { status: 'error' };
-            this.props.onError('Missing fingerprintToken parameter');
+            // TODO - confirm that we should do this, or is it possible to proceed to the challenge anyway?
```
m1aw, line 24: How do we verify this TODO?

m1aw (approved):

packages/lib/src/components/ThreeDS2/components/DeviceFingerprint/PrepareFingerprint3DS2.tsx
sponglord: I'm awaiting an answer. But for now we can proceed (and just keep this here as a reminder)

PR #1937, opened 2023-01-12
Accessibility: Select combobox rework

m1aw:
## Summary

The objective of this PR is to update our current Select/Combobox field and bring it up to the expected accessibility standards. 
The main work with this has to do with keeping the focus always on the input. This allows keyboard/accessible technology users to interact with the search by writing while at the same time using the keyboard to navigate between options.
Previously this behaviour was impossible since the user would have to opt between searching, and choosing one of the options, making it very hard to use with accessible technologies since the user doesn't have a full overview of the 'possible' options.

## Tested scenarios

The main components tested where:
- Card component (with instalments)
- Adress Component (standalone)
- Issuer List (iDeal)
- Kiwi Component (OldPhoneNumber)
- MBway (NewPhoneNumber)
- OpenInvoice (readonly mode)

**Fixed issue**:


sponglord:
Overall it looks great.
A _minor_ thing, one bit of UX I am having trouble with is in the following scenario:
- click in a dropdown to open it
- start typing to search
- use arrow keys + Enter to make a selection
- cursor stays active in the search field
- clicking again does nothing - but I would expect it to open the dropdown list again (which is what happens if you click Enter or Space)

sponglord (approved):

sponglord:
I don't know if the failing playwright tests are significant? (They all seem to be related to the issuerList)

m1aw:
@sponglord the IssuerList tests where indeed relevant, it's now fixed

PR #1938, opened 2023-01-13
Feature/a11y error panel improvements

sponglord:
## Summary
Our third party a11y assessment pointed out a number of areas for improvement - one of which was how we handle errors across components and how those errors are themselves handled by a Screen reader (SR).

In this PR we take on the suggestions from the assessment which revolve around:
- having the element that will be used to provide messages for the screen readers (the `SRPanel`) to be present in the DOM from first page load
- to differentiate between "on input" and "on submit" errors:
    - **"on input" errors** happen as the user is inputting into a field. 
    - The errors are written into an `aria-describedby` element for that field. This acts as both a visual error and one that the SR can read. 
    - It also means that when this field is clicked back into, whilst it is in error, the SR can use the `aria-describedby` element to re-read the error.
    - **"on submit" errors** happen as a user leaves the field (the "on blur" scenario) or when the user presses the pay button. 
    -  These appear visually in the `aria-describedby` element for that field, but are also written to the `SRPanel` which has an `aria-live="polite"` attribute, causing them to be read out. 
    - In the case of "onblur" errors it means that these errors are still read by the SR even though focus has moved to another field. 
    - In the case of "onsubmit" errors, where multiple fields could be in error, all these errors are collated into the SRPanel where they are all read out.
- in the "on submit" error scenario it is acceptable for focus to then be programatically shifted to the first element in error
- iframes, i.e. the securedFields, should maintain their own `aria-describedby` element within their own local document/DOM. For these fields the _visual_ error element, found within the Checkout UI, has an `aria-hidden="true"` attribute to avoid the risk of the SR reading the error twice.

## Notes for reviewers
- It should be noted that a lot of the functionality for this PR was already in place, tested and working, in the `ErrorPanel` (the forerunner to the `SRPanel`) which was used by the `CardInput` component. This included:
    -  collating errors from multiple components
    - allowing for regional variations on the error message texts
    - ordering these errors based on how the fields were presented in the UI. 
    - setting focus on the first field in error
These are all things that can change at runtime, based on user interaction with the component.
- This is a _massive_ PR in terms of the number of files changed.
- However many of the changes are superficial (related to prop-drilling (or removal of), or e2e testing)
- _The key changes/files-of-interest are as follows_:
    - `core.ts` which creates the `SRPanel` as a module, making it available to all components.
    - `SRPanel.tsx` (and it's child `SRMessages.tsx`) which accept a config object, `srConfig`, a property on the Checkout config object. This takes the default form of: 
    ```srConfig: { enabled: true, showPanel: true, node: 'body', moveFocus:true }```
    which allows us to specify whether:
        -  the SRPanel is used at all
        - whether it is visible
        - which DOM element it is added to at page load 
        - and, whether focus should be shifted to the first element in error when the "Pay" button is pressed.
    - The universally available `SRPanel` module exposes two methods:
        - `setMessages` which accepts an array of strings that it will place into the `SRPanel`  DOM element
        - `moveFocus`, a getter, which allows a component to know whether it is expected to shift focus to an element in the "onsubmit" scenario
    - `CardInput.tsx` which now implements the `SRPanel` instead of the `ErrorPanel`
    - `internal/PersonalDetails.tsx` which implements the `SRPanel`
    - `internal/Address.tsx` which implements the `SRPanel`
    - `internal/OpenInvoice.tsx` which implements the `SRPanel`. This is a similar usecase to the `CardInput` i.e. multiple child components with a dynamically changing set of errors, error texts, and UI
    - `core/Errors/utils.ts` which provides a set of generic utils for a component to use, to build and order error objects and pass them to the `SRPanel.setMessages` method

## Tested scenarios
A new unit test has been created for the `SRPanel/SRMessages` components.
All unit tests, new & pre-existing, pass
Existing e2e tests, which worked against the forerunner to the `SRPanel`, the `ErrorPanel`, have been modified to test the functionality of the `aria-describedby` / `SRPanel` way of presenting errors.
All e2e tests pass.


sponglord:
Reopened, as draft, since at the moment this is still the master branch for _all_ the changes

PR #1957, opened 2023-01-23
Removed auto redirect from QR code loader

jobsturm:
## Summary
Removed the auto redirect from QR code loader on mobile. Instead the user will now have to use the "open app" button below the QR code.

## Tested scenarios
- [x] Tested in the playground.


packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -106,15 +106,7 @@ function Await(props: AwaitComponentProps) {
     };
 
     useEffect(() => {
-        const { shouldRedirectOnMobile, url } = props;
```
ribeiroguilherme, line 109: [shouldRedirectOnMobile](https://github.com/Adyen/adyen-web/blob/4d55ec1349a83fb6694238c2eebb2a16f47a7add/packages/lib/src/components/internal/Await/Await.tsx#L223)  still defined as defaultProps in the `Await.defaultProps = {...` . It should be removed

packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -106,15 +106,7 @@ function Await(props: AwaitComponentProps) {
...
-        const isMobile: boolean = window.matchMedia('(max-width: 768px)').matches && /Android|iPhone|iPod/.test(navigator.userAgent);
-
-        if (shouldRedirectOnMobile && url && isMobile) {
-            redirectToApp(url, checkStatus);
```
ribeiroguilherme, line 113: In the [redirectToApp](https://github.com/Adyen/adyen-web/blob/4d55ec1349a83fb6694238c2eebb2a16f47a7add/packages/lib/src/components/internal/Await/Await.tsx#L104) function, the `setTimeout` can be removed (same as you did in the QRLoader.tsx)

ribeiroguilherme (changes requested):

ribeiroguilherme (approved):

PR #1963, opened 2023-01-26
IbanInput propagates errors including on showValidation

sponglord:
## Summary
The `IbanInput` component wasn't propagating its errors to its parent component - neither when an input value changed nor when `showValidation` was called on the component

## Tested scenarios
`IbanInput` propagates errors to its parent (tested in both `Sepa` & `OpenInvoice` components)
All unit tests pass

### DIFF
--- a/packages/lib/src/components/internal/IbanInput/IbanInput.tsx
+++ b/packages/lib/src/components/internal/IbanInput/IbanInput.tsx
@@ -82,7 +82,7 @@ class IbanInput extends Component<IbanInputProps, IbanInputState> {
         const holderNameValid = this.props.holderName ? isValidHolder(this.state.data['ownerName']) : true;
         const ibanValid = checkIbanStatus(this.state.data['ibanNumber']).status === 'valid';
         const isValid = ibanValid && holderNameValid;
-        const data = { data: this.state.data, isValid };
+        const data = { data: this.state.data, isValid, errors: this.state.errors };
 
         this.props.onChange(data);
     }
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/internal/IbanInput/IbanInput.tsx
```diff
@@ -32,6 +33,18 @@ interface IbanInputState {
...
 
+const ibanHolderNameErrorObj: GenericError = {
+    isValid: false,
+    errorMessage: 'ach.accountHolderNameField.invalid', // TODO create bespoke translation key
+    error: 'ach.accountHolderNameField.invalid'
+};
+
+const ibanErrorObj: GenericError = {
+    isValid: false,
+    errorMessage: 'sepaDirectDebit.ibanField.invalid',
```
ribeiroguilherme, line 44: It looks weird to have both `error` and `errorMessage` containing the same values.  I'd expect the`errorMessage` to already has the error message and not the i18n error key value. 

Can't error message be derived from the `error` property?

ribeiroguilherme (approved):

packages/lib/src/components/internal/IbanInput/IbanInput.tsx
sponglord: I agree. And I got rid of it. Then I found that the visual errors stopped showing up. It seems that the visual errors rely on the `error` prop and the screenreader messages derive from the `errorMessage` prop.
This is something that can hopefully be fixed in a future iteration.

PR #1968, opened 2023-01-27
Fix/remove this keyword from functional comps

sponglord:
## Summary
Some of the components that are part of the changes for the new Screen Reader panel (aka SRPanel) are functional components, yet they still rely on the use of the `this` keyword (a legacy from when they were Class components).

This PR proceeds the SRPanel PR - replacing the use of `this` in the affected components.

Instead of relying on a `ref` to the overall component, we pass in a `setComponentRef` prop that allows the component to pass back an object on which it can expose members that the parent UIElement expects - e.g. `showValidation` & `setStatus`.
(This is the same solution that has been used in the `CardInput` solution for the last year or so)

## Tested scenarios
All unit & e2e tests pass


packages/lib/src/components/Ach/components/AchInput/types.ts
```diff
@@ -58,4 +58,11 @@ export interface ACHInputProps {
     styles?: StylesObject;
     type?: string;
     forceCompat?: boolean;
+    setComponentRef?: (ref) => void;
+}
+
+// An interface for the members exposed by a component to its parent UIElement
+export interface AchRef {
+    showValidation?: (who) => void;
```
ribeiroguilherme, line 66: Regarding this `who` object - do we ever pass something to the `showValidation` function? Perhaps we don't need it

packages/lib/src/components/Ach/components/AchInput/types.ts
```diff
@@ -58,4 +58,11 @@ export interface ACHInputProps {
...
+    setStatus?: any;
```
ribeiroguilherme, line 67: I believe the setStatus  could be defined as:
```ts
setStatus?(status: UIElementStatus): void
```
I added similar type definition for the UPI pm a while ago, which I expect to be the case for all payment methods

packages/lib/src/components/Boleto/types.ts
```diff
@@ -30,3 +30,9 @@ export interface BoletoInputErrorState {
     shopperEmail?: boolean;
     socialSecurityNumber?: boolean;
 }
+
+// An interface for the members exposed by a component to its parent UIElement
+export interface BoletoRef {
+    showValidation?: (who) => void;
+    setStatus?: any;
```
ribeiroguilherme, line 37: As this is a pattern that applies to most (if not all) of Preact Components that the UIElement interacts directly, perhaps we could try to create an interface and reuse it across different Components. 

Ex:

```ts
interface ComponentMethodsRef {
   showValidation?(...
   setStatus?(...
   ...
}
```

```jsx
const boletoRef = useRef<ComponentMethodsRef>({});
```

Maybe I'm  overengineering here 😅  But this pattern is very repetitive. It feels like can be improved somehow

ribeiroguilherme (approved):
Is this PR supposed to change all functional components? 

I'm asking because I see that the UPI wasn't affected by this PR, although the UPI uses `this` inside the `UPIComponent`.

packages/lib/src/components/Ach/Ach.tsx
```diff
@@ -64,6 +64,10 @@ export class AchElement extends UIElement<AchElementProps> {
         return this.props.name;
     }
 
+    public setComponentRef = ref => {
```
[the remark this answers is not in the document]
sponglord: Agreed

packages/lib/src/components/Ach/components/AchInput/types.ts
sponglord: Yeah, I think you're right

packages/lib/src/components/Ach/components/AchInput/types.ts
sponglord: ok, cool

packages/lib/src/components/Boleto/types.ts
sponglord: Yup, good idea!

PR #1969, opened 2023-01-27
Adding build step

dsaouda:
## Summary

I was testing the project and noticed that the readme does not have the build step. I thought it was cool to add

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,11 @@ To run the development environment:
   ```sh
   yarn install
   ```
-4. Run the development environment, which starts a server listening on [http://localhost:3020](http://localhost:3020):
+4. Run build
+  ```sh
+  yarn build
+  ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,11 @@ To run the development environment:
   ```sh
   yarn install
   ```
-4. Run the development environment, which starts a server listening on [http://localhost:3020](http://localhost:3020):
+4. Run build
```
ribeiroguilherme, line 52: Hi @dsaouda . Thanks for the contribution. There are some errors indeed when you run by the first time, but that doesn't break the development environment.  Perhaps it would make more sense to point out that the build script needs to be ran only by the first time. Can you update that?

```suggestion
4. If you are running the project by the first time, run the build script
```

ribeiroguilherme (changes requested):

ribeiroguilherme (approved):

Issue #941, opened 2021-04-09
Fire event when additional action has loaded successfully


bitbytebit1:
**Describe the bug**
No event is fired when the Swish QR code is ready

**To Reproduce**
On desktop
Select Swish as a payment method
Click pay
A QR code is displayed 

**Expected behavior**
On desktop
Select Swish as a payment method
Click pay
A QR code is displayed 
An event is fired indicating the QR code has loaded successfully 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Desktop

**Additional context**
We would like to scroll the QR code into view, currently we are forced to wait arbitrary amount of time

marcperez:
Hi @bitbytebit1,

Thanks for reaching. This seems like an idea worth exploring.
We will discuss internally and come back to you as soon as possible.

In the meantime, could you maybe tell us a bit more about an example use case where such event could be useful?

bitbytebit1:
Here is an example of how we are getting around this currently.
If dropin.handleAction returned a promise that resolved once the QR code had finished booting up then we could simply wait for it to resolve and perform any additional actions needed then

```onSubmit: (state, dropin) => {
  this.debug && console.log('ADYEN.onSubmit: ', state);
  if (this.ADYEN_DROPIN?.data?.paymentMethod?.type === 'applepay' && !this.ADYEN_DROPIN?.data?.paymentMethod?.applePayToken) {
    console.debug('Apple pay payment does not have token! Cancel onSubmit');
    return;
  }
  this.apiMakePayment(state.data)
    .then((response) => {
      if (response.action) {
        dropin.handleAction(response.action);
        if (response.action.paymentMethodType === 'swish' && response.action.type === 'qrCode') {
          // Here we are forced to wait an abitrary amount of time
          setTimeout(() => {
            this.elems.dropinContainer.scrollIntoView();
          }, 100);
        }
      } else {
        this.redirectToPaymentCompletePage(response);
      }
    })
}```

marcperez:
Thanks for the example @bitbytebit1, that makes sense. We'll discuss internally and update this ticket as soon as possible.

sponglord:
@bitbytebit1 - the next release of adyen-web will contain this PR: #1975 
In keeping with our current policy of providing callbacks that you can add to the configuration for checkout we have added an `onActionHandled` callback (details in the [PR](https://github.com/Adyen/adyen-web/pull/1975))

PR #1971, opened 2023-01-31
After handleAction retrieve Promise to know when action has been performed

sponglord:
## Summary
When `handleAction` or `createFromAction` is called a reference to the component is returned.

Now it is possible to call `getStatus` on this component instance which will return a Promise that will resolve when the action, or some event that is part of this action, has been performed.

In this example PR we resolve this Promise when the QR code image has loaded.

In theory any component can use this functionality to inform the merchant when an action has been implemented.
For example, I've also tested it with the card component - using it to know when the 3DS2 component has been rendered.

## Tested scenarios
Tested with both `handleAction` and `createFromAction` against both Dropin and a standalone component

To test 
- either run the Dropin and select the first WeChatPay pm and then press "continue..."
- or, go to the QRCodes playground page
- then look at the console to see how the promise is resolved when the QR code _loads_

**Relates to issue**:  #941

### DIFF
--- a/packages/lib/src/components/UIElement.tsx
+++ b/packages/lib/src/components/UIElement.tsx
@@ -16,6 +16,8 @@ import Core from '../core';
 export class UIElement<P extends UIElementProps = any> extends BaseElement<P> implements IUIElement {
     protected componentRef: any;
     public elementRef: UIElement;
+    public status: UIElementStatus;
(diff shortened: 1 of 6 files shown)


sponglord:
Abandoned in favour of the `onActionHandled` callback (#1975)

PR #1975, opened 2023-02-01
Added onActionHandled callback

sponglord:
## Summary
This PR adds a callback `onActionHandled` which can be listened to to know that an action has proceeded as planned.

For example: 
- for an action that leads to the presentation of a QR code we call this callback when the QR code image has loaded
- similarly for a 3DS2 action we call it when the generated iframe has loaded its content
- and for something that leads to an `Await` action (UPI, Blik, MBWay) we call it once the `/status` polling starts

What is received within this handler is an object describing the component and the action that has been handled,
examples:
- `{componentType: 'wechatpayQR', actionDescription: 'qr-code-loaded'}`
- `{componentType: '3DS2Fingerprint', actionDescription: 'fingerprint-iframe-loaded' }`
- `{componentType: 'mbway', actionDescription: 'polling-started'}`

## Tested scenarios
Tested both the described scenarios from within Dropin, and also from within the respective standalone components

**Relates to issue**:  #941 

This also relates to the many queries we have received about whether it is possible to know that a 3DS action has actually taken place as expected

### DIFF
--- a/packages/lib/src/components/types.ts
+++ b/packages/lib/src/components/types.ts
@@ -81,6 +81,11 @@ export type UIElementStatus = 'ready' | 'loading' | 'error' | 'success';
 
 export type PayButtonFunctionProps = Omit<PayButtonProps, 'amount'>;
 
+export interface ActionHandledReturnObject {
+    componentType: string;
+    actionType: string;
+}
+
 export interface UIElementProps extends BaseElementProps {
     session?: Session;
     onChange?: (state: any, element: UIElement) => void;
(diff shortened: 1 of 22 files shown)


packages/lib/src/components/types.ts
```diff
@@ -81,13 +81,19 @@ export type UIElementStatus = 'ready' | 'loading' | 'error' | 'success';
 
 export type PayButtonFunctionProps = Omit<PayButtonProps, 'amount'>;
 
+export interface ActionHandledReturnObject {
+    componentType: string;
+    actionType: string;
```
ribeiroguilherme, line 86: Maybe we could define the set of possible values here for the `actionType` ? 
It will make more clear for who is implementing this callback to know what to expect from the `actionType` property

Ex:
```ts
actionType: 'qr-code-loaded' | 'polling-started' ...
```

ribeiroguilherme (approved):

packages/lib/src/components/types.ts
sponglord: Changed `actionType` to `actionDescription` and give it a type (`ActionDescriptionType`)

PR #1982, opened 2023-02-03
CI pipeline: updated to Node 16

ribeiroguilherme:
## Summary
- Updating scripts to use node 16

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -10,10 +10,10 @@ jobs:
     steps:
 
     - uses: actions/checkout@v3
-    - name: Use Node.js 14.x
+    - name: Use Node.js 16.x
       uses: actions/setup-node@v3
       with:
-        node-version: 14.x
+        node-version: 16.x
 
     - name: install, coverage
       run: |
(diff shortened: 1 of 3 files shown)


packages/lib/src/language/utils.ts
```diff
@@ -113,11 +113,12 @@ export const getTranslation = (translations: object, key: string, options: { [ke
...
 /**
- * Returns an array with all the locales
+ * Returns an Object which contains all the key/values of the translation labels
+ *
  * @param locale - The locale the user wants to use
- * @param customTranslations -
+ * @param customTranslations - Custom translations provided by the merchant
  */
-export const loadTranslations = async (locale: string, customTranslations: object = {}) => {
+export const loadTranslations = async (locale: string, customTranslations: object = {}): Promise<Record<string, string>> => {
```
sponglord, line 121: Is this really the right return Type?

sponglord (approved):

packages/lib/src/language/utils.ts
ribeiroguilherme: Yes. It is asynchronous and it returns the merged `translations` object

Issue #1981, opened 2023-02-03
Expiry Hint Image Hardcoded to Adyen Test domain (checkoutshopper-test.adyen.com) 


nicangeli:
**Describe the bug**

The expiry date image is hardcoded to checkoutshopper-test, other images load from checkoutshopper-live, causing our CSP to refuse to load this image. 

https://github.com/Adyen/adyen-web/blame/master/packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx#L68

**To Reproduce**
1. Go to Adyen Demo Test shop checkout page add something to your basked
2. Click on checkout
3. Click on Credit Card
4. observe that the expiry date image is being loaded from checkoutshopper-test not checkoutshopper-live.adyen.com

**Expected behavior**
On test domains images should be loaded from checkoutshopper-test, in production they should be loaded from checkoutshopper-live.adyen.com

![image](https://user-images.githubusercontent.com/1451735/216569350-ce7917a0-3fcd-4a72-b4b6-6776be0babc2.png)

**Desktop (please complete the following information):**
 - OS: All
 - Browser All
 - Version All

**Smartphone (please complete the following information):**
 - Device: All
 - OS: All
 - Browser All
 - Version All

**Additional context**

PR #1986, opened 2023-02-07
Removed hardcoded url on expiryDate hint icon

sponglord:
## Summary
The url for the icon used as an expiryDate hint was hardcoded to the Test environment.
It now uses the dynamic `loadingContext` to set the root for this url

## Tested scenarios
expiryDate hint icon loads based on `loadingContext`

**Fixed issue**:  #1981

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx
@@ -12,6 +12,7 @@ import {
     DATE_POLICY_REQUIRED,
     ENCRYPTED_EXPIRY_DATE
 } from '../../../../internal/SecuredFields/lib/configuration/constants';
+import getImage from '../../../../../utils/get-image';
 
 export default function ExpirationDate(props: ExpirationDateProps) {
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.tsx
```diff
@@ -57,15 +58,13 @@ export default function ExpirationDate(props: ExpirationDateProps) {
...
-                }
-            )}>
+            <div
+                className={classNames('adyen-checkout__field__exp-date_hint_wrapper', [styles['checkout__field__exp-date_hint_wrapper']], {
+                    'adyen-checkout__field__exp-date_hint_wrapper--hidden': error || isValid
+                })}
+            >
                 <img
-                    src="https://checkoutshopper-test.adyen.com/checkoutshopper/images/components/expiry_date_hint.svg"
+                    src={`${loadingContext}images/components/expiry_date_hint.svg`}
```
m1aw, line 67: We should handle this with .getImage() and never directly with URL
Going this way we make it harder for the CDN changes.

sponglord: Fair enough!

Issue #1952, opened 2023-01-20
Attribute `onPaymentCompleted` and `onError` not available in type `CoreOptions`


nam-truong-le:
**Describe the bug**
Attribute `onPaymentCompleted` and `onError` not available in type `CoreOptions`

```javascript
export interface CoreOptions {
    session?: any;
    /**
     * Use test. When you're ready to accept live payments, change the value to one of our {@link https://docs.adyen.com/checkout/drop-in-web#testing-your-integration | live environments}.
     */
    environment?: 'test' | 'live' | 'live-us' | 'live-au' | 'live-apse' | 'live-in' | string;
    /**
     * A public key linked to your web service user, used for {@link https://docs.adyen.com/user-management/client-side-authentication | client-side authentication}.
     */
    clientKey?: string;
    /**
     * The shopper's locale. This is used to set the language rendered in the UI.
     * For a list of supported locales, see {@link https://docs.adyen.com/checkout/components-web/localization-components | Localization}.
     * For adding a custom locale, see {@link https://docs.adyen.com/checkout/components-web/localization-components#create-localization | Create localization}.
     * @defaultValue 'en-US'
     */
    locale?: Locales | string;
    /**
     * Custom translations and localizations
     * See {@link https://docs.adyen.com/checkout/components-web/localization-components | Localizing Components}
     */
    translations?: CustomTranslations;
    /**
     * The full `/paymentMethods` response
     */
    paymentMethodsResponse?: PaymentMethodsResponseObject;
    /**
     * Amount of the payment
     */
    amount?: PaymentAmountExtended;
    /**
     * Secondary amount of the payment - alternative currency & value converted according to rate
     */
    secondaryAmount?: PaymentAmountExtended;
    /**
     * The shopper's country code. A valid value is an ISO two-character country code (e.g. 'NL').
     */
    countryCode?: string;
    /**
     * Optional per payment method configuration
     */
    paymentMethodsConfiguration?: PaymentMethodsConfiguration;
    /**
     * Display only these payment methods
     */
    allowPaymentMethods?: string[];
    /**
     * Never display these payment methods
     */
    removePaymentMethods?: string[];
    analytics?: AnalyticsOptions;
    risk?: RiskModuleOptions;
    order?: Order;
    [key: string]: any;
}

```

**To Reproduce**

**Expected behavior**
Attributes are available as in [documentation](https://docs.adyen.com/online-payments/web-drop-in#-adyencheckout-configuration)

**Screenshots**

[...]

PR #1987, opened 2023-02-07
Adjusting types of CoreOptions

ribeiroguilherme:
## Summary
- Added missing types of CoreOptions object

**Fixed issue**: #1952


packages/lib/src/core/core.ts
```diff
@@ -59,44 +65,23 @@ class Core {
...
+        if (this.options.onAdditionalDetails) {
+            return this.options.onAdditionalDetails(details);
+        }
 
         if (this.session) {
             this.session
                 .submitDetails(details)
                 .then(response => {
-                    if (this.options.onPaymentCompleted) this.options.onPaymentCompleted(response, this);
+                    this.options.onPaymentCompleted?.(response);
```
longyulongyu, line 81: If we remove `this`, it might break some merchants who are using the element ref. Same goes for `this.options.onError?.(error);`.

longyulongyu: I see, thanks for the explanation. It makes sense to omit it.

longyulongyu (approved):

PR #1994, opened 2023-02-17
Fix card brand image alignment

lindelleric:
## Summary

I've fixed a case where the card brand images are not correctly aligned. Like so: 

image

Probably due to a combination of the margin here, plus the `margin-top: -8px` of the brand image wrapper. 

image

Setting the images as `display: block` instead of the default `inline` is one solution to the problem. Another viable solution would be to set `vertical-align: top` on the image element. 

image

## Tested scenarios

**Fixed issue**:


ribeiroguilherme:
Closing this PR for now, as we can't reproduce this problem.

PR #2008, opened 2023-02-28
Optional Chaining for Blik Code

bordeo:
## Summary

When access Blik `component.data` before component rendered it throw an error because `state` is undefined during `formatData` call.
It works only if BlikInput is rendered and onChange runs inside useEffect at least once.

## Tested scenarios

```js
const component = checkout.create('blik', {});
component.mount('.blik-field');
console.log(component.data); // throw error
```

**Fixed issue**:


ribeiroguilherme (approved):

ribeiroguilherme:
Nice catch @bordeo , thanks for the contribution.

PR #2037, opened 2023-03-14
New, more accessible, translations following a11y guidelines

sponglord:
## Summary
Following our recent a11y assessment, and in consultation with our UX copywriting team, some improved translations relating, mainly, to the credit card component


sponglord:
Closing because translations files are missing some translations due to Smartling/line-endings error

PR #2044, opened 2023-03-16
Feature/SRPanel a11y improvements

sponglord:
## Summary
Our third party a11y assessment pointed out a number of areas for improvement - one of which was how we handle errors across components and how those errors are themselves handled by a Screen reader (SR).

In this PR we take on the suggestions from the assessment which revolve around:
- having the element that will be used to provide messages for the screen readers (the `SRPanel`) to be present in the DOM from first page load
- to differentiate between "on input" and "on submit" errors:
    - **"on input" errors** happen as the user is inputting into a field. 
    - The errors are written into an `aria-describedby` element for that field. This acts as both a visual error and one that the SR can read. 
    - It also means that when this field is clicked back into, whilst it is in error, the SR can use the `aria-describedby` element to re-read the error.
    - **"on submit" errors** happen as a user leaves the field (the "on blur" scenario) or when the user presses the pay button. 
    -  These appear visually in the `aria-describedby` element for that field, but are also written to the `SRPanel` which has an `aria-live="polite"` attribute, causing them to be read out. 
    - In the case of "onblur" errors it means that these errors are still read by the SR even though focus has moved to another field. 
    - In the case of "onsubmit" errors, where multiple fields could be in error, all these errors are collated into the SRPanel where they are all read out.
- in the "on submit" error scenario it is acceptable for focus to then be programatically shifted to the first element in error
- iframes, i.e. the securedFields, should maintain their own `aria-describedby` element within their own local document/DOM. For these fields the _visual_ error element, found within the Checkout UI, has an `aria-hidden="true"` attribute to avoid the risk of the SR reading the error twice.

## Notes for reviewers
- It should be noted that a lot of the functionality for this PR was already in place, tested and working, in the `ErrorPanel` (the forerunner to the `SRPanel`) which was used by the `CardInput` component. This included:
    -  collating errors from multiple components
    - allowing for regional variations on the error message texts
    - ordering these errors based on how the fields were presented in the UI. 
    - setting focus on the first field in error
These are all things that can change at runtime, based on user interaction with the component.
- This is a _massive_ PR in terms of the number of files changed.
- However many of the changes are superficial (related to prop-drilling (or removal of), or e2e testing)
- _The key changes/files-of-interest are as follows_:
    - `core.ts` which creates the `SRPanel` as a module, making it available to all components.
    - `SRPanel.tsx` (and it's child `SRMessages.tsx`) which accept a config object, `srConfig`, a property on the Checkout config object. This takes the default form of: 
    ```srConfig: { enabled: true, showPanel: true, node: 'body', moveFocus:true }```
    which allows us to specify whether:
        -  the SRPanel is used at all
        - whether it is visible
        - which DOM element it is added to at page load 
        - and, whether focus should be shifted to the first element in error when the "Pay" button is pressed.
        (only the `moveFocus` prop is really of interest to merchants, the first 3 are primarily to help with development/testing)
    - The `SRPanel` module exposes two methods:
        - `setMessages` which accepts an array of strings that it will place into the `SRPanel`  DOM element
        - `moveFocus`, a getter, which allows a component to know whether it is expected to shift focus to an element in the "onsubmit" scenario
    - `SRPanelProvider` & `SRPanelContext` which encapsulate the functionality of the `SRPanel` and expose it to any "interested" components via a custom hook:  `useSRPanelContext`
    - `CardInput.tsx` which now implements the `SRPanel` instead of the `ErrorPanel`
    - `core/Errors/utils.ts` which provides a set of generic utils for a component to use, to build and order error objects and pass them to the `SRPanel.setMessages` method

## Tested scenarios
A new unit test has been created for the `SRPanel/SRMessages` components.
All unit tests, new & pre-existing, pass
Existing e2e tests, which worked against the forerunner to the `SRPanel`, the `ErrorPanel`, have been modified to test the functionality of the `aria-describedby` / `SRPanel` way of presenting errors.
All e2e tests pass.


packages/lib/src/components/Card/Card.tsx
```diff
@@ -234,11 +225,7 @@ export class CardElement extends UIElement<CardElementProps> {
 
     render() {
         return (
-            <CoreProvider
-                i18n={this.props.i18n}
-                loadingContext={this.props.loadingContext}
-                commonProps={{ isCollatingErrors: this.props.SRConfig.collateErrors }}
-            >
+            <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext}>
```
ribeiroguilherme, line 228: I don't know how it is going to be when adding this feature to other payment methods and how much prop drilling will happen, but did you consider to create a provider/context exclusively for the screen reader mechanism? 

Ex:
```jsx
<SRPanelProvider srPanel={this.props.modules.srPanel} >
...
</SRPanelProvider>
```

Then, on the UI components, you could use a custom hook:

```jsx
const { srPanel, ...anyOtherHelperFunction } = useScreenReaderPanel();
```

The advantages I see is this can encapsulate any related screen reader logic in the hook, plus potentially avoid any prop drilling.

packages/lib/src/core/Errors/SRPanel.tsx
```diff
@@ -0,0 +1,87 @@
...
+        showPanel: false
+    };
+
+    private readonly srPanelContainer = null;
+
+    private readonly id;
+    private readonly showPanel;
+    private readonly _enabled: boolean;
+    private readonly _moveFocus: boolean;
+
```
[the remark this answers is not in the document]
sponglord: So I can access them through getters of the same name (but without the underscore)

packages/lib/src/components/Card/components/CardInput/CardInput.test.tsx
```diff
@@ -151,14 +151,60 @@ describe('CardInput > holderName', () => {
...
+        // First visible element is the Holder name
+        const firstFormElement = loadingWrapperChildren.item(0);
+        // eslint-disable-next-line testing-library/no-node-access
+        const firstFormElementChildren = firstFormElement.children;
+
+        const label = firstFormElementChildren.item(0);
+        // eslint-disable-next-line testing-library/no-node-access
+        const labelChildren = label.children;
+
+        expect(labelChildren.item(0).textContent).toEqual('Name on card');
```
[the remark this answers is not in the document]
sponglord: How, oh react-testing-library guru?

packages/lib/src/core/Errors/SRPanel.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+        /* eslint-disable-next-line */ // linter is wrong:this is valid because we are waiting to test for non-existence
+        await waitFor(() => expect(screen.queryByTestId('message3')).toBeNull()); // non-existence
+
+        // Clear messages
+        srPanel.setMessages(null);
+
+        /* eslint-disable-next-line testing-library/prefer-find-by*/
+        await waitFor(() => expect(screen.queryByTestId('message1')).toBeNull());
+    });
+});
```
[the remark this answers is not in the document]
sponglord: the line above: `srPanel.setMessages(null);` - will clear all the errors out of the panel

packages/lib/src/core/Errors/SRMessages.tsx
```diff
@@ -0,0 +1,35 @@
+import { h, Fragment } from 'preact';
+import { useRef, useState } from 'preact/hooks';
+import { SRMessagesProps } from './types';
+
+// An interface for the members exposed by SRMessages to its parent
+export interface SRMessagesRef {
+    setMessages?: (who) => void;
```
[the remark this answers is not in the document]
sponglord: ok

packages/lib/src/core/Errors/SRPanel.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { SRPanelProps } from './types';
+import BaseElement from '../../components/BaseElement';
+import { SRMessages, SRMessagesRef } from './SRMessages';
+
+/**
+ * A panel meant to hold messages that will be read out by ScreenReaders on an aria-live="polite" basis
+ * Expects a string or string array of message to add to the panel to be read out
+ * For testing purposes can be made visible
+ */
+export class SRPanel extends BaseElement<SRPanelProps> {
```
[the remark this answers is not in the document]
sponglord: We can iterate on this

packages/lib/src/components/Card/Card.tsx
m1aw: I'm lost, how does it avoid prop drilling? 
Right now you can do the same, you can `useCoreProvider` to get the module.

packages/lib/src/core/Errors/utils.ts
```diff
@@ -45,3 +48,137 @@ export const addErrorTranslationsToObject = (originalObj, i18n) => {
...
+
+    if (currentErrorsSortedByLayout.length) {
+        /** If validating i.e. "on submit" type event - then display all errors in the SR panel */
+        if (isValidating.current) {
+            const errorMsgArr: string[] = currentErrorsSortedByLayout.map(errObj => errObj.errorMessage);
+            if (doLog) console.log('### setSRMessagesFromErrors:: #1 multiple errors:: (validating) errorMsgArr=', errorMsgArr);
+            SRPanelRef.setMessages(errorMsgArr);
+
+            // Remove 'showValidation' mode - allowing time for collation of all the fields in error whilst it is 'showValidation' mode (some errors come in a second render pass)
+            setTimeout(() => {
```
m1aw, line 153: I think we should look into stopping using Refs for state management. We will run into more and more of these issues.
@ribeiroguilherme what do you think?

sponglord: How is this state management? `SRPanelRef` is an object on which we expose the public method needed to pass the messages array

packages/lib/src/components/Card/Card.tsx
ribeiroguilherme: > I'm lost, how does it avoid prop drilling?

Same way that you end up using `useCoreProvider` to get the `i18n`. You can have similar approach to get the SRPanel and any other SR helper function.

> Right now you can do the same, you can useCoreProvider to get the module.

You definitely can, but I don't think we should keep adding more functionality to the main CoreProvider - otherwise in the future it will end up handling multiple non-related things. 

It is better to follow single responsability principle and each provider takes care of its own business. It is easier to test, to maintain, and to read. In this case, the `SRPanelProvider` could provide the SRPanel instance and potentially provide useful helper functions if needed for the UI components

packages/lib/src/core/Errors/SRPanel.test.tsx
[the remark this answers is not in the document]
ribeiroguilherme: I see. I expected that you would need to do that within the `render()` function

packages/lib/src/core/Errors/utils.ts
m1aw: Not `SRPanelRef` but `isValidating`

sponglord: ok. That didn't make it into the code snippet for your comment...
But the problem is surely not with the `isValidating` flag so much as it's a shame we need to set a timeout whilst the render passes, that generate all the errors, settle.
I would also prefer it if there was a better mechanism.

packages/lib/src/components/Card/Card.tsx
m1aw: Ah yeah, that makes sense. I think it would simplify indeed. I think I misunderstood.

packages/lib/src/core/Errors/utils.ts
m1aw: Oh, yeah right I "selected" the `setTimeout` line. 
The objective would be to `useState`.
This could be something like:
```
[1. Submit setValidating(true)] -> [2. useEffect(()=>,[isValidate])] -> [3. setSRMessage() => .... setValidating(false)]
1. We set the state on submit
2. The setSRMessage is triggered on by listening on the isValidate
3. setSRMessage sets the isValidate state again
```

packages/lib/src/components/Card/Card.tsx
sponglord: Done. Great suggestion. Don't really having many props to drill but using a Provider/Context/useContext hook encapsulates the SRPanel functionality beautifully.

packages/lib/src/core/Errors/utils.ts
ribeiroguilherme: Food for thought: There is a new way of handling state management in the React ecosystem, they call it [signals](https://preactjs.com/guide/v10/signals/). 

I haven't looked into it myself yet, but I heard it is promising. Perhaps we could take a look into this another time and see if it can help with the problem of using `useRef` to store values.

sponglord: Yes, lets. 
I've tried a useState/useEffect based solution to avoid the setTimeout, but when and where we should set the validating state and how we should react to it changing is complicated, so for now I'm sticking with the original useRef flag

Issue #1801, opened 2022-10-19
Allow asynchronous check before ApplePay sheet is shown


phuhl:
**Is your feature request related to a problem? Please describe.**
When any kind of async check needs to be performed (e.g. we need to do an API call to check for conflicts) when the user clicks on the Apple Pay button, with the current implementation the Apple Pay sheet won't open due the "Must create a new ApplePaySession from a user gesture handler" limitation.

I thought for a moment, that I could use the "onClick" method, but it is awaited **before** the ApplePaySession is created. So this is not possible for this use case.

**Describe the solution you'd like**
An optional async callback between `new ApplePayService` and `session.begin();` in `src/components/ApplePay/ApplePay.tsx` would resolve my issue. E.g. a function `confirmSession` that can be passed in the `create` method:

```diff
    private startSession(onPaymentAuthorized) {
        const { version, onValidateMerchant, onCancel, onPaymentMethodSelected, onShippingMethodSelected, onShippingContactSelected } = this.props;

        return new Promise((resolve, reject) => this.props.onClick(resolve, reject)).then(() => {
            const paymentRequest = preparePaymentRequest({
                companyName: this.props.configuration.merchantName,
                ...this.props
            });

            const session = new ApplePayService(paymentRequest, { /* ... */ });
+           this.props.confirmSession && await this.props.confirmSession();
            session.begin();
        });
    }
```

**Describe alternatives you've considered**
I think, due to the limitations of the Apple Pay API, my use case can only be done this way.

**Additional context**
The way I propose the feature, no existing APIs are broken and I know from experience with another PSP that the approach I suggest is viable (Apple Pay does allow an await before the `session.begin` call). That also matches Apples documentation, which describes the limitation only for the creation of the ApplePaySession object (https://developer.apple.com/documentation/apple_pay_on_the_web/apple_pay_js_api/creating_an_apple_pay_session) but not for the begin method (https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778001-begin).

![image](https://user-images.githubusercontent.com/16835755/196741085-d8781b8a-bc4e-4691-8cde-946369d01755.png)

phuhl:
A possible alternative would be to change how onClick behaves:

```ts
    private startSession(onPaymentAuthorized) {
        const { version, onValidateMerchant, onCancel, onPaymentMethodSelected, onShippingMethodSelected, onShippingContactSelected } = this.props;

        const paymentRequest = preparePaymentRequest({
[...]

PR #2050, opened 2023-03-17
ApplePay - Allow asynchronous check before ApplePay sheet is shown 

ribeiroguilherme:
## Summary
Enables Merchants to perform asynchronous checks using the `onClick` callback (Ex: to check if there is product in the stock) before displaying the ApplePay sheet

Highlights:
- Moved the ApplePay session creation to be before the `onClick` callback. ApplePay session must be created from an user gesture event, therefore it should happen before the async checks.
- If `onValidateMerchant` function fails for any reason, now the `onError` callback is triggered

## Tested scenarios
- Tested on web playground performing an async task on onClick event (5 seconds async task), then resolved Promise and then performed the payment 
- Verified `onError` callback is triggered when onValidateMerchant throws exception

**Fixed issue**:  #1801


packages/lib/src/components/ApplePay/ApplePayService.ts
```diff
@@ -86,13 +91,15 @@ class ApplePayService {
...
      * The onpaymentmethodselected function must resolve before the 30 second timeout
+     *
      * @param event - The event parameter contains the payment (ApplePayPayment) attribute.
      * @param onPaymentMethodSelected - A promise that will complete the payment when resolved. Use this promise to process the payment.
      * @see {@link https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778013-onpaymentmethodselected}
      */
     onpaymentmethodselected(event: ApplePayJS.ApplePayPaymentMethodSelectedEvent, onPaymentMethodSelected) {
         return new Promise((resolve, reject) => onPaymentMethodSelected(resolve, reject, event))
             .then((paymentMethodUpdate: ApplePayJS.ApplePayPaymentMethodUpdate) => {
+                console.log('onpaymentmethodselected', paymentMethodUpdate);
```
sponglord, line 102: Can this be removed?

packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -58,33 +57,41 @@ class ApplePayElement extends UIElement<ApplePayElementProps> {
...
-            });
+        const { version, onValidateMerchant, onPaymentMethodSelected, onShippingMethodSelected, onShippingContactSelected } = this.props;
 
-            session.begin();
+        const paymentRequest = preparePaymentRequest({
+            companyName: this.props.configuration.merchantName,
+            ...this.props
         });
+
+        const session = new ApplePayService(paymentRequest, {
```
sponglord, line 67: The conversation in issue #1801 talks about this being a breaking change - is it?

sponglord (approved):

packages/lib/src/components/ApplePay/ApplePay.tsx
ribeiroguilherme: I thought that it could be a breaking change by that time, but after performing the code change and some tests, it seems like that isn't the case. 

The callback is the same, and it uses the same parameters. The only difference is that we are creating the AppleSession before calling onClick, which doesn't affect the flow

PR #2055, opened 2023-03-22
Enable messages component for configuring pay later in Paypal

gsiag:
## Summary
 Currently Adyen web lacks the functionality to configure/customise pay-later messages if somebody is trying to integrate paypal's [pay in 4](https://developer.paypal.com/docs/checkout/pay-later/us/)
With the proposed changes the library will expose the Messages component from the paypal's sdk (refer url above)

Highlights:
- Added new configuration property for Paypal, named `enableMessages`  (default value: false) , which based on its value the SDK enables the [PayPal Pay Later Messages](https://developer.paypal.com/docs/checkout/pay-later/us/integrate/)

## Tested scenarios

-NA-

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/PayPal/defaultProps.ts
+++ b/packages/lib/src/components/PayPal/defaultProps.ts
@@ -44,6 +44,8 @@ const defaultProps: PayPalElementProps = {
     blockPayPalCreditButton: false,
 
     blockPayPalPayLaterButton: false,
+    
+    enableMessages: false,
 
     configuration: {
         /**
--- a/packages/lib/src/components/PayPal/types.ts
+++ b/packages/lib/src/components/PayPal/types.ts
@@ -81,6 +81,12 @@ interface PayPalCommonProps {
     blockPayPalPayLaterButton?: boolean;
 
     /**
+     * Set to true to force the UI to load Paypal Messages Component
+     * @defaultValue false
+     */
+    enableMessages?: boolean;
+
+    /**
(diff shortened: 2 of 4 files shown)


gsiag:
Hi reviewers , 
Hopefully if somebody can review this code and get it released that would be great.
I can see that you are doing weekly releases and would like this to be part of the next release if possible.
The change is pretty trivial and self explanatory . Please feel free to ask me any questions around the change. 
Cheers

gsiag:
Yes , that should work. 
I cannot share the code since I integrated and tested it with my custom code but in principal that should work.

gsiag:
`<div
        data-pp-message
        data-pp-placement="inline"
        data-pp-style-layout="text"
        data-pp-buyercountry="AU"
        data-pp-style-logo-type="alternative"
        data-pp-currency=AUD
        data-pp-amount=100
      ></div>`
      My code looked similar to this

gsiag:
@ribeiroguilherme : Any update ?

ribeiroguilherme:
I managed to test and it worked.

It would be better if this is a opt-in feature instead of enabling it for everyone, as it isn't available in all countries. Besides that,  adding `messages` also increases the bundle size of the PayPal SDK, which can affect the amount of time that the button takes to load.  

Can you create a PayPal component configuration to enable this feature? The default value would be `false`, and then we can add it to the script depending on the the value

Ex:
```js
checkout.create('paypal', {
    enableMessages: true,
    ...
}) ...
```

gsiag:
Done . Please take a look again. 
Thanks

ribeiroguilherme (approved):

PR #2059, opened 2023-03-24
Feature/small a11y fixes

sponglord:
## Summary
Some small fixes/improvements for a11y:
- the `gender` field (in the `PersonalDetails` component) now shows  a _visual_ error message
- `Address` _input_ fields have a `required/aria-required` attribute, where appropriate
- labels for securedField elements no longer have a `for` attribute pointing to the securedField input (since it exists in another DOM)
- `aria-describedby` elements (the visual error fields) are always present in the DOM

## Tested scenarios
Manually confirmed above scenarios
All unit tests pass

### DIFF
--- a/packages/lib/src/language/locales/en-US.json
+++ b/packages/lib/src/language/locales/en-US.json
@@ -68,6 +68,7 @@
     "dateOfBirth": "Date of birth",
     "shopperEmail": "Email address",
     "gender": "Gender",
+    "gender.notselected": "Select a gender",
     "male": "Male",
     "female": "Female",
     "billingAddress": "Billing address",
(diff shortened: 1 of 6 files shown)


packages/lib/src/language/locales/en-US.json
```diff
@@ -68,6 +68,7 @@
     "dateOfBirth": "Date of birth",
     "shopperEmail": "Email address",
     "gender": "Gender",
+    "gender.notselected": "Select a gender",
```
ribeiroguilherme, line 71: Did you request translations for it?

ribeiroguilherme (approved):

packages/lib/src/language/locales/en-US.json
sponglord: They still need to be done. But the english translation will work for now re. the a11y re-test

PR #2060, opened 2023-03-24
fixes select not working on keyboard

m1aw:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/internal/FormFields/Select/Select.tsx
+++ b/packages/lib/src/components/internal/FormFields/Select/Select.tsx
@@ -78,10 +78,7 @@ function Select({
     };
 
     const extractItemFromEvent = (e: Event): SelectItem => {
-        // If the target is not one of the list items, select the first list item
-        const target: HTMLInputElement = selectListRef.current.contains(e.currentTarget) ? e.currentTarget : selectListRef.current.firstElementChild;
-
-        const value = target.getAttribute('data-value') as string;
+        const value = (e.currentTarget as HTMLInputElement).getAttribute('data-value') as string;
         return filteredItems.find(listItem => listItem.id == value);
     };
 


sponglord:
For the sake of our Docs team can you provide a summary/description ;-)

sponglord (approved):

packages/e2e-playwright/models/issuer-list.ts
```diff
@@ -28,6 +28,23 @@ class IssuerList {
     async selectHighlightedIssuer(issuerName: string) {
         await this.highlightedIssuerButtonGroup.getByRole('button', { name: issuerName }).click();
     }
+
+    async typeToFilterTerm(filter: string) {
+        await this.selectorCombobox.focus();
```
ribeiroguilherme, line 33: I wonder if there is any difference between doing `.focus()` against `.click()` here. I believe `click` emulates the 'real user behavior' more than focusing on it programatically. Any thoughts on this?

packages/e2e-playwright/models/issuer-list.ts
```diff
@@ -28,6 +28,23 @@ class IssuerList {
...
+        await this.selectorCombobox.type(filter);
+    }
+
+    async pressKeyboardToNextItem() {
+        await this.selectorCombobox.press('ArrowDown');
```
ribeiroguilherme, line 38: Couldn't we just simulate a global keyboard event here, without explicitly using the 'selectorCombobox' item? (Ex: `await keyboard.press('ArrowDown')`);

My point is that, in a real scenario, the keyboard is kinda 'global' and it just fires events, and if the Component is properly focused, it should receive those keyboard events and handle it. 

In this case, we are explicitly selecting the combobox. I expect that this function might work even if the Combobox isn't focused/active by the 'user' (I am assuming, I didn't have time to test it though!)

packages/e2e-playwright/models/issuer-list.ts
m1aw: My objective using focus, instead of click was to emulate keyboard/screen reader movement. We already test flows with `.click()` does aims to cover a different use case.

ribeiroguilherme (approved):

PR #2090, opened 2023-04-03
Adds missing alternative text on some images 

m1aw:
## Summary

This PR adds the missing alternative text in some images. Mainly the available brands in the card component, the QR code component, and the CVC hint.

Note theres a few limitations with this PR. 
Not all available brands will have alt image, and they are not localised, this is something we should solve by getting this information from the backend.
If the QR code component is loaded from an action, outside of Dropin the alt is not going to be available since the `payment/` call will not retrieve the `displayName`.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
+++ b/packages/lib/src/components/internal/QRLoader/QRLoader.tsx
@@ -12,6 +12,7 @@ import AdyenCheckoutError from '../../../core/Errors/AdyenCheckoutError';
 import useCoreContext from '../../../core/Context/useCoreContext';
 import ContentSeparator from '../ContentSeparator';
 import { StatusObject } from '../Await/types';
+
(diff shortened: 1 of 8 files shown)


packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -136,7 +137,7 @@ class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
             });
     };
 
-    render({ amount, url, brandLogo, countdownTime, type, onActionHandled }: QRLoaderProps, { expired, completed, loading }) {
+    render({ amount, url, brandLogo, brandName, countdownTime, type, onActionHandled }: QRLoaderProps, { expired, completed, loading }) {
```
sponglord, line 140: Where is this `brandName` coming from? `QRLoaderContainer`?

sponglord (approved):

PR #2099, opened 2023-04-05
Feat(online-banking-PL): add terms and conditions

longyulongyu:
## Summary
In order to improve Online banking PL shopper journey we can skip approving P24 T&Cs on their hosted page. That part will be handled by specifying one parameter (by default for all merchant) on the backend. However P24 requires that we display a disclaimer sentence with a link towards their T&Cs in case shopper wants to review them.

- Add "By continuing you agree with the [regulations](https://www.przelewy24.pl/regulamin) and [information obligation](https://www.przelewy24.pl/obowiazek-informacyjny-rodo-platnicy) of Przelewy24" terms and condition to the online banking PL
- Added all translations
- Refactor to expand the `DisclaimerMessage` component
- Move the T&C above the _Pay_ button
- Conditional render the Field error message. Because when there is no error message, it still renders empty element with top margin, which creates unwanted extra space between the input element and the T&C

## Tested scenarios
- Open the playground: http://localhost:3020/?countryCode=PL
- Expand the Online Banking PL, you should be able to see 
Screenshot 2023-04-13 at 15 22 24

- Change the URL to http://localhost:3020/?countryCode=PL&shopperLocale=PL, you should be able to see:
Screenshot 2023-04-13 at 15 22 56


packages/lib/src/components/OnlineBankingCZ/index.ts
```diff
@@ -7,11 +7,16 @@ const ICON = 'bankTransfer_IBAN';
 class OnlineBankingCZElement extends IssuerListContainer {
     public static type = 'onlineBanking_CZ';
 
+    private static termsAndConditions = {
```
ribeiroguilherme, line 10: What is the benefit of using a static variable for this? 

We currently use static properties to define only the `type` and `defaultProps` , which kinda makes sense due to how we implement to get the payment method type and Component defaultProps in the library. But for this case, I don't understand the benefit

longyulongyu: I think it would make sense to use the static here because `termsAndConditions` doesn't change per `OnlineBankingCZElement` instance, it kinda belongs to the class itself.

sponglord:
Is this an agreed design? It differs from what has been done in the Credit Card and Donation components in that these terms and conditions appear _below_ the Pay button.

Also it seems we are now using two different methods to add Disclaimers to a component. 
Was it not possible to expand the `DisclaimerMessage` component (to handle multiple urls) _or_ should we be using this string interpolation method _instead of_ the DisclaimerMessage component?

longyulongyu:
> Is this an agreed design? It differs from what has been done in the Credit Card and Donation components in that the terms and conditions here, appear _below_ the Pay button.
> 
> Also it seems we are now using two different methods to add Disclaimers to a component. Was it not possible to expand the `DisclaimerMessage` component (to handle multiple urls) _**or**_ should we be using this string interpolation method _instead of_ the DisclaimerMessage component?

Checked internally, it should be the same design for the Giving component and online Banking PL, SK and CZ - terms and conditions should be above the Pay button. My first impression is to expand the `DisclaimerMessage` component (to handle multiple urls), but will investigate further because of the translation stuff.

packages/lib/src/components/internal/FormFields/Field/Field.tsx
```diff
@@ -126,13 +126,15 @@ const Field: FunctionalComponent<FieldProps> = props => {
                         </span>
                     )}
                 </div>
+                {errorMessage && (
```
sponglord, line 129: With the new a11y guidance we want this element to _always_ be present in the DOM, regardless of whether we have an `errorMessage` or not.
This change would be a regression to how it used to be - the element only showing _if_ we have an `errorMessage` prop

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.scss
```diff
@@ -1,12 +1,13 @@
 @import "../../../style/index";
 
 .adyen-checkout-disclaimer__label {
-    position: relative;
+    width: 100%;
+    text-align: center;
```
sponglord, line 5: This cause the T&Cs to align centrally. 
For the Card component they should be left aligned. 
I don't know about for OnlineBanking_PL or Donation component?

packages/lib/src/components/internal/FormFields/Field/Field.tsx
longyulongyu: Ok, thanks, I'll revert it back.

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.scss
longyulongyu: For OnlineBanking_PL or Donation component, the design is to align it centrally. I'll remove the `text-align: center;` from this component, and let the parent decides how to position it.

sponglord:
Overall I like this approach _a lot_.

I'm thinking that for a (future) iteration `DisclaimerMessage` could just work one way - with the `InternalDisclaimerMsgObject`, and drop the whole ` if ('disclaimer' in props){...` if-clause

This would mean the Card Component (i.e `CardFieldsWrapper` & `StoredCardFieldsWrapper`) and the `DonationComponent` could keep the external (merchant) facing `disclaimerMessage` config, but internally they would convert to use the `message` and `urls` props i.e.
```
{disclaimerMessage && (
    <DisclaimerMessage
        message={disclaimerMessage.message.replace('%{linkText}', `%#${disclaimerMessage.linkText}%#`)}
        urls={[disclaimerMessage.link]}
    />
)}
```
instead of the current
```
{disclaimerMessage && <DisclaimerMessage disclaimer={disclaimerMessage} />}
```

longyulongyu:
> Overall I like this approach _a lot_.
> 
> I'm thinking that for a (future) iteration `DisclaimerMessage` could just work one way - with the `InternalDisclaimerMsgObject`.
> 
> This would mean the Card Component (i.e `CardFieldsWrapper` & `StoredCardFieldsWrapper`) and the `DonationComponent` could keep the external (merchant) facing `disclaimerMessage` config, but internally they would convert it, to use the `message` and `urls` props i.e.
> 
> ```
> {disclaimerMessage && (
>     <DisclaimerMessage
>         message={disclaimerMessage.message.replace('%{linkText}', `%#${disclaimerMessage.linkText}%#`)}
>         urls={[disclaimerMessage.link]}
>     />
> )}
> ```
> 
> instead of the current
> 
> ```
> {disclaimerMessage && <DisclaimerMessage disclaimer={disclaimerMessage} />}
> ```
> 
> Then `DisclaimerMessage` could drop that whole `if ('disclaimer' in props){...}` if-clause
> 
> The only things missing though would be the current way that `DisclaimerMessage` checks that the urls it's been passed are actually valid (`isValidHttpUrl(disclaimer.link)`) _and_ that the message & link texts are actually strings (and not JS injections) - which are good features to have for security purposes. Perhaps the components that use the `DisclaimerMessage` could validate these before they markup the `DisclaimerMessage` comp...?

I agree, in the future iteration, we can use the same interface. 
About validating urls, we don't have the risk at this moment because onlineBanking PL, CZ and SK components hardcoded urls. In the future if there are more components need to render the `DisclaimerMessage` based on merchant's configuration, I agree we need to validate them. I don't have a strong preference about where to validate the urls though..

sponglord (approved):

PR #2105, opened 2023-04-07
Payment method - CashAppPay

ribeiroguilherme:
## Summary
Adding support for CashAppPay payment method.

Highlights:
- It works similar to GooglePay, as we need to load 3rd party SDK and then display its Pay button
- It supports tokenization:
  - In case the merchant wants to display the 'Save payment method' checkbox, the property `enableStoreDetails` should be set
  - In case the merchant wants to tokenize without asking for consent, the property `storePaymentMethod: true` should be set

## Tested scenarios
- Unit tests
- Tested using CashApp sandbox iOS app


sponglord:
Even after I've dismissed the modal it continues to try and load the `defaultlogo.jpg` every x number of seconds

sponglord (approved):

PR #2108, opened 2023-04-11
E2E - Fixing issuer list flaky tests

ribeiroguilherme:
## Summary
Attempt to fix issuer list flaky tests

Current issue: 
When playwright clicks in the select box, it moves to the next task which is to type in the filterable select field, although the field isn't ready for handling the keyboard events (sometimes it is, but sometimes it isn't). Therefore, in some cases, the filter mechanism wasn't working as expected and was taking wrong values instead of the provided ones.

The fix:
- Added a small delay when the select field is clicked, so the options can be fully rendered
- Added a small delay after typing in the field, so the UI can be properly updated with the available options


longyulongyu (approved):
Looks good to me 👍.

PR #2109, opened 2023-04-12
Removed isCollatingErrors prop since it is never used

sponglord:
## Summary
The `isCollatingErrors` prop was part of the `ErrorPanel` functionality.
Now we have moved to using the `SRPanel` it is a redundant property.
(The only reason it wasn't removed in the SRPanel PR was in an attempt to keep down the number of changed files for that PR)

Have also fixed some e2e tests to reflect the new markup that has come with the a11y changes i.e. `aria-describedby` elements are _always_ in the DOM, and `label` elements for securedFields don't have a `for` attribute

## Tested scenarios
All unit tests pass
All e2e tests pass


longyulongyu (commented):
I just did a global search on the project `src` folder, I noticed the `src/components/internal/SecuredFields/lib/securedField/SecuredField.test.ts` is setting the `isCollatingErrors`. For the sake of completeness, maybe remove it?

longyulongyu (approved):

Issue #2078, opened 2023-03-31
Card (BrowserInfo) Typescript definition does not match the actual data types


pvhespen:
**Describe the bug**
The `BrowserInfo` Typescript definition does not match the actual data returned by the `onChange` event for the Card implementation. 

**To Reproduce**
1. Create an AdyenCheckout:
```
  const configuration = {
      environment: '',
      clientKey: '',
      locale: '',
   };

  const checkout AdyenCheckout(configuration)
      .then(adyenCheckout => {
        const cardConfiguration: PaymentMethodOptions<'card'> = {
            hasHolderName: true,
            holderNameRequired: true,
            brands: ['mc', 'visa'],
            onChange: state => {
              console.log(state);
           },
         };

  adyenCheckout.create('card', cardConfiguration).mount('#card-container');
````
2. Use the component:
Fill in some card information to trigger the onChange event. 

**Expected behavior**
I would expect the `onChange` event to pass down a state object containing the `BrowserInfo`:
```
/**
 * {@link https://docs.adyen.com/api-explorer/#/PaymentSetupAndVerificationService/v52/post/payments__reqParam_browserInfo API Explorer /payments browserInfo}
 */
export interface BrowserInfo {
    acceptHeader: string;
    colorDepth: string;
    language: string;
    javaEnabled: boolean;
    screenHeight: string;
    screenWidth: string;
    userAgent: string;
    timeZoneOffset: number;
}
```
But the actual returned data contains:
```
{
    acceptHeader: "*/*",
​​​    colorDepth: 24,
​​​    javaEnabled: false,
​​​    language: "en-US"
​​​    screenHeight: 1440,
​​​    screenWidth: 2560,
​​    timeZoneOffset: -120,
​​    userAgent: "Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/111.0"
}
```
**Screenshots**
![Selection_007](https://user-images.githubusercontent.com/29913281/229118383-bb2938a9-c432-42b2-b43a-d8da1703fe70.png)

Running `@adyen/adyen-web` `5.38.0`

PR #2131, opened 2023-05-03
Update typings for BrowserInfo

ossiggy:
- Corrected typings for BrowserInfo to reflect the actual data that is being sent (https://docs.adyen.com/api-explorer/Checkout/70/post/payments#request-browserInfo)
- Adds optional javaScriptEnabled field

## Tested scenarios

 
Logged `onChange` event and the data types of each field in `browserInfo`.  Confirmed that these were incorrect in the type definition of BrowserInfo

**Fixed issue**:  #2078

### DIFF
--- a/packages/lib/src/types/index.ts
+++ b/packages/lib/src/types/index.ts
@@ -258,11 +258,12 @@ export interface OrderStatus {
  */
 export interface BrowserInfo {
     acceptHeader: string;
-    colorDepth: string;
+    colorDepth: number;
     language: string;
     javaEnabled: boolean;
-    screenHeight: string;
-    screenWidth: string;
+    javaScriptEnabled?: boolean;
+    screenHeight: number;
+    screenWidth: number;
     userAgent: string;
     timeZoneOffset: number;
 }


packages/lib/src/types/index.ts
```diff
@@ -258,11 +258,12 @@ export interface OrderStatus {
  */
 export interface BrowserInfo {
     acceptHeader: string;
-    colorDepth: string;
+    colorDepth: number;
     language: string;
     javaEnabled: boolean;
-    screenHeight: string;
-    screenWidth: string;
+    javaScriptEnabled?: boolean;
```
sponglord, line 264: I don't know if there's any point adding this `javaScriptEnabled` typing without going deeper into the code to _actually_ add this property to the browser info that we collect i.e. in `browserInfo.ts`.
_And_, if we do add it then I think you'd need to checkout with our API team whether they're actually going to use it.
Although it's mandated for 3DS2 we haven't been sending it in up til now - so I don't quite know how the backend handles it - whether they generate their own `javaScriptEnabled` property?
I guess it's safe, from their side, to assume that if the frontend is even sending browserInfo, which is collected via javaScript, then javaScript must be enabled

PR #2138, opened 2023-05-05
Click to Pay - Standalone component

ribeiroguilherme:
## Summary

Adding support for Click to Pay to be used as standalone Component. 

The standalone Component must be displayed only if the Shopper is recognized. In order to achieve that, the Component provides the async `isAvailable()` method which will report if the shopper was found, and if so, the Component can be mounted by the merchant.

The component can be created by doing:

```js
const clickToPay = checkout.create('clicktopay', {
    shopperEmail: 'shopper@example.com',
    onReady() { console.log('Component is ready to interact with the shopper') }
});

clickToPay.isAvailable().then(() => {
     // Shopper is recognized (by cookies or by his email)
    clickToPay.mount('.clicktopay-container');
}).catch(() => {
    // Click to Pay isn't available (Ex: shopper wasn't recognized)
    // Then display something else
});
```

Highlights:
- Click to Pay does not show up in the Drop-in as payment method. It is supposed to be used as part of the Card component, or as standalone Component. Therefore it won't be displayed in the drop-in payment method list
- `onReady` event reports when the Component is ready to be used. There are three moments which that can happen: When the OTP screen is displayed; When the cookies are recognized and the Card lists screen is displayed; or when something goes wrong and the Login screen is displayed;
- `onReady` can also be used by the Card component implementation. Example:

```js
const clickToPay = checkout.create('card', {
    clickToPayConfiguration: {
        shopperEmail: 'shopper@example.com',
        onReady() { console.log('Component is ready to interact with the shopper') }
    } ,
    ...
});
```

## Tests
- Added unit tests for the new element
- Manual tests for Standalone component and Card component version


packages/lib/src/components/ClickToPay/ClickToPay.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    private handleClickToPaySubmit = (payload: ClickToPayCheckoutPayload) => {
+        this.setState({ data: { ...payload }, valid: {}, errors: {}, isValid: true });
+        this.submit();
+    };
+
+    render() {
+        return (
+            <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources}>
+                <ClickToPayProvider
```
sponglord, line 106: Is there any value in using the `ClickToPayWrapper` here, like we do in the `Card` component? 
Like, consistency of implementation?

packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLoginInput.tsx
```diff
@@ -57,6 +57,11 @@ const CtPLoginInput = (props: CtPLoginInputProps): h.JSX.Element => {
         [props.onPressEnter]
     );
 
+    const handleOnKeyPress = useCallback((event: h.JSX.TargetedKeyboardEvent<HTMLInputElement>) => {
```
sponglord, line 60: Do you need _separate_ handlers for `onKeyPress` and onKeyUp` to handle the user pressing the `Enter` key?
Couldn't it all be handled in the `handleOnKeyPress` function

sponglord (approved):

packages/lib/src/components/internal/ClickToPay/context/ClickToPayProvider.tsx
```diff
@@ -49,6 +52,14 @@ const ClickToPayProvider = ({
         ctpService?.subscribeOnStateChange(status => setCtpState(status));
```
longyulongyu, line 52: Do we need any cleanup, to `logout` the shopper before the `ClickToPayProvider` 'destroyed'? Or it's not necessary?

packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLoginInput.tsx
ribeiroguilherme: The `onKeyPress` seems to be the only event that can properly prevent the form of being submitted when pressing Enter. I tried the `onKeyUp` and it didn't work.

Perhaps we could try it out to use only `onKeyPress` instead of `onKeyUp`, but that is a big change and would impact all the fields. If we go for it, it is better try that in a different PR

packages/lib/src/components/internal/ClickToPay/context/ClickToPayProvider.tsx
ribeiroguilherme: That is not necessary. We must logout the shopper only if he clicks in the button to be logged out :)

packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLoginInput.tsx
sponglord: ok

packages/lib/src/components/ClickToPay/ClickToPay.tsx
sponglord: ok.

Issue #2135, opened 2023-05-04
v5.41.0: Cannot read properties of undefined (reading 'getImage')


pguilbert:
**Describe the bug**

We are using the [Adyen custom card integration](https://docs.adyen.com/payment-methods/cards/custom-card-integration).

After updating to the new 5.41.0 version of the package we no longer received `onChange` event from the secured fields. 

This makes the form unusable, we had to revert to the previous version.

**To Reproduce**

Our integration is similar to the [documentation](https://docs.adyen.com/payment-methods/cards/custom-card-integration).

```
  this.checkoutInstance = await AdyenCheckout({
    locale: this.locale,
    clientKey: this.adyenPublicKey,
    environment: this.adyenEnvironment,
  })

const component = this.checkoutInstance.create('securedfields', {
      type: 'card',
      styles: ADYEN_COMPONENT_STYLES,

      onConfigSuccess: () => {
        resolve()
      },

      onError: (payload) => {
        reject(Object.assign(new Error(payload?.error), { payload }))
      },

      onBrand: (payload) => {
        this.brandPayload = cloneDeep(payload)
      },
      onChange: (payload) => {
        console.log('onChange', payload)
        this.changePayload = cloneDeep(payload)
      },
      onFocus: (payload) => {
        this.focusPayload = cloneDeep(payload)
      },
    })

    component.mount(container)
  })
```

The `console.log('onChange', payload)` is called once when the form is initialized, but never after. I can see the following console log as soon as I type in the card number field:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'getImage')
    at Qr (index.js?57f2:1:1)
    at $s.eval (index.js?57f2:1:1)
    at eval (preact.module.js?da65:1:1)
    at Array.some (<anonymous>)
    at eval (preact.module.js?da65:1:1)
    at Array.some (<anonymous>)
    at z (preact.module.js?da65:1:1)
    at eval (preact.module.js?da65:1:1)
    at Array.some (<anonymous>)
    at g (preact.module.js?da65:1:1)
```

**Screenshots**
<img width="607" alt="image" src="https://user-images.githubusercontent.com/3878835/236237315-81937eb6-b63c-4691-a847-5b5ad491f5aa.png">.

**Desktop:**
 - OS: MAC
 - Browser: chrome
 - Version : 111

Thank you in advance!

m1aw:
[...]

PR #2139, opened 2023-05-05
Added conditional to getCardImageUrl

ossiggy:
## Summary

- Component breaks when `onBrand` event is fired in Custom Card component due to `this.props.resources` being `undefined` when passed to `getCardImageUrl()` in [SecuredFieldsProviderHandlers.ts](https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/components/internal/SecuredFields/SFP/SecuredFieldsProviderHandlers.ts#L176). This conditional checks that `resources` is truthy before trying to execute the `getImage()` method

## Tested scenarios
- confirmed this was an issue by running playground, any change to the card fields caused an error
- added change and retested, successfully entering full card details and creating an authorization in my test account
- tested all card brands on Custom Card Component, Drop-in, and Cards after change to ensure this did not affect working code

**Fixed issue**:  #2135

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/utils.ts
+++ b/packages/lib/src/components/internal/SecuredFields/utils.ts
@@ -55,7 +55,7 @@ export const getCardImageUrl = (brand, resources: Resources) => {
         extension: 'svg'
     };
 
-    return resources.getImage(imageOptions)(type);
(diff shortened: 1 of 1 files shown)


m1aw:
Thanks for the PR @ossiggy but we are going with this solution #2141.

Issue #2100, opened 2023-04-05
Typescript definition for create always returns InstanceType<PaymentMethods['redirect']>


brendanoffer:
**Describe the bug**

Typescript definition for `Core`'s create methods are incorrect. https://github.com/Adyen/adyen-web/pull/1566 introduced a fallback however this typescript interface will be used for all calls using a string paymentMethod paramter.

**To Reproduce**
1. Use typescript
2. Follow https://docs.adyen.com/payment-methods/apple-pay/web-component documentation for adding a Apple Pay web component
3. When `const applePayComponent = checkout.create('applepay', applePayConfiguration);` is called it returns a type `InstanceType<PaymentMethods['redirect']>`
4. See error `Type error: Property 'isAvailable' does not exist on type 'RedirectElement'.` 

**Expected behavior**

Apple Pay Component or `InstanceType<PaymentMethods[T]>` type is returned from `.create`

obouchari:
I have had the same issue but it turned out if somehow the configuration you pass in doesn't match what "applepay" expect it will return a Redirect type.

dprcoles:
Hey @brendanoffer, I recently ran into the same issue. It looks like `totalPriceLabel` is a required property on the [`ApplePayElementProps`](https://github.com/Adyen/adyen-web/blob/cc7aa40e89d46ce9b121c63576190a4b62faae4b/packages/lib/src/components/ApplePay/types.ts#L51) type, which is the type that needs to be passed in to the `options` parameter of the `checkout.create()` method for it to return the expected `ApplePayElement` type.

If you change your call to `const applePayComponent = checkout.create<'applepay'>('applepay', applePayConfiguration);` you should see a typescript error flagging this.

ribeiroguilherme:
Hi @brendanoffer ,

@dprcoles made a fix for this issue, and it should be available on 5.43.0 onwards. Thanks for reporting it. Cheers!

PR #2140, opened 2023-05-05
make `totalPriceLabel` optional in `ApplePayElementProps` type

dprcoles:
## Summary

Raised a change to make the `totalPriceLabel` property optional within the `ApplePayElementProps` type.

According to the example in the docs [here](https://docs.adyen.com/payment-methods/apple-pay/web-component?tab=_code_payments_code__2#use-adyens-apple-pay-certificate), when you set up an `applepay` web component you only need to pass the `amount` and `currencyCode` properties. However, within the `ApplePayElementProps` type the `totalPriceLabel` property is required, which causes a typing issue when initialising the component via the `checkout.create()` method as it doesn't correctly identify the `applePayConfiguration` object as an `ApplePayElementProps` type.

The result of this is a typescript error when trying to use the `.isAvailable()` method on the component as defined in Step 3 on the docs [here](https://docs.adyen.com/payment-methods/apple-pay/web-component?tab=_code_payments_code__2#step-3-mount-the-component), as `applePayComponent` is of the `RedirectElement` type and not the `ApplePayElement` type that `isAvailable` exists on.

## Tested scenarios

I can see [here](https://github.com/Adyen/adyen-web/blob/cc7aa40e89d46ce9b121c63576190a4b62faae4b/packages/lib/src/components/ApplePay/ApplePay.tsx#L39) that `totalPriceLabel` gets set to `props.configuration?.merchantName` as a fallback in the event of it not being set, along with test coverage on that scenario [here](https://github.com/Adyen/adyen-web/blob/cc7aa40e89d46ce9b121c63576190a4b62faae4b/packages/lib/src/components/ApplePay/ApplePay.test.ts#L22-L25) - which suggests that this value probably doesn't need to explicitly be set when creating the component?

**Fixed issue**:  
#2100

### DIFF
--- a/packages/lib/src/components/ApplePay/types.ts
+++ b/packages/lib/src/components/ApplePay/types.ts
@@ -48,7 +48,7 @@ export interface ApplePayElementProps extends UIElementProps {
      * Part of the 'ApplePayLineItem' object, which sets the label of the payment request
      * @see {@link https://developer.apple.com/documentation/apple_pay_on_the_web/applepaylineitem ApplePayLineItem docs}
      */
-    totalPriceLabel: string;
+    totalPriceLabel?: string;
 
     /**
      * @default 'final'


ribeiroguilherme (approved):

ribeiroguilherme:
Thanks for contributing @dprcoles and for the detailed investigation. The fix will be available in the next release, which should take place likely next week. Cheers!

PR #2151, opened 2023-05-11
feature: onAddressLookup

m1aw:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/internal/Address/Address.tsx
+++ b/packages/lib/src/components/internal/Address/Address.tsx
@@ -13,6 +13,7 @@ import { ADDRESS_SCHEMA, FALLBACK_VALUE } from './constants';
 import { getMaxLengthByFieldAndCountry } from '../../../utils/validator-utils';
 import useCoreContext from '../../../core/Context/useCoreContext';
 import { ComponentMethodsRef } from '../../types';
+import AddressSearch from './components/AddressSearch';
 
 export default function Address(props: AddressProps) {
     const { i18n } = useCoreContext();
@@ -29,6 +30,8 @@ export default function Address(props: AddressProps) {
     const specifications = useMemo(() => new Specifications(props.specifications), [props.specifications]);
 
     const requiredFieldsSchema = specifications.getAddressSchemaForCountryFlat(props.countryCode).filter(field => requiredFields.includes(field));
+
+    const showAddressSearch = !!props.onAddressLookup;
 
     const { data, errors, valid, isValid, handleChangeFor, triggerValidation } = useForm<AddressData>({
         schema: requiredFieldsSchema,
@@ -135,11 +138,17 @@ export default function Address(props: AddressProps) {
 
     return (
         <Fragment>
-            <Fieldset classNameModifiers={[label || 'address']} label={label}>
-                {addressSchema.map(field => (field instanceof Array ? getWrapper(field) : getComponent(field, {})))}
-            </Fieldset>
-            {/* Needed to easily test when showValidation is called */}
-            {process.env.NODE_ENV !== 'production' && props.showPayButton && props.payButton({ label: i18n.get('continue') })}
+            {showAddressSearch ? (
+                <AddressSearch onAddressLookup={props.onAddressLookup} />
+            ) : (
+                <Fragment>
+                    <Fieldset classNameModifiers={[label || 'address']} label={label}>
+                        {addressSchema.map(field => (field instanceof Array ? getWrapper(field) : getComponent(field, {})))}
+                    </Fieldset>
+                    {/* Needed to easily test when showValidation is called */}
+                    {process.env.NODE_ENV !== 'production' && props.showPayButton && props.payButton({ label: i18n.get('continue') })}
+                </Fragment>
+            )}
         </Fragment>
     );
 }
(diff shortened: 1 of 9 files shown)


sponglord:
I'm seeing that if I use addressLookup to fill an address and then use it to select a second address, the filled address doesn't change

m1aw:
Both scenarios have been tested and fixed.

packages/playground/src/pages/Wallets/Wallets.html
```diff
@@ -11,24 +11,6 @@
...
-
-                <div class="merchant-checkout__payment-method">
-                    <div class="merchant-checkout__payment-method__header">
-                        <h2>Cash App Pay</h2>
-                    </div>
-                    <div class="merchant-checkout__payment-method__details">
-                        <div class="cashapp-field"></div>
-                    </div>
-                </div>
-
```
ribeiroguilherme, line 31: They are being used there, no?

packages/playground/src/pages/Dropin/session.js
```diff
@@ -52,6 +52,7 @@ export async function initSession() {
 
     const dropin = checkout
         .create('dropin', {
+            showRemovePaymentMethodButton: 'true',
```
ribeiroguilherme, line 55: shouldn't it be boolean?

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+export default function AddressSearch({ onAddressLookup, onSelect, onManualAddress, externalErrorMessage, hideManualButton }: AddressSearchProps) {
+    const [formattedData, setFormattedData] = useState([]);
+    const [originalData, setOriginalData] = useState([]);
+
+    const [errorMessage, setErrorMessage] = useState('');
+
+    const { i18n } = useCoreContext();
+    const mapDataToSelect = data => data.map(({ id, name }) => ({ id, name }));
```
ribeiroguilherme, line 33: Is this mapping doing something? 🤔

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    const onInput = useCallback(
+        async event => {
```
ribeiroguilherme, line 36: Is this an actual Input `event` or `value`? Seems we pass the value but by looking it here looks like we pass the event itself. Adding proper types would help to understand the code

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+    actions: {
+        resolve: (value: Array<AddressLookupItem>) => void;
+        reject: (reason?: any) => void;
+    }
+) => Promise<void>;
+
+interface AddressSearchProps {
+    onAddressLookup?: OnAddressLookupType;
+    onSelect: any; //TODO
+    onManualAddress: any;
```
ribeiroguilherme, line 21: Missing types

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+        if (!event.target.value) {
+            setErrorMessage(i18n.get('address.errors.incomplete'));
+            return;
+        }
+        const value = originalData.find(item => item.id === event.target.value);
+        onSelect(value);
+        setFormattedData([]);
+    };
+
+    const debounceInputHandler = useMemo(() => debounce(onInput), []);
```
longyulongyu, line 68: I think we don't need to use `useMemo` here?
We can use: 
```javascript
const debounceInputHandler = useCallback(debounce(onInput), [onInput]);
```

packages/lib/src/components/internal/Address/Address.tsx
```diff
@@ -30,17 +31,49 @@ export default function Address(props: AddressProps) {
...
+            // Cast everything to string
+            setData(propKey, String(providedValue));
+            triggerValidation();
+        });
+        setHasSelectedAddress(true);
+    };
+
+    const onManualAddress = () => {
+        setUseManualAddress(true);
+    };
```
ribeiroguilherme, line 67: You can wrap those functions in `useCallback` . They can/will end up being recreated multiple times without need for that

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    const onChange = event => {
+        if (!event.target.value) {
+            setErrorMessage(i18n.get('address.errors.incomplete'));
+            return;
+        }
+        const value = originalData.find(item => item.id === event.target.value);
+        onSelect(value);
+        setFormattedData([]);
+    };
```
ribeiroguilherme, line 66: Out of curiosity - What is the difference between `onInput` and `onChange` ? They seem be the same thing but they are not 😅

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
m1aw: Yeah, its cleaning data into a new array.

PR #2163, opened 2023-05-15
Add a paymentMethodVariants array to the object sent to the onBinLookup callback

sponglord:
## Summary
This PR adds a `paymentMethodVariants` array to the object sent to the merchant defined, `onBinLookup`, callback.
This is the callback we use to tell the merchant what the results of the call to the `/binLookup` endpoint were.

That means the object sent to this callback now looks something like this:
```
{
    "type": "card",
    "detectedBrands": ["mc"],
    "supportedBrands": ["mc"],
    "paymentMethodVariants": ["mcdebit"],
    "brands": ["visa", "mc", "amex"],
    "issuingCountryCode": "US"
}
```

The `paymentMethodVariant` represents a more granular description of the payment method, e.g. '"mcdebit"` rather than just `"mc"`. See, https://docs.adyen.com/development-resources/paymentmethodvariant.

So now, as well as telling the merchant what brands we detect the entered card number to be, and whether it is one of the brands they support, we also say what the `paymentMethodVariant` is of the entered card number.

## Tested scenarios
A `paymentMethodVariants` array is present in the object sent to the `onBinLookup` callback in both the "supported brands detected" and "supported brands not detected" scenarios.

**Fixed issue**:  COAPI-79

### DIFF
--- a/packages/lib/src/components/Card/types.ts
+++ b/packages/lib/src/components/Card/types.ts
@@ -211,6 +211,7 @@ export interface BrandObject {
     supported: boolean;
     brandImageUrl?: string;
     panLength?: number;
+    paymentMethodVariant?: string;
 }
 
 export interface BinLookupResponseRaw {
--- a/packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
+++ b/packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
@@ -86,6 +86,8 @@ export default parent => {
                             (acc, item) => {
                                 // All brand strings end up in the detectedBrands array
                                 acc.detectedBrands.push(item.brand);
+                                // Also add the paymentMethodVariants (more granular description of the txvariant)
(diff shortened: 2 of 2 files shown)


packages/lib/src/components/Card/types.ts
```diff
@@ -211,6 +211,7 @@ export interface BrandObject {
     supported: boolean;
     brandImageUrl?: string;
     panLength?: number;
+    paymentMethodVariant?: string;
```
ribeiroguilherme, line 214: Isn't that an array of strings? Is it possible to have more than one value in that array?

ribeiroguilherme (approved):

packages/lib/src/components/Card/types.ts
sponglord: This is for the single object(s) that get returned in the /binLookup response.
We collect the `paymentMethodVariant` strings and put them into a `paymentMethodVariants` array

PR #2182, opened 2023-05-19
Feat: add a11y for the count down component

longyulongyu:
## Summary
- Feeds the count down information to the SR panel
- Refactor the `Countdown` to a functional component
- The `Countdown` component initiates the A11yManager when mounted, clean up the when unmounted, and calls the `update` every 1 sec. 
- The A11yManager decides how frequent to update the SR panel, and it also handles the translation part

## Tested scenarios
- Added tests for the `Countdown` component
- All tests passed

**Relates to ticket:** COWEB-1232 & a11y _instanceId_: 2049095407

### DIFF
--- a/packages/lib/src/components/internal/Countdown/A11yManager.ts
+++ b/packages/lib/src/components/internal/Countdown/A11yManager.ts
@@ -1,0 +1,78 @@
+// The component knows how to communicate with SR, how to build translation messages
+// Listen to the new time minutes and seconds change, knows the timer component state
+// If mins > 5, updates the SR every 3 mins
+// If 1 < mins <=5, update the SR every 1 min
+// If min <=1 && seconds>0, update SR every 30s
+// Clear SR every 5 seconds
+// Un-mounding will clear all timers
+
+export class A11yManager {
+    protected static translationKey = 'sr.wechatpay.timetopay';
+    protected static longTimeout = 1000 * 5; // 3 minutes 3 * 1000 * 60;
+    protected static midTimeout = 3000; // 1 minutes 1000 * 60;
+    protected static shortTimeout = 1000; // 30 seconds 1000 * 30;
+
+    private interval;
+    private srPanel;
+    private i18n;
+    private timeout;
+    private timeLeft;
+
+    constructor(props) {
+        // todo: make duration configurable
+        const { srPanel, i18n } = props;
+        this.srPanel = srPanel;
+        this.i18n = i18n;
+    }
+
+    public update(time) {
+        this.timeLeft = time;
+        const { minutes: minutesLeft, seconds: secondsLeft } = time;
+        if (minutesLeft > 5 && this.timeout !== A11yManager.longTimeout) {
+            this.timeout = A11yManager.longTimeout;
+            this.setInterval(this.timeout);
+        }
+        if (minutesLeft >= 1 && minutesLeft <= 5 && this.timeout !== A11yManager.midTimeout) {
+            this.timeout = A11yManager.midTimeout;
+            this.setInterval(this.timeout);
+        }
+        if (minutesLeft < 1 && secondsLeft > 0 && this.timeout !== A11yManager.shortTimeout) {
+            this.timeout = A11yManager.shortTimeout;
+            this.setInterval(this.timeout);
+        }
+    }
+
+    public tearDown() {
+        this.clearInterval();
+    }
+
+    private setInterval(timeout) {
+        this.clearInterval();
+
+        const setSrMessages = () => {
+            this.srPanel.setMessages(this.getSrMessages(this.timeLeft));
+        };
+        // to execute immediately
+        setSrMessages();
+        this.interval = setInterval(setSrMessages, timeout);
+    }
+
+    private getSrMessages({ minutes, seconds }) {
+        const translation = this.i18n.get(A11yManager.translationKey);
+        const fns = [minutes, seconds].map(time => translation => `${time} ${translation}`);
+        return [this.build(translation, fns).join('')];
+    }
+
+    private build(translation, cbFunctions) {
+        const matches = translation.split(/%#(.*?)%#/gm);
+
+        return matches.map((term, index) => {
+            const indexInFunctionArray = Math.floor(index / 2);
+            return index % 2 === 0 ? term : cbFunctions[indexInFunctionArray](term);
+        });
+    }
+
+    private clearInterval() {
+        if (this.interval) clearInterval(this.interval);
+    }
+}
--- a/packages/lib/src/components/internal/Countdown/A11yManager.ts
+++ b/packages/lib/src/components/internal/Countdown/A11yManager.ts
@@ -1,10 +1,10 @@
 import { interpolateElement } from '../../../language/utils';
 
 export class A11yManager {
-    protected translationKey = 'sr.wechatpay.timetopay';
-    protected longTimeout = 3 * 60 * 1000; // 3 minutes
-    protected midTimeout = 60 * 1000; // 1 minute
-    protected shortTimeout = 30 * 1000; // 30 seconds
+    protected TRANSLATION_KEY = 'sr.wechatpay.timetopay';
+    protected LONG_TIMEOUT = 180_000;
+    protected MID_TIMEOUT = 60_000;
+    protected SHORT_TIMEOUT = 30_000;
 
     private srPanel;
     private srInterval;
@@ -28,16 +28,16 @@ export class A11yManager {
         const secondsLeft = parseInt(seconds, 10);
         this.timeLeft = { minutes: minutesLeft, seconds: secondsLeft };
 
-        if (minutesLeft > 5 && this.timeout !== this.longTimeout) {
-            this.timeout = this.longTimeout;
+        if (minutesLeft > 5 && this.timeout !== this.LONG_TIMEOUT) {
+            this.timeout = this.LONG_TIMEOUT;
             this.setInterval(this.timeout);
         }
-        if (minutesLeft >= 1 && minutesLeft <= 5 && this.timeout !== this.midTimeout) {
-            this.timeout = this.midTimeout;
+        if (minutesLeft >= 1 && minutesLeft <= 5 && this.timeout !== this.MID_TIMEOUT) {
+            this.timeout = this.MID_TIMEOUT;
             this.setInterval(this.timeout);
         }
-        if (minutesLeft < 1 && secondsLeft > 0 && this.timeout !== this.shortTimeout) {
-            this.timeout = this.shortTimeout;
+        if (minutesLeft < 1 && secondsLeft > 0 && this.timeout !== this.SHORT_TIMEOUT) {
+            this.timeout = this.SHORT_TIMEOUT;
             this.setInterval(this.timeout);
         }
     }
@@ -61,7 +61,7 @@ export class A11yManager {
     }
 
     private getSrMessages({ minutes, seconds }) {
-        const translation = this.i18n.get(this.translationKey);
+        const translation = this.i18n.get(this.TRANSLATION_KEY);
         const getTimeTranslation = time => (time !== 0 ? translation => `${time} ${translation}` : () => '');
         const fns = [minutes, seconds].map(getTimeTranslation);
         return [interpolateElement(translation, fns).join('')];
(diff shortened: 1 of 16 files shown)


packages/lib/src/core/Errors/SRPanel.tsx
```diff
@@ -10,11 +10,12 @@ import { SRMessages, SRMessagesRef } from './SRMessages';
...
  */
 export class SRPanel extends BaseElement<SRPanelProps> {
-    public static defaultProps = {
+    public static defaultProps: SRPanelProps = {
         enabled: true,
         node: 'body',
         showPanel: false,
-        id: 'ariaLiveSRPanel'
+        id: 'ariaLiveSRPanel',
+        arialRelevant: 'all'
```
sponglord, line 18: typo!  Should be `ariaRelevant` :-)

sponglord:
The advice from LevelAccess was that the countdown message in the SRPanel should clear after a few seconds (5 or 10), but I can see that it remains there until the next update

packages/lib/src/components/internal/Countdown/A11yManager.ts
```diff
@@ -0,0 +1,73 @@
...
+    private i18n;
+    private timeout;
+    private timeLeft;
+
+    constructor(props) {
+        const { srPanel, i18n } = props;
+        this.srPanel = srPanel;
+        this.i18n = i18n;
+        // Force the srPanel to update arialRelevant
+        this.srPanel.update({ arialRelevant: 'additions text' });
```
sponglord, line 20: I wonder if updating the component (which then unmounts & remounts the components is the best way here?
Could you also pass the `ariaRelevant` value as a prop when declaring the `SRPanelProvider` and have the Provider set the value into the SRPanel?
Just wondering...

longyulongyu:
> The advice from LevelAccess was that the countdown message in the SRPanel should clear after a few seconds (5 or 10), but I can see that it remains there until the next update

I have some doubt about this 🤔 , because a new message will feed into this `aria-live` section after certain time and will be announced by the SR, and SR only announces `additions text` as we instruct it to do so, so it won't repeat the same message.
I suspect the advice is to avoid the same message being repeated, but with `additions text` it won't happen. 
Hence I think we don't need to clear out the message..

packages/lib/src/components/internal/Countdown/index.tsx
```diff
@@ -8,75 +12,55 @@ interface CountdownProps {
...
+        const tick = () => {
+            const { minutes, seconds, percentage, completed } = getTimeDifference(startTime, endTime);
+            if (completed) {
+                onCompleted();
+            }
+            const timeLeft = { minutes, seconds, percentage };
+            setTime(timeLeft);
+            onTick(timeLeft);
         };
+        a11yManager.current = new A11yManager({ srPanel, i18n });
```
ribeiroguilherme, line 44: If I am not mistaken, you can do this at the beginning of this file. Better for readability as this needs to be created only once
```js
const a11yManager = useRef(new A11yManager({ srPanel, i18n }));
```

packages/lib/src/components/internal/Countdown/index.tsx
```diff
@@ -8,75 +12,55 @@ interface CountdownProps {
...
-    tick() {
-        const newTime = getTimeDifference(this.state.startTime, this.state.endTime);
+    const { i18n } = useCoreContext();
+    const { srPanel } = useSRPanelContext();
 
-        if (newTime.completed) {
-            this.props.onCompleted();
-            return this.clearInterval();
-        }
+    const a11yManager = useRef(null);
```
ribeiroguilherme, line 28: Minor thing - just sharing: the other idea that we discussed before, more "reactish" way, would be to use custom hook for this:

```js
const { updateCoutdownSrPanel,  teardownCoutdownSrPanel } = useCoutdownA11yReporter({ srPanel, i18n });
```

The syntax could be less verbose IMO. Instead of `a11yManager.current.update(time)` you could do `updateCoutdownSrPanel(time)`

packages/lib/src/components/internal/Countdown/A11yManager.ts
```diff
@@ -0,0 +1,73 @@
+import { interpolateElement } from '../../../language/utils';
+
+export class A11yManager {
```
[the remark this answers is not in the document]
longyulongyu: Hmm, I think since it's under the Countdown folder, it would be implicitly belonging to the `Countdown`?

packages/lib/src/components/internal/Countdown/A11yManager.ts
```diff
@@ -0,0 +1,73 @@
...
+    protected TRANSLATION_KEY = 'sr.wechatpay.timetopay';
+    protected LONG_TIMEOUT = 180_000;
+    protected MID_TIMEOUT = 60_000;
+    protected SHORT_TIMEOUT = 30_000;
```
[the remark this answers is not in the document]
longyulongyu: That's a good point, there are some browsers not supporting this, or only newer versions support it. Will change.

packages/lib/src/components/internal/Countdown/A11yManager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    private srPanel;
+    private srInterval;
+    private i18n;
+    private timeout;
+    private timeLeft;
```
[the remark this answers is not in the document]
longyulongyu: Will add.

packages/lib/src/components/internal/Countdown/A11yManager.ts
```diff
@@ -0,0 +1,73 @@
...
+        const setSrMessages = () => {
+            this.srPanel.setMessages(null);
+            this.srPanel.setMessages(this.getSrMessages(this.timeLeft));
+        };
+        // To execute immediately
+        setSrMessages();
+        this.srInterval = setInterval(setSrMessages, timeout);
+    }
+
+    private getSrMessages({ minutes, seconds }) {
```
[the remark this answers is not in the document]
longyulongyu: Will add.

packages/lib/src/components/internal/Countdown/index.tsx
```diff
@@ -8,75 +12,55 @@ interface CountdownProps {
...
-        onTick: () => {},
-        onCompleted: () => {}
-    };
-
-    protected interval;
+function Countdown({ minutesFromNow, onTick = () => {}, onCompleted = () => {} }: CountdownProps) {
+    const secondsFromNow = minutesFromNow * 60000;
+    const nowTime = new Date().getTime();
+    const startTime = new Date(nowTime);
+    const endTime = new Date(nowTime + secondsFromNow);
```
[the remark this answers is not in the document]
longyulongyu: Thanks, nice spot again, will change.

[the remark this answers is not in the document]
longyulongyu: Would it be better to use `useMemo` since we don't need to cache function?

packages/lib/src/components/internal/Countdown/index.tsx
longyulongyu: This style is indeed less verbose and easy to read, but I am struggling to find a proper way to use/implement it.
I need to provide the context (initiate the context) somewhere. As a result, the mark down will become more verbose.. Or there is another way to do so?
```html
<CoutdownA11yContext.Provider value={{ srPanel, i18n }}>
       <Countdown />
</CoutdownA11yContext.Provider>
```
However, I could implement something like this:
```javascript
function useCoutdownA11yReporter({ srPanel, i18n }) {
    const a11y = new A11yManager({ srPanel, i18n });
    return { updateCoutdownSrPanel: a11y.update, teardownCoutdownSrPanel: a11y.tearDown };
}
```
Furthermore `useCoutdownA11yReporter` will be called only once on mount.

packages/lib/src/components/internal/Countdown/index.tsx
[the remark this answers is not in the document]
sponglord: `useMemo` is really only for very complex calculations

packages/lib/src/components/internal/Countdown/A11yManager.ts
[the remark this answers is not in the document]
ribeiroguilherme: Maybe when this code gets transpiled, this will be converted to 'old' Javascript. You can also check that instead as well

packages/lib/src/components/internal/Countdown/index.tsx
ribeiroguilherme: Why would you need a context for it? I believe you can do it with custom hook without providing context/provider.

Context/Provider is useful when you have to pass things down the tree for different Components, avoiding prop drilling. In this case, seems like we do not need it,

longyulongyu: I read it wrongly somehow, hence I was thinking about the provider/context.

packages/lib/src/components/internal/Countdown/useCountdownA11yReporter.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useRef } from 'preact/hooks';
+import { CountdownA11yReporter } from './CountdownA11yReporter';
+import useSRPanelContext from '../../../core/Errors/useSRPanelContext';
+import useCoreContext from '../../../core/Context/useCoreContext';
+import { CountdownTime } from './types';
+
+export const useCountdownA11yReporter = (time: CountdownTime): void => {
+    const { i18n } = useCoreContext();
+    const { srPanel } = useSRPanelContext();
+    const reporter = useRef<CountdownA11yReporter>(null);
```
longyulongyu, line 10: `useRef` keeps reinitiating new instance if we set `initial value` to `new CountdownA11yReporter({ i18n, srPanel })`. 
Hence I have to use `reporter.current = new CountdownA11yReporter({ i18n, srPanel });` in the `useEffect`.

packages/lib/src/components/internal/Countdown/index.tsx
longyulongyu: See [my comment](https://github.com/Adyen/adyen-web/pull/2182#discussion_r1211793950)

sponglord:
So can the `A11yManager` component be deleted?

longyulongyu:
> So can the `A11yManager` component be deleted?

`A11yManager` is renamed to `CountdownA11yReporter`. Do you mean move all the logic from `CountdownA11yReporter` to the custom hook?

sponglord:
> > So can the `A11yManager` component be deleted?
> 
> `A11yManager` is renamed to `CountdownA11yReporter`. Do you mean move all the logic from `CountdownA11yReporter` to the custom hook?

No I mean delete the `A11yManager.tsx` file. It still exists in the `Countdown` directory

longyulongyu:
> > > So can the `A11yManager` component be deleted?
> > 
> > 
> > `A11yManager` is renamed to `CountdownA11yReporter`. Do you mean move all the logic from `CountdownA11yReporter` to the custom hook?
> 
> No I mean delete the `A11yManager.tsx` file. It still exists in the `Countdown` directory

Hm, it's not there [Countdown directory](https://github.com/Adyen/adyen-web/tree/feat/accessibility_qr_code_count_down/packages/lib/src/components/internal/Countdown). Maybe you didn't pull the latest?

sponglord:
> > > > So can the `A11yManager` component be deleted?
> > > 
> > > 
> > > `A11yManager` is renamed to `CountdownA11yReporter`. Do you mean move all the logic from `CountdownA11yReporter` to the custom hook?
> > 
> > 
> > No I mean delete the `A11yManager.tsx` file. It still exists in the `Countdown` directory
> 
> Hm, it's not there [Countdown directory](https://github.com/Adyen/adyen-web/tree/feat/accessibility_qr_code_count_down/packages/lib/src/components/internal/Countdown). Maybe you didn't pull the latest?

It's still there in my project... weird

longyulongyu:
> > > > > So can the `A11yManager` component be deleted?
> > > > 
> > > > 
> > > > `A11yManager` is renamed to `CountdownA11yReporter`. Do you mean move all the logic from `CountdownA11yReporter` to the custom hook?
> > > 
> > > 
> > > No I mean delete the `A11yManager.tsx` file. It still exists in the `Countdown` directory
> > 
> > 
> > Hm, it's not there [Countdown directory](https://github.com/Adyen/adyen-web/tree/feat/accessibility_qr_code_count_down/packages/lib/src/components/internal/Countdown). Maybe you didn't pull the latest?
> 
> It's still there in my project... weird. Must just be WebStorm behaving oddly.

Sometimes `git reset --hard` to the latest commit might work.

PR #2187, opened 2023-05-23
Adding extra checks on the config sent when instantiating Checkout

sponglord:
## Summary
Adding checks for:
- when `installmentOptions` are set on top level Checkout config object (gives warning that this is not the correct place to set this)
- when a 'paymentMethodsConfiguration` object is passed when configuring Dropin (gives warning that this will not work)
- when trying to create a component and passing an unrecognisable object in place of the expected string (enhances error message with a stringified representation of what was actually passed)

## Tested scenarios
The second and third checks both catch issues that have previously been reported in Q&A

### DIFF
--- a/packages/lib/src/core/core.ts
+++ b/packages/lib/src/core/core.ts
@@ -180,6 +180,12 @@ class Core {
             );
         }
 
+        if (hasOwnProperty(options, 'installmentOptions')) {
+            console.warn(
+                "WARNING: you are setting installmentOptions directly in the top level configuration object. They should be set via the 'paymentMethodsConfiguration' object or directly on the 'card' component."
+            );
+        }
+
         this.options = { ...this.options, ...options };
         this.options.loadingContext = resolveEnvironment(this.options.environment);
         this.options.cdnContext = resolveCDNEnvironment(this.options.resourceEnvironment || this.options.environment);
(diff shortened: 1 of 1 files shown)


packages/lib/src/core/core.ts
```diff
@@ -330,7 +342,9 @@ class Core {
      */
     private handleCreateError(paymentMethod?): never {
         const paymentMethodName = paymentMethod && paymentMethod.name ? paymentMethod.name : 'The passed payment method';
-        const errorMessage = paymentMethod ? `${paymentMethodName} is not a valid Checkout Component` : 'No Payment Method component was passed';
+        const errorMessage = paymentMethod
+            ? `${paymentMethodName} is not a valid Checkout Component. (What was passed:${JSON.stringify(paymentMethod)})`
+            : 'No Payment Method component was passed';
```
ribeiroguilherme, line 347: If I am not mistaken, most of the times this error happens is because merchants try to instantiate a payment method but it is not being returned as part of the /paymentMethods response (maybe it wasn't configured properly on BO).

So it could be that it is a valid Checkout component, but it is not configured in the merchant account.

Maybe we can enhance the message and say something like : 
`Failed to create 'txvariant' . Check if this payment is configured in the Backoffice or if the txVariant is a valid one.`

ribeiroguilherme (approved):

ribeiroguilherme (approved):

Issue #1691, opened 2022-08-15
[Accessibility] Card images have no alt description


gazjoy:
**Describe the bug**
The payment cards have no alt text but they are conveying information to a sighed user that a low vision or screen reader user does not get. (Which card type).

**To Reproduce**
1. Go to 'card number' field
2. Enter a valid card number
3. Adyen works out the card type (Visa)
4. The card images go dimmed (visually)
5. There is no alt text in the image or anything readable by a screen reader to convey the same information

**Expected behavior**
We could convey this information to a screen reader users when we know the card type. When we know the card type we can add the alt text to the card image. As the user moves to the next field they should hear something like "Card type Visa" being announced when the screen reader encounters the image.

**Screenshots**
<img width="676" alt="image" src="https://user-images.githubusercontent.com/628064/184648361-31630df9-7079-4469-840d-cba5ef41314f.png">

**Desktop (please complete the following information):**
 - OS: macOS Monterey 12.5
 - Browser: Safari 15.6

sponglord:
Hi @gazjoy - thanks for raising this issue.

Due to merchant complaints in the past about the card brand logos being "extraneous information" for the shopper (with regards to what gets read out by the screenreader), we chose to adopt a "decorative image" approach re. https://www.w3.org/WAI/tutorials/images/decorative/
So these images either have an `alt=""` or an `aria-hidden="true"` attribute.

But on the other hand they _do_ convey information... but is this _vital_ for the shopper?

For this, and all other a11y questions, and for our general approach to a11y - we are currently in the process with an external US-based ADA / WCAG accessibility agency of having our frontend code reviewed. They are now finalising their reports with their findings and we expect to receive this in the coming weeks 
The main reason why we engaged this very experienced third-party is to get conclusive answers to these kind of questions and whether, and how, to fix certain accessibility related issues

gazjoy:
@sponglord its good to hear you are actively doing something about the accessibility issues. Can you tell us who the agency is and a bit about their approach? Are they including people in the testing or only using automated tooling to check against WCAG?

sponglord:
Hi @gazjoy They are a US-based ADA/WCAG accessibility agency. It's not only automated testing, but also includes comprehensive manual testing using a variety of methods, including functional testing by users with disabilities using leading assistive technologies.

gazjoy:
[...]

PR #2188, opened 2023-05-24
Remove aria-hidden attribute on brand icon

sponglord:
## Summary
The brand image in the credit card, visible once a brand has been identified, should have an `alt` tag detailing the card brand, and thus making it readable by screenreaders. 

This was the case _except_  we were also adding an `aria-hidden` attribute, which hid this element from screenreaders.

This has now been fixed. 

## Tested scenarios
`aria-hidden` attribute no longer set on brand or dual-brand icons

**Fixed issue**:  #1691

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/components/BrandIcon.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/components/BrandIcon.tsx
@@ -18,7 +18,6 @@ export default function BrandIcon({ brand, brandsConfiguration = {} }: BrandIcon
             onError={handleError}
             alt={brand}
             src={imageUrl}
-            aria-hidden={'true'}
         />
     );
 }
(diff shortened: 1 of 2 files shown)


m1aw (approved):

m1aw (changes requested):
1. I had approved but then I went to check and we should also be loading the alt readable text from BRAND_READABLE_NAME_MAP.
2. This actually makes me thing we should make this a prop or function and just leave it in place so it works when we get the brands name form the `paymentMethods` call (so a util function won't work).

We can also do 2. later.

sponglord:
@m1aw - I don't understand what your comment has to do with removing the `aria-hidden` attributes?
It sounds like what your are proposing belongs in another PR?

m1aw (approved):

PR #2193, opened 2023-05-30
fix: Add onOrderCancel to CoreOptions

m1aw:
## Summary

Updates CoreOptions to also have onOrderCancel. Related to #1987 

## Tested scenarios

- Run elslint and checked with existing typings

**Fixed issue**:

### DIFF
--- a/packages/lib/src/core/types.ts
+++ b/packages/lib/src/core/types.ts
@@ -100,9 +100,24 @@ export interface CoreOptions {
 
     setStatusAutomatically?: boolean;
 
-    beforeRedirect?(resolve: PromiseResolve, reject: PromiseReject, redirectData: { url: string; method: string; data?: any }): Promise<void>;
+    beforeRedirect?(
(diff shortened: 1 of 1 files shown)


packages/lib/src/core/types.ts
```diff
@@ -120,6 +120,7 @@ export interface CoreOptions {
 
     onOrderRequest?(resolve: PromiseResolve, reject: PromiseReject, data: PaymentData): Promise<void>;
 
+    onOrderCancel?(order: Order): void;
```
ribeiroguilherme, line 123: Are we sure we return the `order` directly? I think we return an object which contains the `order` (by checking [this line](https://github.com/Adyen/adyen-web/blob/master/packages/lib/src/components/Dropin/components/DropinComponent.tsx#L103))

ribeiroguilherme (approved):

ribeiroguilherme (approved):

PR #2205, opened 2023-06-01
Fixes mismatched label for Selectbox

m1aw:
## Summary

This is a fix for a problem that came back from the second round of a11y testing. The `label` was pointing for a different element then the `combobox`. This occurs because this element that had the `id` can be either a button or div depending  if it's a filterable select or not.

Also remove `title`, not sure if this was doing anything, but doesn't seem necessary for a11y purposes. And was causing duplication of placeholder when using the Chrome a11y inspector.

## Tested scenarios

- Tested with Voice Over to make sure "Country" was announced
- Tested with Chrome a11y inspector 

**Fixed issue**:


sponglord:
For cross referencing purposes can you add the a11y test `instanceId` to your description, plz

sponglord (approved):

PR #2207, opened 2023-06-02
v6: showPayButton defaults to true

sponglord:
## Summary
Update `showPayButton` property to default to `true` for both Components and Drop-in.

Previously it defaulted to `true` for Drop-in and `false` for Components (due legacy reasons, when we didn't have a separate `PayButton` component).

## Tested scenarios
Tested across:
- AmazonPay,
- GooglePay,
- Oxxo,
- IssuerLists,
- PayPal,
- MultiBanco,
- BankTransfer,
- InstantPaymentMethods,
- Pix,
- Redirect,
- Twint,
- Vipps

All of which were setting their own default values for `showPayButton`

Also tested in Dropin, Cards & Components generally


packages/lib/src/components/helpers/QRLoaderContainer.tsx
```diff
@@ -27,14 +28,15 @@ export interface QRLoaderContainerProps extends UIElementProps {
...
+    // Using the generic here allow to fully extend the QRLoaderContainer (including its props)
     protected static defaultProps = {
         qrCodeImage: '',
         amount: null,
         paymentData: null,
         onError: () => {},
         onComplete: () => {},
-        onActionHandled: () => {}
+        onActionHandled: () => {},
+        showPayButton: false
```
ribeiroguilherme, line 39: Why `showPayButton: false` here?

packages/playground/src/pages/Components/Components.js
```diff
@@ -97,10 +96,13 @@ getPaymentMethods({ amount, shopperLocale }).then(async paymentMethodsResponse =
     window.upi = checkout.create('upi').mount('.upi-field');
 
     // PIX
-    window.pix = checkout.create('pix', { countdownTime: 5 }).mount('.pix-field');
+    window.pix = checkout.create('pix', { countdownTime: 5, showPayButton: true }).mount('.pix-field');
 
     // Oxxo
-    window.oxxo = checkout.create('oxxo').mount('.oxxo-field');
+    window.oxxo = checkout.create('oxxo', { showPayButton: true }).mount('.oxxo-field');
+
```
ribeiroguilherme, line 103: Do we need to set `showPayButton: true` here?

packages/lib/src/components/Dropin/elements/createInstantPaymentElements.ts
```diff
@@ -9,6 +9,6 @@ import UIElement from '../../UIElement';
  * @param create - Reference to the main instance `Core#create` method
  */
 const createInstantPaymentElements = (paymentMethods: PaymentMethod[] = [], props, create): Promise<UIElement[]> | [] =>
-    paymentMethods.length ? createElements(paymentMethods, { ...props, isInstantPayment: true, showPayButton: true }, create) : [];
```
ribeiroguilherme, line 12: We need the `showPayButton: true` here for Instant payment methods - they are not supposed to be used with custom pay button

ribeiroguilherme (approved):

packages/lib/src/components/helpers/QRLoaderContainer.tsx
sponglord: QRCode components, outside of the Dropin, don't need to show a pay button

packages/playground/src/pages/Components/Components.js
sponglord: This is purely so the Pix component in the components page in the playground isn't just an empty html element.
TBH I don't know why components like these are added to the components page

packages/lib/src/components/Dropin/elements/createInstantPaymentElements.ts
sponglord: ok

packages/lib/src/components/helpers/QRLoaderContainer.tsx
ribeiroguilherme: But we want the showPayButton to default to `true`, right? In this case, it is defaulting to false.

I tested on dev playground, and if I create the Swish component ( `checkout.create('swish')...`). , the Pay button isn't displayed with this change. Removing the default prop make the button show up.

sponglord: Ok, I'll set it to true!

packages/lib/src/components/BcmcMobile/BcmcMobile.test.ts
```diff
@@ -20,7 +20,7 @@ describe('BcmcMobile', () => {
     describe('render', () => {
         test('does not render anything by default', () => {
```
ribeiroguilherme, line 21: @sponglord I think the test descriptions are misleading now, as we are rendering the something by default!

sponglord: Doh! I'll fix them!

Issue #2209, opened 2023-06-05
loadingContext is missing in CoreOptions


david-eventfrog:
Error: node_modules/@adyen/adyen-web/dist/types/core/Analytics/Analytics.d.ts:2:48 - error TS2344: Type '"amoun" | "loadingContext" | "locale" | "clientKey" | "analytics"' does not satisfy the constraint 'keyof CoreOptions.
  Type '"loadingContext"' is not assignable to type 'keyof CoreOptions'.

2 export type AnalyticsProps = Pick<CoreOptions, 'loadingContext' | 'locale' | 'clientKey' | 'analytics' | 'amont'>;

As you can see on the error, there is a problem with the 'loadingContext' in the CoreOptions.
In your CoreOptions (../types/core/types.ts) 'loadingContext' is missing. Please add the attribute and fix your package.

ribeiroguilherme:
Hi @david-eventfrog ,

Which version are you testing? It seems like we have the [loadingContext](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/types.ts#L156) defined in the CoreOptions object

david-eventfrog:
Hi @ribeiroguilherme 

I am using version 5.44.0.

And yes, I see you have `loadingContext` defined in your `CoreOptions`.

BUT: You have declared it as `@internal`, see [here](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/types.ts#L154) and set `stripInternal` to true in [tsconfig.json](https://github.com/Adyen/adyen-web/blob/main/packages/lib/tsconfig.json#L18). So the variable is not exposed and not available in your package.

You can find more information about this topic [here](https://www.typescriptlang.org/tsconfig#stripInternal).

PR #2213, opened 2023-06-08
Typescript - Disabling stripInternal feature

ribeiroguilherme:
## Summary
We use JSDoc `@internal` annotation in some places to document that certain types are internal types, but we still want to keep those types in the exported type definition files.  This PR tweaks the tsconfig to not strip these types from the final build

**Fixed issue**: #2209

### DIFF
--- a/packages/lib/tsconfig.json
+++ b/packages/lib/tsconfig.json
@@ -15,7 +15,6 @@
         "noImplicitAny": false,
         "strictPropertyInitialization": false,
         "suppressImplicitAnyIndexErrors": true,
-        "stripInternal": false,
 
         "importHelpers": true,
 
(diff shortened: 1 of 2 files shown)


sponglord (approved):

packages/lib/tsconfig.json
```diff
@@ -15,7 +15,7 @@
         "noImplicitAny": false,
         "strictPropertyInitialization": false,
         "suppressImplicitAnyIndexErrors": true,
-        "stripInternal": true,
+        "stripInternal": false,
```
david-eventfrog, line 18: Since false is the default value, I would recommend simply removing the property. This also gives you a better overview if only properties that do not have the default value can be found in the tsconfig file.

PR #2230, opened 2023-06-21
fix(sr-panel): remove existing SR panel container when initiating SRPanel

longyulongyu:
## Summary
Remove existing SR panel containers (if there are) when initiating SRPanel, as a result all checkout instances share one SR panel.

Reusing the existing SR panel messes up (or loses the `SRMessages` reference) with the `SRMessages` reference, hence it's better to remove all SR panel containers and remount it again.

## Tested scenarios
Tested in `MyStoreDemo`, only one SR panel exists in the DOM when switching pages from the checkout to any other pages.

### DIFF
--- a/.changeset/neat-llamas-build.md
+++ b/.changeset/neat-llamas-build.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Remove existing SR panel container when initiating SRPanel, as a result all checkout instances share one SR panel.
--- a/packages/lib/src/core/Errors/SRPanel.tsx
+++ b/packages/lib/src/core/Errors/SRPanel.tsx
@@ -43,10 +43,15 @@ export class SRPanel extends BaseElement<SRPanelProps> {
         if (this.props.enabled) {
             this._enabled = true;
             if (document.querySelector(this.props.node)) {
+                const containers = document.getElementsByClassName('sr-panel-holder');
+                if (Array.from(containers).length) {
+                    Array.from(containers).forEach(ele => {
+                        ele.remove();
+                    });
+                }
                 this.srPanelContainer = document.createElement('div');
                 this.srPanelContainer.className = 'sr-panel-holder';
                 this.srPanelContainer.id = this.id;
-
                 document.querySelector(this.props.node).appendChild(this.srPanelContainer);
                 this.mount(this.srPanelContainer);
             } else {
--- a/packages/lib/src/core/Errors/SRPanel.tsx
+++ b/packages/lib/src/core/Errors/SRPanel.tsx
@@ -43,6 +43,15 @@ export class SRPanel extends BaseElement<SRPanelProps> {
         if (this.props.enabled) {
             this._enabled = true;
             if (document.querySelector(this.props.node)) {
+                /*
+                We want to remove the sr panel container from the DOM after the checkout instance destroyed.
+                SRPanel extends from BaseElement which is not a preact component,
+                there is no way to know when component unmounts.
+                Hence, we remove all sr panel container elements during the initializing phase.
+                A 'side effect' to this approach - if there are multiple adyen checkout instances,
+                there will be only one sr panel which belongs to the last instance.
+                Usually there is only one checkout instance.
+                 */
                 const containers = document.getElementsByClassName('sr-panel-holder');
                 if (Array.from(containers).length) {
                     Array.from(containers).forEach(ele => {


sponglord (dismissed):

sponglord:
Will you add a comment in `SRPanel` about why you've added this code - what issue/scenario it fixes, plz.

Also I've approved it - but can you hold back on the merge... I need to test how it will work in PBL

longyulongyu:
> Will you add a comment in `SRPanel` about why you've added this code - what issue/scenario it fixes, plz.
> 
> Also I've approved it - but can you hold back on the merge... I need to test how it will work in PBL

Sure, no worries!

sponglord:
>  I need to test how it will work in PBL

This fix doesn't work in  PBL. The generic nature of the search for a pre-existing SRPanel, i.e. using the class name, is removing _both*_ SRPanels that operate in PBL.

I think I know what the solution is (to remove the panel based on _id_) I just need to test it...

_* PBL has 2 instances of Checkout, one for the personalDetails form, one for the actual Dropin_

longyulongyu:
> 

ok, this would indeed won't work, the sr panel will belong to the last checkout instance. 
Removing by id would be a safer option.

sponglord:
This is how the code needs to look:
```
const preExistingSRPanel = document.getElementById(this.id);
if (preExistingSRPanel) {
    document.querySelector(this.props.node).removeChild(preExistingSRPanel);
}
```

longyulongyu:
> This is how the code needs to look:
> 
> ```
> const preExistingSRPanel = document.getElementById(this.id);
> if (preExistingSRPanel) {
>     document.querySelector(this.props.node).removeChild(preExistingSRPanel);
> }
> ```

> id

Hmm, but with this one, we will remove the sr panel from the first instance (because the id is the same), will the first instance still knows how to communicate with its own panel?
Because:
```
1st checkout -> produce sr panel id = 'ariaLiveSRPanel'
2nd checkout => find the sr panel id = 'ariaLiveSRPanel' and remove it, create another one id = 'ariaLiveSRPanel'

1st checkout cannot feed messages into sr panel because the element is removed.
```
I have a proposal which requires a bit of refactoring, which is to let Preact handles all the cleanup, as a result every checkout instance has its own sr panel, and will be removed by Preact when the instance is destroyed. 

The `RenderInBody` is a utility component, which will render anything inside the `this.container`, `this.container` is inside the `body` element.

```javascript
import { Component, Fragment, render } from 'preact';

export class RenderInBody extends Component {
    componentDidMount() {
        this.container = document.createElement('div');
        this.container.className = 'container-test';
        document.body.appendChild(this.container);
        this._renderLayer();
    }
    componentWillUnmount() {
        document.body.removeChild(this.container);
    }

    _renderLayer() {
        render(this.props.children, this.container);
    }

    render() {
        return <Fragment />;
    }
}
```
`SRPanel` renders inside the `RenderInBody` component.

```javascript
import { Component } from 'preact';
import { RenderInBody } from './RenderInBody';

export class SRPanel extends Component {
    render() {
        const srPanel = <div id="12345">this is a srPanel</div>;
        return <RenderInBody>{srPanel}</RenderInBody>;
    }
}
```

sponglord:
> 1st checkout cannot feed messages into sr panel because the element is removed.

If you're going to have multiple instances of checkout then you should give each SRPanel it's own id!

I have manually tested my proposal extensively both in a scenario where Checkout keeps getting re-initialised as well as with the PBL setup.

However, I would like to try your `renderInBody solution - can you perhaps make it in a separate PR so I can test it in the playground and PBL?

longyulongyu:
> > 1st checkout cannot feed messages into sr panel because the element is removed.
> 
> If you're going to have multiple instances of checkout then you should give each SRPanel it's own id! (This is what is done in PBL)
> 
> I have manually tested my proposal extensively both in a scenario where Checkout keeps getting re-initialised as well as with the PBL setup.
> 
> However, I would like to try your `renderInBody solution - can you perhaps make it in a separate PR so I can test it in the playground and PBL? And are you sure that this _will_ tap into the component _unmount_ lifecycle?

Hmm, i am a bit confused, because `this.id` only knows the current instance's `id`. If every time, we pass a different id the new checkout instance just doesn't remove anything. But it's my speculation 😝, it also depends on how PBL's implementation. If it works in PBL, we can for sure just use it!

`componentWillUnmount` will be called, I created and tested this in `MyStoreDemo`. Let me try it out though...

longyulongyu:
> > 1st checkout cannot feed messages into sr panel because the element is removed.
> 
> If you're going to have multiple instances of checkout then you should give each SRPanel it's own id! (This is what is done in PBL)
> 
> I have manually tested my proposal extensively both in a scenario where Checkout keeps getting re-initialised as well as with the PBL setup.
> 
> However, I would like to try your `renderInBody solution - can you perhaps make it in a separate PR so I can test it in the playground and PBL? And are you sure that this _will_ tap into the component _unmount_ lifecycle?

Ahh i see what you mean, it will work. I was thinking the something different, like passing the random id to the panel, but no one would do that..

longyulongyu:
Close this one as it doesn't work

PR #2234, opened 2023-06-22
Feature/using checkoutanalytics

sponglord:
## Summary
Updating Analytics functionality to work against the new `/checkoutanalytics` endpoint
This involves:
- a new procedure to retrieve `checkoutAttemptId` and send in initial analytics details (`channel`, `flavor`, `locale` etc)
- a new events queue for the "error", "logs" & "info" that the new endpoint expects
- generating "log" events for when: 
    - `onSubmit` is called (indicating a "Pay" button press)
    - `createFromAction` is called (indicating a `/payments` response mandating an "action")
- generating "info" events for when a component is rendered, the Dropin PM list is rendered, a PM is selected in the Dropin
- groundwork for the checkoutanalytics events that the 3DS2 process will generate

## Tested scenarios
New unit tests added

**Relates to issue**:  COAPI-6


packages/lib/src/core/core.ts
```diff
@@ -143,6 +149,18 @@ class Core {
         }
 
         if (action.type) {
+            // Call analytics endpoint
+            const aObj: AnalyticsObject = createAnalyticsObject({
+                class: 'log',
+                component: `${action.type}${action.subtype}`,
+                type: ANALYTICS_ACTION_STR,
+                subtype: capitalizeFirstLetter(action.type),
+                message: `${action.type}${action.subtype} is initiating`
```
ribeiroguilherme, line 158: Can't we somehow move this logic to the Analytics module? This way we keep analytics logic all there and do not spread in the Core/UIElement

```js
analytics.createAnalyticsEvent({ type: 'log', data: { class, component, type, etc.. } });
```

In the analytics module:
```js
createAnalyticsEvent({ type, data }) {
  const analyticsData =  createAnalyticsObject(...data, etc);
  this.addAnalyticsAction(type, analyticsData);
}

File: packages/lib/src/core/Analytics/utils.ts [hunk 2]
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Log objects have, in addition to the base props:
+ *  "type" & "target" (e.g. when onSubmit is called after a pay button click), or,
+ *  "type" & "subtype" & "message" (e.g. when an action is handled), or,
+ *  "type" & "message" (e.g. logging during the 3DS2 process)
+ *
+ * Event objects have, in addition to the base props:
+ *   "type" & "target"
+ */
+export const createAnalyticsObject = (aObj): AnalyticsObject => ({
```
ribeiroguilherme, line 21: Is there any specific type for the `aObj` or should it be `any` in this case?

packages/lib/src/components/UIElement.tsx
```diff
@@ -45,6 +48,25 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
         return state;
     }
 
+    /* eslint-disable-next-line */
+    protected submitAnalytics(obj = null) {
```
ribeiroguilherme, line 52: The `obj` parameter isn't not used in this function

packages/lib/src/core/Analytics/CAEventsQueue.ts
```diff
@@ -0,0 +1,72 @@
...
+    events: AnalyticsObject[];
+    errors: AnalyticsObject[];
+    logs: AnalyticsObject[];
+}
+
+export interface EQObject {
+    add: (t, a) => void;
+    run: (id) => Promise<any>;
+    getQueue: () => CAActions;
+    _runQueue: (id) => Promise<any>;
```
ribeiroguilherme, line 15: Missing types?

packages/lib/src/core/Analytics/CAEventsQueue.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const CAEventsQueue = ({ analyticsContext, clientKey, analyticsPath }: EventQueueProps) => {
```
ribeiroguilherme, line 18: What does `CA` stands for? `CheckoutAdyen`? 
(Every time I read it I think it is `CustomerArea` 😅  )

I find it very hard to read and understand the types/objects when they get abbreviated like that (CAActions, EQObject, and so on). We don't have guidelines for that, but I'd rather try to not abbreviate that much or try to be more explicit. I believe it improves the readability of the codebase

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,74 +1,82 @@
...
+import { AnalyticsModule } from '../../components/types';
 
-export type AnalyticsProps = Pick<CoreOptions, 'loadingContext' | 'locale' | 'clientKey' | 'analytics' | 'amount'>;
+export type AnalyticsProps = Pick<CoreOptions, 'loadingContext' | 'locale' | 'clientKey' | 'analytics' | 'amount' | 'analyticsContext'>;
 
-class Analytics {
-    private static defaultProps = {
+let _checkoutAttemptId = null;
+
+const Analytics = ({ loadingContext, locale, clientKey, analytics, amount, analyticsContext }: AnalyticsProps) => {
```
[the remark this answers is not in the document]
sponglord: My thinking on this was why use a class when you don't need one?
Classes are something of an artificial construct for javascript and add add weight to the bundle.

packages/lib/src/components/UIElement.tsx
```diff
@@ -45,6 +48,25 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
...
+
+        const aObj: AnalyticsObject = createAnalyticsObject({
+            class: 'log',
+            component,
+            type: ANALYTICS_SUBMIT_STR,
+            target: 'pay_button'
+        });
+
+        this.props.modules?.analytics.addAnalyticsAction('log', aObj);
+    }
```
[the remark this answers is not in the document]
sponglord: Good point

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,74 +1,82 @@
...
             }
+        },
 
-            if (telemetry === true) {
-                const telemetryTask = checkoutAttemptId =>
-                    this.logTelemetry({ ...event, ...(payload && { ...payload }), checkoutAttemptId }).catch(() => {});
+        getCheckoutAttemptId: (): string => _checkoutAttemptId,
 
-                this.queue.add(telemetryTask);
+        addAnalyticsAction: (type: ANALYTICS_ACTION, obj: AnalyticsObject) => {
```
longyulongyu, line 58: I think maybe we could simplify the interface a bit (later stage).

The idea is more like:
`Components` either dispatch custom events which will be captured by the `Analytics` or have the `Analytics` as the context. The only thing Components need to do is to send an event to the `Analytics` module.

`Analytics` is responsible to dump the events to `Queue` and fetch the `checkoutAttemptId` when missing (just some pre work before sending the message to the `Queue`).

`Queue` manages to dispatch the events to the correct queue type - the error queue or log queue.
When the error queue has a message, send it right away. When the log queue has messages, it will wait until it's full. The queue size is configurable.

I think this is more decoupled. The `Queue` can have unlimited types of queues and `Analytics` module doesn't care which type of queue to send. Furthermore, the `Analytics` module doesn't need to know how to `debounce`. 
The `Analytics` only needs to expose one public function to all components - `send`(or none if components just dispatch events.)

packages/lib/src/components/UIElement.tsx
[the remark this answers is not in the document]
sponglord: Because the `onSubmit` function that gets called (which then calls `submitAnalytics`) is part of the _active payment method_ i.e. the `UIElement`, it is not part of `Dropin.tsx`

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,74 +1,82 @@
...
+                if (telemetry === true && !_checkoutAttemptId) {
+                    // fetch a new checkoutAttemptId if none is already available
+                    _collectId({ ...initialEvent, ...(payload && { ...payload }) })
+                        .then(checkoutAttemptId => {
+                            _checkoutAttemptId = checkoutAttemptId;
+                        })
+                        .catch(e => {
+                            // Caught at collectId level. We do not expect this catch block to ever fire, but... just in case...
+                            console.debug(`Fetching checkoutAttemptId failed.${e ? ` Error=${e}` : ''}`);
+                        });
```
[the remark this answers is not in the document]
sponglord: You would then need to wrap the `await` call in a `try-catch` block - which only looks slightly cleaner!
But I'll do it...

packages/lib/src/core/types.ts
```diff
@@ -154,6 +154,11 @@ export interface CoreOptions {
      * @internal
      */
     loadingContext?: string;
+
+    /**
+     * @internal
+     */
+    analyticsContext?: string;
```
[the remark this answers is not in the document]
sponglord: yup

packages/lib/src/components/UIElement.tsx
sponglord: But it will be in the function that `ThreeDS2DeviceFingerprint` & `ThreeDS2DeviceChallenge` override

packages/lib/src/core/core.ts
sponglord: yeah, good call

packages/lib/src/core/Analytics/utils.ts
sponglord: Fixed

packages/lib/src/core/Analytics/CAEventsQueue.ts
sponglord: CA stands for CheckoutAnalytics of course! :-D
However I only added the "CA" to distinguish it from the existing EventsQueue. Now that is gone we can drop the prefix

packages/lib/src/core/Analytics/CAEventsQueue.ts
sponglord: maybe :-p

packages/lib/src/core/Analytics/Analytics.ts
sponglord: Maybe a refactor for later _if_ we end up needing a more generic kind of queue. 
At the moment Analytics & EventsQueue only have one purpose - to access the analytics endpoint

packages/lib/src/core/Analytics/Analytics.ts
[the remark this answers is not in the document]
ribeiroguilherme: Sorry, this is actually called clojure 🤓

Not sure what do you mean by saying that classes are an artificial construct. Classes are supported natively since a long time! And in terms of weight - I don't think they are very much different after all. I didn't run the test by myself, but I couldn't find anything saying that this is a big deal.

I agree that is better for React.Components to be moved to functional components for sure - that is the "new" standard of writing UI components. But that does not mean that we are forced to write other parts of the code base (which are not Preact components) using functional approach.

My main point is: we generally don't use clojures in our project to write JS modules. We usually leverage Typescript/Classes or simple helper functions. Do we want to start introducing another way of writing/encapsulating objects/modules in the codebase? All Core modules are class-based - do we want to rewrite this one and make it different than the others, following another pattern? I'd rather avoid that and keep it as we are doing currently. If we want to introduce another way of writing modules, then we should discuss internally our coding guidelines and align on this.

PR #2250, opened 2023-07-06
POC: Tree shaking UI components

ribeiroguilherme:
### Proposal for introducing tree shaking for Payment methods, in order to reduce the bundle size.

In order to support tree shaking, we need to slightly change how merchants integrate with us.
There are two possibilities:

-  In case merchant does not care about bundle size and wants to go live ASAP:
```js
import AdyenCheckout from '@adyen/adyen-web/auto';
import '@adyen/adyen-web/styles/adyen.css';
```

- In case the merchant wants to have control of the bundle size, he can import the specific components that will be used in the page:
```js
import { AdyenCheckout, Card, PayPal, ApplePay, Boleto, Ideal } from '@adyen/adyen-web';
import '@adyen/adyen-web/styles/adyen.css';

AdyenCheckout.register(Card, PayPal, ApplePay, Boleto, Ideal);
```

 In terms of integration, we are changing how the merchant is importing the library AND adding the need of registering/injecting the required Payment methods before rendering them. The rest of the library API stays the same.

Some technical details about the changes:
- We were using a mapping '`txVariant` -> `Component` to get the proper Payment Method, and to use this mapping we had to import all Components.  Now, it became the other way around: the Payment Method component contains the txVariants that it supports.  The txVariants supported by the Component are mapped using the `static type` and `static txVariants` fields from UIElement;
- The old mapping '`txVariant` -> `Component`' was kept in the code to allow Typescript to properly assess the types when doing `checkout.create('paypal')`. It is not part of the final bundle though, as it is used only for Types;
- Some PaymentMethods require other Components as dependencies( Ex: `Card` requires `ThreeDS` components; `IssuerList` requires `Redirect`). In order to include them in the bundle and register them as part of the Components mapping, this PR introduces `static dependencies = []` to `UIElement` which holds the reference for these dependencies. Doing that, we avoid doing `AdyenCheckout.register(Card, ThreeDSFingerPrint, ThreeDSChalange)` . We can simply do `AdyenCheckout.register(Card)`;
- With the changes done on `package.json`(`type: module`, `exports: {..`), it means that we support ESM by default;
- `UMD` version auto register all Components - there is no way of tree shaking that. Tree shaking is available only for merchants integrating using ESM;
- As we are dropping IE11 support, I replaced babel and all its polyfills in favor of `SWC`;

## Tested scenarios
Testing using local playground:
1. Run `yarn` to install new dependencies
2. `yarn build`
3. `yarn start` and then check the Components in the page

Testing ESM with tree shaking:
1. Go to the lib folder and run `yarn build`
4. Then run `npm pack` to generate the tarball file
5. Extract the tarball file in a folder; open its `package.json` and remove the script tasks `prepublishOnly` and `prepare`
6. Run `yarn link`
7. Go to another project (playground?), and run `yarn link "@adyen/adyen-web"`
8. Play around importing different Components; build the project files and then run any sort of bundle analyzer on it  ([like this](https://create-react-app.dev/docs/analyzing-the-bundle-size/))

Testing UMD bundle:
1. Copy the generated UMD file to another web project
2. Import it using <script> markup
3. Verify that `window.AdyenCheckout` is available
4. Play around creating Components etc

TODO: There are more scenarios to test which we will perform later after deciding how to move forward


packages/lib/src/components/Card/Card.tsx
```diff
@@ -13,9 +13,12 @@ import { ClickToPayCheckoutPayload, IClickToPayService } from '../internal/Click
 import ClickToPayWrapper from './components/ClickToPayWrapper';
 import { UIElementStatus } from '../types';
 import SRPanelProvider from '../../core/Errors/SRPanelProvider';
+import { ThreeDS2Challenge, ThreeDS2DeviceFingerprint } from '../ThreeDS2';
 
 export class CardElement extends UIElement<CardElementProps> {
     public static type = 'scheme';
+    public static txVariants = ['amex', 'card', 'diners', 'discover', 'jcb', 'kcp', 'maestro', 'mc', 'scheme', 'storedCard', 'visa'];
```
sponglord, line 20: We need to do some research to find out if merchants ever actual go to do `checkout.create('mc')`, for example.
Because then we could thin down this list

sponglord:
Now we are using `swc` is there any possibility we can run into the `swcMinify` issue that many merchants complain about?

packages/lib/src/core/core.ts
```diff
@@ -33,6 +31,12 @@ class Core {
         buildId: process.env.ADYEN_BUILD_ID
     };
 
+    public static registry = registry;
```
sponglord, line 34: Do we _need_ to store the registry as a `static` member?

packages/lib/src/components/Card/Card.tsx
sponglord: This list can, in fact, just be `['card']`. 
If merchants want to create a single branded card they can do so by just setting the `brands` config prop to contain a single txvariant

PR #2258, opened 2023-07-11
Fix: Klarna widget to make drop-in clickable after loaded

longyulongyu:
## Summary
When Klarna widget is loaded inside the drop-in, the drop-in is still in 'loading' status, we need to tell dropin to be 'ready' - clikable.

## Tested scenarios
When we set the following config to the checkout, the Klarna widget renders inside the drop-in and it should not block the drop-in to be clickable after loaded.
 
```javascript
         klarna: { 
               useKlarnaWidget: true 
         }
 };
```

**Internal ticket id**:  COWEB-1251

### DIFF
--- a/.changeset/spicy-crews-worry.md
+++ b/.changeset/spicy-crews-worry.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fix Klarna widget blocks drop-in to be clickable on loaded.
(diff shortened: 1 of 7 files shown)


m1aw:
I don't understand where we are setting the dropin to 'loading' in the first place. Because in the Klarna component we `setStatus` locally.

longyulongyu:
> I don't understand where we are setting the dropin to 'loading' in the first place. Because in the Klarna component we `setStatus` locally.

It's the `payButton`, which i think it serves as a redirect button. When submitted, it calls the `submit` function in `UIElement.tsx`, and setting the loading status for dropin.

m1aw (approved):

PR #2266, opened 2023-07-13
fix: Core update function also updates payment methods list

longyulongyu:
## Summary
It's related to the drop-in advanced flow.

There is a use case that merchants feed `paymentMethodsResponse` via the `update` function instead of passing it  while initiating the `AdyenCheckout`.

Fix the issue so that we are backward compatible.

## Tested scenarios
Tested in the playground drop-in manual flow.

Pass the `paymentMethodsResponse` in the `update` function. The drop-in should render the payment methods list.

```javascript
checkout.update({ paymentMethodsResponse: await getPaymentMethods({ amount, shopperLocale }) });
```

### DIFF
--- a/packages/lib/src/core/core.test.ts
+++ b/packages/lib/src/core/core.test.ts
@@ -242,6 +242,7 @@ describe('Core', () => {
                 environment: 'test',
                 clientKey: 'xxxx'
             });
+            await checkout.initialize();
             const paymentMethodsResponse = { paymentMethods: [{ name: 'Credit Card', type: 'scheme', brands: ['visa'] }] };
             expect(checkout.paymentMethodsResponse).toHaveProperty('paymentMethods', []);
             await checkout.update({ paymentMethodsResponse });
--- a/packages/lib/src/core/core.test.ts
+++ b/packages/lib/src/core/core.test.ts
@@ -236,5 +236,16 @@ describe('Core', () => {
 
             expect(spy).toHaveBeenCalled();
         });
+
+        test('should update the payment method list for the advanced flow', async () => {
+            const checkout = new AdyenCheckout({
+                environment: 'test',
+                clientKey: 'xxxx'
+            });
+            const paymentMethodsResponse = { paymentMethods: [{ name: 'Credit Card', type: 'scheme', brands: ['visa'] }] };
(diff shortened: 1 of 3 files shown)


packages/lib/src/core/core.ts
```diff
@@ -80,6 +80,10 @@ class Core {
 
         this.createCoreModules();
 
+        if (!this.paymentMethodsResponse?.paymentMethods?.length) {
+            this.createPaymentMethodsList();
+        }
```
ribeiroguilherme, line 85: Maybe we don't need this if statement? We could directly call `this.createPaymentMethodsList`.
In the `PaymentMethodsResponse` there is a check in place for this!

packages/lib/src/core/core.ts
```diff
@@ -80,6 +80,10 @@ class Core {
 
         this.createCoreModules();
```
ribeiroguilherme, line 82: Could we then remove the call for `this.createPaymentMethodsList()` in the Core.constructor, and let it be done only in the `initialize()` function? 

It seems to make sense to me. Otherwise, when initializing the lib, we would be calling this twice on advanced and sessions flow

packages/lib/src/core/core.ts
```diff
@@ -80,6 +80,10 @@ class Core {
...
```
ribeiroguilherme, line 81: On this `createCoreModules`, there is a check in place to `console.warn` if modules are already created. I believe we can omit that for PROD version, otherwise merchants will get worried. What do you think?

packages/lib/src/core/core.test.ts
```diff
@@ -236,5 +236,16 @@ describe('Core', () => {
 
             expect(spy).toHaveBeenCalled();
         });
+
+        test('should update the payment method list for the advanced flow', async () => {
```
ribeiroguilherme, line 240: 👏

ribeiroguilherme (approved):

packages/lib/src/core/core.ts
longyulongyu: Yes we can enable it only for dev as it doesn't help the merchants.

packages/lib/src/core/core.ts
longyulongyu: It makes sense! Will update.

PR #2276, opened 2023-07-24
feat(trustly): add description

longyulongyu:
## Summary
Add descriptions for Trustly.

## Tested scenarios
All unit tests pass.

**Internal ticket**:  COWEB-1247


packages/lib/src/components/Trustly/Trustly.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+class TrustlyElement extends RedirectElement {
+    public static type = 'trustly';
+
+    get displayName() {
+        return this.props.name || this.constructor['type'];
+    }
+
+    render() {
+        if (this.props.showPayButton) {
```
ribeiroguilherme, line 15: We should hide only the pay/redirect button if showPayButton is `false`. In this case, it is hiding the whole Component

ribeiroguilherme (approved):

ribeiroguilherme:
It can be nice to add it to the Storybook

longyulongyu:
> 



> It can be nice to add it to the Storybook, but it can be done in another PR

Sure will do that in another PR to add more payment methods in the Storybook.

PR #2293, opened 2023-08-10
Adds ANCV payment method component 

m1aw:
## Summary

This adds support for ANCV voucher await component.

This component is a mix between an await component with a gift card (partial payment method).

Payment flow is like follows:
1. user inserts 11-digit id or email address corresponding to their ANCV account
2. a order is created (no balance check is made contrary to gift card)
3. payments call is made with the order details and full amount
4. we get a await response from payments, and render the await component
5. we get `PartiallyAuthorised` in result code from `status` and request paymentDetails receiving an `order`
6. after this we should behave like giftcard and show dropin again or call the relevant callbacks

The main problems with this current PR is the code duplication. Big parts of the code have been ported from giftcard without a good way to extracting the logic.

Also there's currently missing code coverage. 

## Tested scenarios

**Fixed issue**:


packages/lib/src/components/ANCV/ANCV.tsx
```diff
@@ -0,0 +1,134 @@
...
+    formatData() {
+        return {
+            paymentMethod: {
+                type: ANCVElement.type,
+                beneficiaryId: this.state.data?.beneficiaryId
+            }
+        };
+    }
+
+    private onOrderRequest = data => {
```
m1aw, line 38: This logic should be move elsewhere.

sponglord:
Just wanted to confirm this is also working on `sessions`?

sponglord:
Don't forget to add a .changeset file

theghostbel:
Asked in [PR view](https://github.com/Adyen/adyen-web/commit/cb1564c96feab4d39f7adde81ad0a8327cd8d6e9#r131250051), but will repeat here as well: 
Is removal of Trstly-related translations expected? 
https://github.com/Adyen/adyen-web/pull/2293/files#diff-396d6e3eb5952c0eb94623a25a7cf75f8ca5e186a2b0d5d8609c3673f1e6ebfbL293

Issue #363, opened 2020-10-02
[FEAT] Export types


nklaasse:
It would make sense for Typescript projects to also export typings such as `CoreOptions`.

mrkosima:
As a workaround I'm using:

```
import type { CoreOptions } from '@adyen/adyen-web/dist/types/core/types';
```

But I'm agree that it would be much helpful to have exported types in root.

iLaurens:
I would like to see this as well. In my case I am using the custom card integration (secured fields) and I wanted to make it more robust by using TypeScript. I tried mapping function signatures manually for functions like `create`, `onChange`, `onError` etc but found it difficult to find the correct typings due to arbitrary naming and such. I could also imagine that these internals may change with patch releases while they would end up on the integrating side as breaking changes. If types get exported, it would be great if they are precise instead of using types like `any` or `object`.

WoodyWoodsta:
I'm looking for this as well.

Benjamindudok:
> As a workaround I'm using:
> 
> ```
> import type { CoreOptions } from '@adyen/adyen-web/dist/types/core/types';
> ```
> 
> But I'm agree that it would be much helpful to have exported types in root.

We used this workaround too. Though probably an update somewhere in our build system 'broke' this method because the types aren't in the 'exports' of the package. Which probably has something to do with [Typescript not completely adhering to ESM module specifications (yet)](https://github.com/microsoft/TypeScript/issues/46452). 

We're now getting the following error in our project with Vue 3 using Vite:

```[ERROR] [plugin vite:dep-scan] Missing "./dist/types/core" export in "@adyen/adyen-web" package```

Which is pretty much the same as this issue in the vite repository: https://github.com/vitejs/vite/issues/7749.

DanielRitvas:
Up.It would actually be very useful to be able to import types only via [npm](https://www.npmjs.com/search?q=%40types)! I'm using the library as an embedded script so there is no way to access types in the code at the moment.

Fifciu:
Up. I would love to import types easily

camil-adyen:
Hi all!

A BETA version of Web v6 is now available on Github. Feel free to take a look and give us feedback! We aim to release the official v6 in the coming weeks. More information can be found [here](https://github.com/Adyen/adyen-web/releases/tag/v6.0.0-beta.0).

PR #2294, opened 2023-08-11
feat: export types from Adyen Web

Fifciu:
## Summary

This PR simply exports every type from `types` directory to make it available for consumers. I haven't had chance to test it manually.

## Expected behavior

I am able to:
```ts
import type { CoreOptions } from "@adyen/adyen-web";
import type { DropinElementProps } from "@adyen/adyen-web";
```
Instead of:
```ts
import type { CoreOptions } from "@adyen/adyen-web/dist/types/core/types";
import type { DropinElementProps } from "@adyen/adyen-web/dist/types/components/Dropin/types";
```

**Fixed issue**:  #363

### DIFF
--- a/packages/lib/src/index.ts
+++ b/packages/lib/src/index.ts
@@ -15,3 +15,5 @@ async function AdyenCheckout(props: CoreOptions): Promise<Checkout> {
 }
 
 export default AdyenCheckout;
+
+export * from './types';


ribeiroguilherme:
Hey @Fifciu ,

Thanks for contributing to your repo. 

Unfortunately this won't work.  The exposure of the Typescript types is done during the build step which we use `tsc` to emit the types, which are placed in the `types` folder.  I know that currently it is a bit annoying to import the Type relying on  the folder structure, and ideally you would import from the root level.

The good news is that this change is something that we are aiming for the next major release v6, which should take place in the upcoming months. We don't have a strict deadline yet, but we are aiming for the end of this year. So keep an eye on the repo 👍 

Cheers!

Issue #2274, opened 2023-07-24
Disabled cardholder name and zip preventing payments on iOS Safari


ollym:
Test site:
https://cdn.checkout.ventrata.com/v3/staging/index.html

The video below will demonstrate the issue. You can create it using mac safari and enable responsive design mode (Develop > Enter Responsive Design Mode)

I included step by step through the checkout flow so you're able to get to the same page, make sure you click the "Waterpark" button on the test site, this is the only one where we have cardholder name and zip enabled.

The key element to look out for is that for some reason (and only on iOS Safari) when you finish entering the expiry date the cardholder name and zip inputs are automatically disabled (as I demonstrate using the elements inspector). I then manually remove this attribute to demonstrate it then works.

This is a serious production-level bug, we've had to disable cardholder name and zip code for now, but its exposing us to fraud. Please fix this ASAP. 

https://github.com/Adyen/adyen-web/assets/143771/83349ce5-0bf3-488c-9b37-6bcdc6c465fe

URL in case the video above doesn't load: https://share.cleanshot.com/zY8CxRY5

m1aw:
Hello @ollym,

Thank you for providing the detailed issue with the video; it was very useful for reproduction purposes. Let me start by addressing that the behavior you described only occurs in the responsive test environment on MacOS Safari. If you test it on a physical iOS device, you won't encounter the same problem.

To resolve this peculiar behaviour, you can pass the property `disableIOSArrowKeys: false` in the Card component/Dropin configuration.

This option disables a feature that works around an iOS behaviour, which was causing the user's input to jump away from the card details form when using the navigation arrows. The issue arises because the inputs for the card details (securedFields) exist within an iframe, and iOS doesn't recognise them as the "next" input to navigate to.

My recommendation is to test this behaviour to see which option better suits your users.

ollym:
@m1aw sorry that's not accurate. The issue is relayed to partial billing address. We since changed this to full on our test site but the video depicts partial where the issue still exists.

ollym:
@m1aw Sorry this doesn't work on a live device even with full billing address. Video from my personal iphone attached. And you can recreate with the link sent in my original post too.

https://github.com/Adyen/adyen-web/assets/143771/b3d1a958-3d88-4607-8a2f-39e03241288a

ollym:
After adding `disableIOSArrowKeys: false` it fixes the issue but i still think this should be resolved by default without a flag.

sponglord:
[...]

PR #2299, opened 2023-08-14
Chore/handling safari responsive design mode

sponglord:
## Summary
Safari in _Responsive Design Mode_ creates its own unique environment where some features prevented by the Safari security model are allowed - namely setting focus on 3rd party iframes.
This allows our feature to _disable arrow keys on the iOS soft keyboard_ to be partly implemented, but without having the actual touch events that this fixes relies on, in order to fully work

This has lead to some strange behaviour, noted by merchants (e.g. issue #2274), whereby a click on a securedField’s label can lead to that input gaining focus (due to setting focus on 3rd party iframes being allowed) whilst the holderName (or zip code field) field becomes disabled, with no means to enable it again (touch events not being simulated)

This PR fixes this issue by preventing our _disable arrow keys on the iOS soft keyboard_ from working at all, unless touch events are being registered. We look to detect these events occurring both within CheckoutWeb & SecuredFields by adding a boolean flag within the respective `touchstart` handlers

Also in this PR are some extra checks around handling the `expiryYear` in an autofill scenario

## Tested scenarios
- Tested in both Safari (responsive design mode) _and_  both on an actual & a simulated iOS device (BrowserStack) to confirm that the expected behaviour around focusing/disabling inputs is in place in all scenarios
- All unit & e2e tests pass

**Fixed issue**:  #2274


packages/lib/src/components/internal/SecuredFields/lib/CSF/extensions/createSecuredFields.ts
```diff
@@ -245,13 +245,27 @@ export function setupSecuredField(pItem: HTMLElement, cvcPolicy?: CVCPolicyType,
...
+                    /**
+                     * re. this.hasGenuineTouchEvents...
+                     *  There seems to be an issue with Responsive Design mode in Safari that means it allows setting focus on cross-origin iframes,
+                     *  without enabling the touch events that allow the "disableIOSArrowKeys" workaround to fully function.
+                     *  This results in a click on an securedFields *label* leading to, for example, the holderName field being disabled, but w/o access
+                     *  to the touch events that would let it re-enable itself.
+                     *
+                     *  So we prevent the "disableIOSArrowKeys" workaround unless we genuinely have touch events available.
+                     */
+                    if (this.hasGenuineTouchEvents || pFeedbackObj.hasGenuineTouchEvents) {
```
longyulongyu, line 258: What's the difference between `hasGenuineTouchEvents` in `this` vs in `pFeedbackObj`?

packages/lib/src/components/internal/SecuredFields/lib/CSF/partials/processAutoComplete.ts
```diff
@@ -12,7 +12,7 @@ import getIframeContentWin from '../utils/iframes/getIframeContentWin';
  *
  * @param pFeedbackObj -
  */
-export function processAutoComplete({ csfState, csfConfig, csfCallbacks }, pFeedbackObj: SFFeedbackObj): void {
+export function processAutoComplete({ csfState, csfConfig, csfCallbacks }, pFeedbackObj: SFFeedbackObj): boolean {
```
longyulongyu, line 15: Maybe rename the function as well, such as `shouldProcessAutoComplete` or similar? Because the returned value is `boolean` now, the function name doesn't reflect it.

longyulongyu (approved):

packages/lib/src/components/internal/SecuredFields/lib/CSF/extensions/createSecuredFields.ts
sponglord: The difference is where they've originated. `hasGenuineTouchEvents` here, has originated from a _touch_ in checkoutWeb e.g. on a `label` element whereas `pFeedbackObj.hasGenuineTouchEvents` has originated from a _touch_ directly into an input in the securedFields iframe. We don't know exactly where the "confirmation" that we have touch events is coming from, therefore we have to check for both.

packages/lib/src/components/internal/SecuredFields/lib/CSF/partials/processAutoComplete.ts
sponglord: Returning a boolean is less to do with whether we process the event or not, and more to do with returning something that a unit test (still to be written) can inspect

longyulongyu: OK, then leave it for now :). 
It's more about giving function a name that describes what is does and align with the returned type. So if the return type is void, the function name is more like 'do' something. If the returned type is a boolean, the function name is more like 'should', 'is'.

PR #2300, opened 2023-08-16
initial implementation of DA

davordavidovic:
## Summary
initial implementation of delegated authentication:
- handle enrollment scenario after successful 3ds2 challenge (via `DelegatedAuthenticationEnrollment.tsx` component)
- introduce new `delegatedAuthentication` action type
- handle followup (authentication) scenario `delegatedAuthentication` action type (via `DelegatedAuthentication.tsx` component)
- submit delegated authentication result to new endpoint (`/v1/submitDelegatedAuthenticationResult`) and handle followup action

### DIFF
--- a/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
+++ b/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
@@ -1,5 +1,6 @@
 import { Component, h } from 'preact';
 import DoChallenge3DS2 from './DoChallenge3DS2';
+import DelegatedAuthenticationEnrollment from '../DelegatedAuthentication/DelegatedAuthenticationEnrollment';
 import { createChallengeResolveData, handleErrorCode, prepareChallengeData, createOldChallengeResolveData } from '../utils';
 import { PrepareChallenge3DS2Props, PrepareChallenge3DS2State } from './types';
 import { ChallengeData, ThreeDS2FlowObject } from '../../types';
@@ -60,10 +61,15 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
              *  This is different for the 'old',v66, flow triggered by a 'threeDS2Challenge' action (which includes the threeds2InMDFlow)
              *  than for the new, v67, 'threeDS2' action
              */
-            const resolveDataFunction = this.props.useOriginalFlow ? createOldChallengeResolveData : createChallengeResolveData;
-            const data = resolveDataFunction(this.props.dataKey, resultObj.transStatus, this.props.paymentData);
 
-            this.props.onComplete(data); // (equals onAdditionalDetails - except for 3DS2InMDFlow)
+            if (this.state.challengeData.delegatedAuthenticationSDKInput && resultObj.transStatus === 'Y') {
+                this.setState({ status: 'delegatedAuthenticationEnrollment' });
+            } else {
+                const resolveDataFunction = this.props.useOriginalFlow ? createOldChallengeResolveData : createChallengeResolveData;
+                const data = resolveDataFunction(this.props.dataKey, resultObj.transStatus, this.props.paymentData);
+
+                this.props.onComplete(data); // (equals onAdditionalDetails - except for 3DS2InMDFlow)
+            }
         });
     }
 
@@ -133,6 +139,18 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
                         {this.state.errorInfo ? this.state.errorInfo : this.props.i18n.get('error.message.unknown')}
                     </div>
                 </div>
+            );
+        }
+
+        if (this.state.status === 'delegatedAuthenticationEnrollment') {
+            return (
+                <DelegatedAuthenticationEnrollment
+                    dataKey={this.props.dataKey}
+                    token={this.state.challengeData.delegatedAuthenticationSDKInput}
+                    authorisationToken={this.props.paymentData}
+                    onComplete={this.props.onComplete}
+                    useOriginalFlow={this.props.useOriginalFlow}
+                />
             );
         }
 
--- a/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
+++ b/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
@@ -10,6 +10,13 @@ import './challenge.scss';
 import { hasOwnProperty } from '../../../../utils/hasOwnProperty';
 import useImage from '../../../../core/Context/useImage';
 
+const enum ChallengeStatus {
+    complete = 'complete',
+    retrievingChallengeToken = 'retrievingChallengeToken',
+    delegatedAuthenticationEnrollment = 'delegatedAuthenticationEnrollment',
+    error = 'error'
+}
+
 class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareChallenge3DS2State> {
     public static defaultProps = {
         onComplete: () => {},
@@ -43,7 +50,7 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
 
             /** All good */
             this.state = {
-                status: 'retrievingChallengeToken',
+                status: ChallengeStatus.retrievingChallengeToken,
                 challengeData,
                 errorInfo: null
             };
@@ -55,7 +62,7 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
     }
 
     setStatusComplete(resultObj) {
-        this.setState({ status: 'complete' }, () => {
+        this.setState({ status: ChallengeStatus.complete }, () => {
             /**
              * Create the data in the way that the /details endpoint expects.
              *  This is different for the 'old',v66, flow triggered by a 'threeDS2Challenge' action (which includes the threeds2InMDFlow)
@@ -63,7 +70,7 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
              */
 
             if (this.state.challengeData.delegatedAuthenticationSDKInput && resultObj.transStatus === 'Y') {
-                this.setState({ status: 'delegatedAuthenticationEnrollment' });
+                this.setState({ status: ChallengeStatus.delegatedAuthenticationEnrollment });
             } else {
                 const resolveDataFunction = this.props.useOriginalFlow ? createOldChallengeResolveData : createChallengeResolveData;
                 const data = resolveDataFunction(this.props.dataKey, resultObj.transStatus, this.props.paymentData);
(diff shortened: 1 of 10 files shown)


packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
```diff
@@ -60,10 +61,15 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
              *  This is different for the 'old',v66, flow triggered by a 'threeDS2Challenge' action (which includes the threeds2InMDFlow)
              *  than for the new, v67, 'threeDS2' action
              */
-            const resolveDataFunction = this.props.useOriginalFlow ? createOldChallengeResolveData : createChallengeResolveData;
-            const data = resolveDataFunction(this.props.dataKey, resultObj.transStatus, this.props.paymentData);
 
-            this.props.onComplete(data); // (equals onAdditionalDetails - except for 3DS2InMDFlow)
+            if (this.state.challengeData.delegatedAuthenticationSDKInput && resultObj.transStatus === 'Y') {
+                this.setState({ status: 'delegatedAuthenticationEnrollment' });
```
sarobrien, line 66: Since there are a few different status' it would be cleaner to create an enum rather than relying on strings.

packages/lib/src/components/ThreeDS2/components/utils.ts
```diff
@@ -207,3 +237,11 @@ export const get3DS2FlowProps = (actionSubtype, props) => {
         i18n: props.i18n
     };
 };
+
+export const isInIframe = () => {
+    try {
```
sarobrien, line 242: There is no need for a try/catch here, just need to check window is defined and then return the boolean expression.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+    authorisationToken,
+    onComplete
+}: DelegatedAuthenticationEnrollmentProps) => {
+    const [status, setStatus] = useState<string>('loading');
+    const [paymentCredential, setPaymentCredential] = useState<PublicKeyCredential>(null);
+    const [registrationData, setRegistrationData] = useState<DelegatedAuthenticationEnrollmentInputData>(null);
+    const [supported, setSupported] = useState<boolean>();
+    const [spcSupported, setSpcSupported] = useState<boolean>();
+
+    const inIframe = useMemo(() => {
```
sarobrien, line 26: This function doesn't need to memoized, it's not doing a heavy calculation just returning a boolean.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+import Button from '../../../internal/Button';
+
+const DelegatedAuthenticationEnrollment = ({
+    token,
+    useOriginalFlow,
+    dataKey,
+    authorisationToken,
+    onComplete
+}: DelegatedAuthenticationEnrollmentProps) => {
+    const [status, setStatus] = useState<string>('loading');
```
sarobrien, line 20: Here also I would use an enum for status rather than strings.

packages/lib/src/components/ThreeDS2/components/utils.ts
```diff
@@ -131,20 +139,42 @@ export const createOldFingerprintResolveData = (dataKey: string, resultObj: Resu
...
 });
 
 // Needed for old 3DS2 flow & threeds2InMDFlow
-export const createOldChallengeResolveData = (dataKey: string, transStatus: string, authorisationToken: string): any => ({
+export const createOldChallengeResolveData = (
+    dataKey: string,
+    transStatus: string,
+    authorisationToken: string,
+    delegatedAuthenticationSDKOutput?: string
+): any => ({
```
sarobrien, line 159: This should also have a return type rather than any.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/utils.ts
```diff
@@ -0,0 +1,8 @@
+export const decodeBase64 = (value: string) => {
```
sarobrien, line 1: Why are these in a different utils file when there already is a utils file?

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+    const [spcSupported, setSpcSupported] = useState<boolean>();
+
+    const inIframe = useMemo(() => {
+        return isInIframe();
+    }, []);
+
+    const complete = (sdkOutput: string) => {
+        setStatus('loading');
+        const resolveDataFunction = useOriginalFlow ? createOldChallengeResolveData : createChallengeResolveData;
+        const data = resolveDataFunction(dataKey, 'Y', authorisationToken, sdkOutput);
```
sarobrien, line 33: It's not good practice to pass strings like this, better to extract it to a constant and name it.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+        const sdkOutput = encodeDelegatedAuthenticationEnrollmentOutputData({ errorCode });
+        complete(sdkOutput);
+    };
+
+    const completeWithErrorDescription = (errorDescription: string) => {
+        const sdkOutput = encodeDelegatedAuthenticationEnrollmentOutputData({ errorDescription });
+        complete(sdkOutput);
+    };
+
+    useEffect(() => {
```
sarobrien, line 47: This could be a extracted into a hook

sarobrien: Or the async function could be a util, that way it's testable

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+        return JSON.parse(decodedEnrollmentDataString);
+    };
+
+    const isSpcSupported = async () => {
+        if (!window.PaymentRequest) {
+            return false;
+        }
+
+        try {
+            return await mockPaymentRequest().canMakePayment();
```
sarobrien, line 119: Does this mean everytime a user does DA a mock payment request is made first to determine if the payment can be made?

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+        } catch (e) {
+            console.error(e);
+            return false;
+        }
+    };
+
+    const mockPaymentRequest = () => {
```
sarobrien, line 126: This should be a util.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthenticationEnrollment.tsx
```diff
@@ -0,0 +1,260 @@
...
+        );
+    };
+
+    const setCredentialCookie = (credentialId: string) => {
+        const expiryDate = new Date();
+        expiryDate.setFullYear(expiryDate.getFullYear() + 100); // expire in 100 years
+        document.cookie = `${credentialId}=; path=/; expires=${expiryDate.toUTCString()};`;
+    };
+
+    const createPaymentCredential = async (registrationData: DelegatedAuthenticationEnrollmentInputData): Promise<PublicKeyCredential> => {
```
sarobrien, line 158: This could be a util, then it is independently testable.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthentication.tsx
```diff
@@ -0,0 +1,331 @@
...
+    challengeWindowSize
+}: DelegatedAuthenticationProps) => {
+    const [loading, setLoading] = useState<boolean>(true);
+    const [authenticationData, setAuthenticationData] = useState<DelegatedAuthenticationInputData>(null);
+    const [hasCookieForCredential, setHasCookieForCredential] = useState<boolean>();
+
+    // todo: this should be a part of the token data
+    const useCookiesInSpc = true;
+
+    const shouldDoSpc = useMemo(() => {
```
sarobrien, line 42: Same here, useMemo is not required.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthentication.tsx
```diff
@@ -0,0 +1,331 @@
...
+                    complete({ errorCode: DelegatedAuthenticationError.NoInputDataPresent });
+                }
+                setAuthenticationData(data);
+            } catch (e) {
+                complete({ errorCode: DelegatedAuthenticationError.Unknown, errorDescription: e.message });
+            }
+        } else {
+            complete({ errorCode: DelegatedAuthenticationError.NoInputDataPresent });
+        }
+    }, []);
```
sarobrien, line 60: This useEffect depends on token then it should have token in it's dependency array.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthentication.tsx
```diff
@@ -0,0 +1,331 @@
...
+        if (!cookieForCredential) {
+            console.error('No cookie matching credential');
+            complete({ errorCode: DelegatedAuthenticationError.NoBrowserCookie });
+            return;
+        }
+        setHasCookieForCredential(cookieForCredential);
+    }, [authenticationData]);
+
+    useEffect(() => {
+        if (!authenticationData) {
```
sarobrien, line 83: This useEffect depends on hasCookieForCrediential but doesn't call it anywhere? Should the if perhaps be (!authenticationData && hasCookieForCredential) otherwise every time hasCookieForCrediential gets updated this will run regardless of it's value.

sarobrien (changes requested):
Review still in progress but added some comments.

sarobrien (changes requested):
There is a lot of logic in the components that could be extracted to utils or hooks. This will make the code easier to test and keep the components clean.

packages/lib/src/components/ThreeDS2/components/DelegatedAuthentication/DelegatedAuthentication.tsx
```diff
@@ -0,0 +1,331 @@
...
+    // todo: this should be a part of the token data
+    const useCookiesInSpc = true;
+
+    const shouldDoSpc = useMemo(() => {
+        return isInIframe();
+    }, []);
+
+    useEffect(() => {
+        if (token) {
+            try {
```
sarobrien, line 48: Why are you using a try/catch here?

packages/lib/src/components/ThreeDS2/components/utils.ts
sarobrien: As discussed over mattermost, window.top can actually throw an exception, this is a security feature and it will throw an exception if the origin is not the same as parent, however that means this code will always return false. So it should return true in the catch I think. but in general I wonder if there is a cleaner way to do this..🤔

Issue #2309, opened 2023-08-23
It is not possible to select bank option from the list


tatsiak:
**Describe the bug**

**To Reproduce**
1. Go to dropin with EPS enabled
2. Click on EPS payment method
3. Click on bank select
4. Click on any option
5. Observe that click is not affecting select in any way

**Expected behavior**
The option is selected

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser chrome
 - Version 116.0.5845.96

**Additional context**
The issue appears with dropin version 4.9

tatsiak:
the issue is caused by the fact that `data-disabled` equals "false"
which is later in code interpreted as boolean `true`

![image](https://github.com/Adyen/adyen-web/assets/17749946/2f7bcf30-1ec3-4a2d-802f-b9d073540e2a)

tatsiak:
Here is the line that causes the issue for 4.9 version
https://github.com/Adyen/adyen-web/blob/613457b7982bdec2b21973003a187e01aa002f0f/packages/lib/src/components/internal/FormFields/Select/Select.tsx#L56

ribeiroguilherme:
Hi @tatsiak .

Thanks for reaching out. We fixed couple of bugs since 4.9, which few of them are related to the bank list. I recommend you to update the library to the most recent version and try it out. Feel free to reach out if the issue persists on the most recent version.

Cheers

PR #2310, opened 2023-08-23
Allow to select item when data-disabled equals to "false"

tatsiak:
## Summary
For some reason `data-disable` attribute is stored as a string.
This is why it is not possible to select any item, since each of them has `data-disabled="false"`

image

In order to prevent this issue from happening I handled both boolean and string.

**Fixed issue**:  [2309](https://github.com/Adyen/adyen-web/issues/2309)


tatsiak:
@kepejas, yes there is a dirty solution - just remove that attribute.

something like this:
```
useEffect(() => {
    setInterval(() => {
      try {
        const list = document.querySelectorAll('#adyen [data-disabled="false"]')
        list.forEach((el) => el.removeAttribute('data-disabled'))
      } catch (e) {
        console.error('Error encountered:', e)
      }
    }, 1000)
  }, [])
```
hope it helps.

sponglord:
I believe this problem was resolved with the answer in #2309 (i.e. we have fixed this issue in a later release)

PR #2311, opened 2023-08-23
Allow to select item when data-disabled equals to "false"

tatsiak:
## Summary
For some reason `data-disable` attribute is stored as a string.
This is why it is not possible to select any item, since each of them has `data-disabled="false"`

image

In order to prevent this issue from happening I handled both boolean and string.

**Fixed issue**:  [2309](https://github.com/Adyen/adyen-web/issues/2309)


kepejas:
Was this merged to v5?

PR #2312, opened 2023-08-23
feature/add issuerlist feature to highlight all issuers and hide dropdown

mohammedmulazada:
## Summary

Hello! This is my first PR here, at my company we are going to slowly make use of the awesome drop-in payment method, so after setting drop-in up in my own codebase I also wanted to set it up locally here to have a look. I saw a scenario where you want all issuers to be highlighted, this is already technically possible by passing all issuer id's to the highlightedIssuers prop, but then the dropdown is also still there which is too much for this use case.

I want to propose this feature where all issuers will be highlighted by enabling the prop and it also hiding the dropdown and content separator.

Screenshot 2023-08-23 at 12 56 56

## Tested scenarios
I added an unit test to test the scenario where the highlightAllIssuersAndHideDropdown is set to true, the name is verbose but it also covers that the dropdown will also be hidden, which is new.

As mentioned, it's my first PR, so if things need to change please let me know :)

### DIFF
--- a/.changeset/khaki-impalas-sin.md
+++ b/.changeset/khaki-impalas-sin.md
@@ -1,0 +1,5 @@
+---
+"@adyen/adyen-web": minor
+---
+
+Add issuerlist feature to highlight all issuers and hide dropdown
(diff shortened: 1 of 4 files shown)


mohammedmulazada:
Hello @ArjenLa,

Thanks for the kudos :) I appreciate hearing the rationale (especially given your position in the Checkout aspect) behind why this was not implemented initially, I also had not really considered the scenario where the list would be too long and problematic on mobile. So with that information I definitely agree that this will not delivery a lot of value. I will definitely keep making suggestions if/when they come up!


Thanks again :)

Issue #2134, opened 2023-05-04
Types not resolved when using `moduleResolution: NodeNext` in tsconfig


sachinahya:
**Describe the bug**

**To Reproduce**
1. Import the `@adyen/adyen-web` module in a TypeScript project that has `moduleResolution` set to `Node16`, `NodeNext` or `bundler`.
2. See error explaining the declaration file could not be found

```
Could not find a declaration file for module '@adyen/adyen-web'. '/Users/sachin/code/work/test/site/node_modules/@adyen/adyen-web/dist/es/index.js' implicitly has an 'any' type.
  There are types at '/Users/sachin/code/work/test/site/node_modules/@adyen/adyen-web/dist/types/index.d.ts', but this result could not be resolved when respecting package.json "exports". The '@adyen/adyen-web' library may need to update its package.json or typings.ts(7016)
```

Reproduction - notice that changing `moduleResolution` to `node` causes the error to go away:
https://stackblitz.com/edit/typescript-yur3dk?file=index.ts

**Expected behavior**
The included types should be resolved by TypeScript and there are no type errors.

ribeiroguilherme:
Hi @sachinahya ,

I'd like to understand a more which kind of web app you are working with.  What are the frameworks involved, and why do you need to use `NodeNext` or `Node16` as moduleResolution for your web app?

sachinahya:
Hi, we need to use `Node16` as the `moduleResolution` because we're using `"type": "module"` in our package.json for ESM support. A number of third-party dependencies we are using are going ESM-only so we're updating our app to support them now that ESM support in Node has been stable for a while.

From what I understand, TypeScript aims to follow Node's resolution algorithm in this mode and only looks for types defined via the `exports` field. Since these are currently not specified, it fails to resolve the type declarations.

FBisca:
I had the same problem.

The `"moduleResolution": "bundle"` also comes by default with NextJS 13 which might cause this problem to be more common as well.

avez-fsd:
please help with this issue.

ribeiroguilherme:
@sachinahya can you test again with the latest released version? We pushed a change that might fix this

rehoutm:
I have a different issues, but I believe it is realted to the same root cause.

In our project, we are using ESM exclusively, with typescript resolvers set to `nodenext`. With this setup, importing adyen like this:

```
import AdyenCheckout from '@adyen/adyen-web'

// tsc will argue about this:

[...]

PR #2313, opened 2023-08-24
chore: fix ts config

janhellar:
## Summary

The `types` field in `package.json` is no longer used if there is an `exports` field. Because of that `tsc` complains that it can not find a declaration file for the module.

### DIFF
--- a/.changeset/brown-eels-hear.md
+++ b/.changeset/brown-eels-hear.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+chore: fix ts config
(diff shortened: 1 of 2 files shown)


janhellar:
Hi, can anybody look at this PR? Is the change correct? Not sure about those 2 failing checks, but probably not related to my changes? E2E seems to be canceled. Unit tests failed because of missing SONAR_TOKEN - could be because pull request from fork?

m1aw:
Hey @janhellar thank you for the PR. This seems correct but I haven't add the time to actually test it. In the mean time we are aware that the tests don't reflect the PR and it's related to the way we load secrets.

ribeiroguilherme (approved):

ribeiroguilherme:
Related to #2134

Issue #2316, opened 2023-08-28
3D Secure redirection fails when the application is loaded in an iframe


anooparveti:
**Describe the bug**
Our application is loaded inside an `iframe`. After adding the Credit Card details (India), we perform `checkout.createFromAction().mount()` which is supposed to redirect user on to URL `https://checkoutshopper-live-in.adyen.com/checkoutshopper/threeDS/checkoutRedirect/xxx`.  We have the error in console (Refer attached screenshots).

The issue occurs when Adyen component tries to redirect the user to URL using `window.location.assign()`.

P.S: Doing this redirection on the top most window would solve our issue. i.e. `window.top.location.assign()`

We dont reproduce the issue when we load the application outside of an iframe.

**Expected behavior**
Redirection to the requested URL must happen without any error even if our application is loaded inside iframe

OR

Provide callback from Adyen component so that we can redirect user to the required URL outside of our `iframe`

**Screenshots**

<img width="1791" alt="ConsoleError" src="https://github.com/Adyen/adyen-web/assets/22835814/813bf6ea-f91b-4906-9ca6-bef98d36ac4d">

sponglord:
@anooparveti  - I'm not currently able to recreate this exact issue.

I've tried running Dropin in an iframe to test 2 other "redirect" scenarios, and in both cases: 
- a 3DS2, MDflow, redirect
- _and_ a regular credit card redirect
everything works as expected.

Because the Dropin is in an iframe - the call to `window.location.assign()` that you mention, works within the context of the Dropin/iframe, and the content of the iframe is correctly replaced with the content from the redirect url.

I've asked our Authentication team if it's possible to recreate the original, 3DS2 redirect flow (that is now only available, afaik, in countries like India & Sri Lanka) - but since this also uses the same `Redirect` component as the other 2 test cases - I would expect it to also work.

sponglord:
@anooparveti the url in the screenshot: https://ca.ovh.com is this your url?

anooparveti:
> @anooparveti - I'm not currently able to recreate this exact issue.
> 
> I've tried running Dropin in an iframe to test 2 other "redirect" scenarios, and in both cases:
> 
> * a 3DS2, MDflow, redirect
> * _and_ a regular credit card redirect
> 
> everything works as expected.
> 
[...]

PR #2325, opened 2023-09-11
Add 'redirectFromTop' config prop to allow top level redirect when Checkout loaded in an iframe

sponglord:
## Summary
If the merchant is running their site within an iframe*, they can now specify a `redirectFromTopWhenInIframe ` boolean in the Checkout config. 
Then, if we end up with a `Redirect` action, we check for ourselves if the site is running in an iframe and, if it is, and `redirectFromTopWhenInIframe ` is set to `true`, then we will attempt to perform the redirect on the top level, parent, window rather than the window holding the iframe

**\*Caveat**:  We don't recommend the use of iframes for hosting Checkout. We cannot guarantee that all payment methods will work when inside an iframe (ApplePay, GooglePay etc). Nor can we guarantee that pages outside of our control e.g. Issuer created pages that are part of the 3DS2 process, will perform correctly.

## Tested scenarios
Manually tested that both `form` submits and `window.location.assign` happen on the top level window when checkout hosted in an iframe and `redirectFromTopWhenInIframe ` is set to true.
New unit test added.

**Fixed issue**:  #2316

### DIFF
--- a/packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
+++ b/packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
@@ -1,10 +1,12 @@
 import { Component, h } from 'preact';
+import detectInIframe from '../../../../utils/detectInIframe';
 
 interface RedirectShopperProps {
     beforeRedirect: (resolve, reject, url) => Promise<void>;
     url: string;
     method: 'GET' | 'POST';
     data?: any;
+    redirectFromTop?: boolean;
 }
 
 class RedirectShopper extends Component<RedirectShopperProps> {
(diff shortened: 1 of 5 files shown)


packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
```diff
@@ -1,10 +1,12 @@
 import { Component, h } from 'preact';
+import detectInIframe from '../../../../utils/detectInIframe';
 
 interface RedirectShopperProps {
     beforeRedirect: (resolve, reject, url) => Promise<void>;
     url: string;
     method: 'GET' | 'POST';
     data?: any;
+    redirectFromTop?: boolean;
```
longyulongyu, line 9: I would rename the property, such as `isInsideIframe`, because to me the purpose is not clear from the name `redirectFromTop`.

longyulongyu (approved):

PR #2334, opened 2023-09-20
fix: Clear timeouts on SecuredFieldsProvider when unmounting the component

kaakock:
## Summary
SecuredFieldsProvider uses timeouts to check whether SecureFields are loaded properly. Those timeouts are not cleared when the component unmounts which causes issues if the fields have already been unmounted. This triggers the onError handler in cases where it should not.

## Tested scenarios
Mounting and almost immediately unmounting adyen drop-in triggers the onError handler.

### DIFF
--- a/.changeset/smooth-jars-buy.md
+++ b/.changeset/smooth-jars-buy.md
@@ -1,0 +1,5 @@
+---
+"@adyen/adyen-web": patch
+---
+
+fix: Clear timeouts on SecuredFieldsProvider when unmounting the component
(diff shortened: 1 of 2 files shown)


sponglord:
Hi @kaakock . Thanks for the PR! Can you tell me is it related to this issue (#2335) - or is this just a coincidence? :-)

sponglord:
In which case this is a very timely PR! Thank you :-)

sponglord (approved):

Issue #2337, opened 2023-09-26
Feat: Have an option to make the default storePaymentMethod to true


Ganthology:
**Is your feature request related to a problem? Please describe.**
My team is looking to build an `opt out system` in our checkout page using the `DropIn` component.
Scenario: 
When customer comes in, the checkbox is ticked, they have to unticked the checkbox if they want to opt out from storing their payment details.

**Describe the solution you'd like**
Able to pass in a prop to make `storePaymentMethod` default to `true`

**Additional context**
The current `storePaymentMethod` state is initialised to `false`
![image](https://github.com/Adyen/adyen-web/assets/18641991/48fa3e03-ef51-44fe-ba4e-cdc0f1437ddb)

Ganthology:
I created a PR attempting to fix this issue https://github.com/Adyen/adyen-web/pull/2338. Please enlighten me on how I can proceed from here, thanks!

PR #2338, opened 2023-09-26
Feat/add default value props for card store payment method

Ganthology:
## Summary

There's now a prop for CardInput component to set default value for `storePaymentMethod`

## Tested scenarios

Please advice on how I can test locally, thanks! 
It's a UI change

**Fixed issue**:  
#2337


Shaunmak1214:
Good initiative! @Ganthology

sponglord:
Hi @Ganthology 
Thank you for the PR. Unfortunately for regulatory reasons, we cannot accept it.

The rules, as [stated by Visa](https://usa.visa.com/content/dam/VCOM/download/about-visa/visa-rules-public.pdf) say that
> Before a Merchant or  DWO either stores a Payment Credential for a future Transaction or completes an Advance Payment or Partial Payment, it must obtain the Cardholder’s **express informed consent**...

We take it that "express informed consent" means that the shopper must deliberately & consciously click this checkbox,
and that it can not be presented to them "pre-checked", where they might not notice it; and end up having their card stored inadvertently.

We hope you understand our reasoning for rejecting this PR, thanks

PR #2348, opened 2023-10-09
feat: new design for the dropin

longyulongyu:
## Summary
New design for the drop-in component:
- separation for the stored and the regular payment methods
- each payment item has rounded corners and there is 16px space between each other
- express payment methods sit side by side

## Tested scenarios
- Fixed failed tests. 
- Modify the dropin config and go to Safari to check express payment list
```javascript
    const dropin = checkout
        .create('dropin', {        
            instantPaymentTypes: ['googlepay', 'applepay']
        })
        .mount('#dropin-container');
```


sponglord:
It's not entirely clear to me which variables that I, as a merchant, should be overwriting...
Is it just the ones that are commented in the blocks: `/*Drop-in*/`, /* Input */ etc?
This is perhaps why a styling guide or "How to..." doc would be a useful thing to have

longyulongyu:
> It's not entirely clear to me which variables that I, as a merchant, should be overwriting... Is it just the ones that are commented in the blocks: `/*Drop-in*/`, `/*Input*/` etc? This is perhaps why a styling guide or "How to..." doc would be a useful thing to have

I agree, I will work on a doc to explain the purpose of each component level variable, or perhaps an interactive story.

Issue #2320, opened 2023-08-30
Web-Dropin Configuration lacks typescript definitions for Sofort


gecko-steffen:
**Describe the bug**
When adding information to the config for the AdyenCheckout instance, `directEbanking` is not a recognized property of `paymentMethodsConfiguration`.

The official documentation also confirms this: https://docs.adyen.com/payment-methods/sofort/web-drop-in/

However, it should be possible to provide a `name` property to configure what the payment drop-in displays on the checkout page.

Ignoring the typing-inconsistencies, manually specifying e.g.

```js
paymentMethodsConfiguration: {
  directEbanking: {
    name: 'Sofort (by Klarna)'
  }
}
```
does indeed work and correctly changes the display name of the button in the drop-in.

**Expected behavior**
The typing should support all payment methods and provide the correct typing information.

**Screenshots**
![image](https://github.com/Adyen/adyen-web/assets/95866823/9f576d50-19ba-40c9-bb59-92d8e763e8ed)

**Additional context**
I fixed compilation errors locally by explicitly overriding the type:

```ts
const config: CoreOptions & {
  paymentMethodsConfiguration: PaymentMethodsConfiguration & {
    directEbanking: { name: string };
  };
}
```

ribeiroguilherme:
Hey @gecko-steffen ,

Thanks for reporting it. We fix this issue on the 5.53.0 release. Please reach out to us if the problem persists. Cheers!

PR #2349, opened 2023-10-10
Adjusting paymentMethodConfiguration type to support txVariants not mapped internally

ribeiroguilherme:
## Summary
`paymentMethodConfiguration` typescript definition does not support tx variants that are not mapped internally. If they are set, the linter will complain. This PR fixes this problem.

## Tested scenarios
- Checked that following code does not trigger TS complains:

```js
    directEbanking: {
        name: 'Sofort Klarna'
    },
}
```

**Fixed issue**:  #2320


packages/lib/src/core/types.ts
```diff
@@ -172,4 +172,7 @@ export type PaymentMethodsConfiguration =
       }
     | {
           [key in PaymentActionsType]?: any;
+      }
+    | {
+          [key: string]: any;
```
sponglord, line 177: Can we make this `any` more specific?

sponglord (approved):

longyulongyu (approved):

PR #2354, opened 2023-10-12
Populate Sepa data with initial values

sponglord:
## Summary
Populate data with initial values (empty strings) for `ibanNumber` and `ownerName`

## Tested scenarios
Trying to access `component.data` before the component has been interacted with no longer causes an error

**Fixed issue**:  COWEB-1113

### DIFF
--- a/packages/lib/src/components/Sepa/Sepa.tsx
+++ b/packages/lib/src/components/Sepa/Sepa.tsx
@@ -15,6 +15,11 @@ class SepaElement extends UIElement {
         showFormInstruction: true
     };
 
+    constructor(props) {
+        super(props);
+        this.state = { ...this.state, ...{ data: { ibanNumber: '', ownerName: '' } } };
+    }
+
     /**
      * Formats props on construction time
      */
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/Sepa/Sepa.tsx
```diff
@@ -15,6 +15,11 @@ class SepaElement extends UIElement {
         showFormInstruction: true
     };
 
+    constructor(props) {
+        super(props);
+        this.state = { ...this.state, ...{ data: { ibanNumber: '', ownerName: '' } } };
```
longyulongyu, line 20: I haven't tested it locally, but isn't the state already initialized in the `IbanInput.tsx` constructor?

sponglord: That's the problem though. It is, (already initialized in the `IbanInput.tsx` constructor) but it doesn't propagate until the shopper interacts with the component

longyulongyu: I see, it looks like the `IbanInput.tsx` only propagates to `Sepa.tsx` when `if (this.state.data['ibanNumber'] || this.state.data['ownerName'])`. But i think let the `Sepa.tsx` decides when to propagate is better, with less responsibility for the shared component - `IbanInput.tsx`.

longyulongyu (approved):

PR #2357, opened 2023-10-16
Supports 3DS method notification URL

othree:
## Summary

I found that the `threeDSMethodNotificationURL` of the 3DS FingerPrint is implemented. But the prop is not exposed.
This means it is impossible to assign it in the API-only flow. 

The PR exposed the prop `notificationURL` of 3DS FingerPrint Component.

### DIFF
--- a/packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
+++ b/packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
@@ -67,6 +67,7 @@ const actionTypes = {
             _parentInstance: props._parentInstance,
             paymentMethodType: props.paymentMethodType,
             challengeWindowSize: props.challengeWindowSize, // always pass challengeWindowSize in case it's been set directly in the handleAction config object
+            notificationURL: props.notificationURL,
 
             // Props unique to a particular flow
             ...get3DS2FlowProps(action.subtype, props)


sponglord:
Hi @othree 
Can you please say what your use case is for this PR?

I've run the idea past our Authentication team and they are of the opinion that there is no solid reason for merchants to actively set their own URLs. That it would just make the integration more complicated and error prone.

We have made a point of providing these URLs for merchants precisely so they don't have to - we handle the response from the 3DS2 fingerprinting/challenge on their behalf and perform the necessary steps to make sure the results are correctly handled and passed back to the SDK in a way that it understands, so that the process can continue correctly.

othree:
@sponglord Thanks for your reply.

We have an API-only integration already. And now we are working on the 3DS and plan to use the adyen-web to do the identifyShoppe and challengeShopper.

We follow the doc: https://docs.adyen.com/checkout-build-your-own-3ds2/#get-the-3d-secure-2-device-fingerprint
We found this notification URL and are wondering if it will help our integration. So not a scenario requires this feature.

sponglord:
Hi @othree - that is some _very_ old documentation you found there! :-)

The process is much, much easier when you just want to use `adyen-web` to do the `identifyShopper` and `challengeShopper` steps: https://docs.adyen.com/online-payments/3d-secure/native-3ds2/web-component/?tab=create-new-component_2

PR #2390, opened 2023-10-25
Chore/v6 getting e2e tests running

sponglord:
## Summary
Fixing or replacing (with Playwright equivalent) e2e tests to run against v6 code

## Tested scenarios
All e2e tests passing


packages/e2e-playwright/models/card.ts
```diff
@@ -90,13 +120,34 @@ class Card {
...
         await this.cvcInput.type(cvc, { delay: USER_TYPE_DELAY });
     }
+
+    async pressKeyboardToNextItem() {
+        await this.page.keyboard.press('ArrowDown', { delay: KEYBOARD_DELAY });
+    }
+
+    async pressKeyboardToSelectItem() {
+        await this.page.keyboard.press('Enter', { delay: KEYBOARD_DELAY });
+    }
```
ribeiroguilherme, line 145: Minor thing: These functions look more like a `utils` rather than  specific functions related the `card` model. 
I can foresee we replicating those in different parts of the tests . An idea would be to move them to something like `utils/keyboard.ts`, and reuse them across different tests.

packages/e2e-playwright/tests/card/expiryDate/card.expiryDatePolicies.hidden.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+        // Clear number
+        await card.deleteCardNumber();
+
+        // UI is reset
+        await expect(card.expiryDateField).toBeVisible();
+
+        // Card seen as invalid
+        cardValid = await page.evaluate('window.card.isValid');
+        await expect(cardValid).toEqual(false);
```
ribeiroguilherme, line 36: Is there another way to check if card is invalid/valid instead of inspecting the window object?

packages/e2e-playwright/tests/card/branding/card.branding.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+            await expect(brandingIconSrc).toContain('nocard.svg');
+
+            // Visible cvc field
+            await expect(card.cvcField).toBeVisible();
+
+            // with regular text
+            await expect(card.cvcLabelText).toHaveText(CVC_LABEL);
+
+            // and required cvc field
+            await expect(card.cvcField).toHaveClass(/adyen-checkout__field__cvc/);
```
ribeiroguilherme, line 68: Is there another way to test this instead of checking css classname?

ribeiroguilherme (approved):
Nice to see tests being moved to playwright 🙌 

My only note is that I see in some tests that we are testing implementation details (inspecting `window` data, or the availability of certain CSS classes,  a11y labels). We ideally should aim to avoid it and test mainly what the user sees on the screen.

But on the other hand, I know that the Card is a 'special' component due to its dependency on the secured fields, and doing these kind of E2E tests is way easier to validate that it works rather than the Unit tests. So I believe that for the Card component, it is ok to test the implementation details to this extend.

packages/e2e-playwright/tests/card/branding/card.branding.spec.ts
sponglord: But this is the _purpose_ of that part of the test - to see that a class is added/removed when a cvc field becomes optional or not

packages/e2e-playwright/models/card.ts
sponglord: Fair point. Will make a utils for this

packages/e2e-playwright/tests/card/expiryDate/card.expiryDatePolicies.hidden.spec.ts
sponglord: Not that I can think of. 
The test needs access to the actual _component_ that is initialised i.e. in `/app/src/pages/Cards/Cards.js` so that it can access the component's state
I don't know how else you could pass this to a test?

PR #2404, opened 2023-11-01
Chore/selected address modification

kunalkp:
## Summary
Handle Google Autocomplete/Experian APIs

- The above APIs requires 2 steps to select an address.
- This PR handles the 2nd step based on the availability of prop.

### DIFF
--- a/packages/lib/src/components/internal/Address/Address.tsx
+++ b/packages/lib/src/components/internal/Address/Address.tsx
@@ -49,8 +49,8 @@ export default function Address(props: AddressProps) {
         formatters: addressFormatters
     });
 
-    const setSearchData = selectedAddress => {
-        const updatedSelectedAddress = props.onSelectedAddress ? props.onSelectedAddress(selectedAddress) : selectedAddress;
+    const setSearchData = async selectedAddress => {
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/internal/Address/Address.tsx
```diff
@@ -50,10 +50,11 @@ export default function Address(props: AddressProps) {
     });
 
     const setSearchData = selectedAddress => {
+        const updatedSelectedAddress = props.onSelectedAddress ? props.onSelectedAddress(selectedAddress) : selectedAddress;
```
lebara-moaead, line 53: We might need await here @kunalkp  maybe do a test with async function

lebara-moaead (changes requested):

packages/lib/src/components/internal/Address/Address.tsx
kunalkp: Yes. Fixed!

ribeiroguilherme:
Hi @kunalkp ,

Thanks for contributing to the repo. Please follow up on the #2406 as we are working on the same changes there.

PR #2406, opened 2023-11-01
Feature: Adds onAddressSelected to address lookup functionality

m1aw:
The objective of this PR is to add `onAddressSelected` callback. This callback works to provide a point of integration with Address completion services that require a second callback to access additional information, such is the case of Google Place Autocomplete API.

## Tested scenarios

- [X] onAddressSelected is called with the full Address data
- [X] on success: the new data is used in the address component
- [X] on failure: nothing is selected

**Fixed issue**:

### DIFF
--- a/.changeset/fifty-parrots-collect.md
+++ b/.changeset/fifty-parrots-collect.md
@@ -2,4 +2,4 @@
 '@adyen/adyen-web': minor
 ---
 
-feature: adds onAddressSelected callback
+feature: adds new onAddressSelected to fill data when an item is selected in AddressSearch
--- a/.changeset/fifty-parrots-collect.md
+++ b/.changeset/fifty-parrots-collect.md
@@ -2,4 +2,4 @@
 '@adyen/adyen-web': minor
 ---
 
-feature: adds onAddresSelected callback
+feature: adds onAddressSelected callback
--- a/.changeset/fifty-parrots-collect.md
+++ b/.changeset/fifty-parrots-collect.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+feature: adds onAddresSelected callback
--- a/packages/lib/src/components/internal/Address/components/AddressSearch.tsx
+++ b/packages/lib/src/components/internal/Address/components/AddressSearch.tsx
@@ -31,7 +31,7 @@ interface AddressSearchProps {
     onManualAddress: any;
     externalErrorMessage: string;
     hideManualButton: boolean;
-    addressSearchDebounceMs: number;
+    addressSearchDebounceMs?: number;
 }
 
 export default function AddressSearch({
--- a/packages/lib/src/components/internal/Address/components/AddressSearch.tsx
+++ b/packages/lib/src/components/internal/Address/components/AddressSearch.tsx
@@ -1,7 +1,7 @@
 import Field from '../../FormFields/Field';
 import { Fragment, h } from 'preact';
 import { AddressLookupItem } from '../types';
-import { useCallback, useEffect, useState } from 'preact/hooks';
+import { useCallback, useEffect, useState, useMemo } from 'preact/hooks';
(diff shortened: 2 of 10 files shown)


.changeset/fifty-parrots-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+feature: adds onAddresSelected callback
```
moaead, line 5: This should be onAddressSelected

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -55,14 +71,32 @@ export default function AddressSearch({ onAddressLookup, onSelect, onManualAddre
...
+            onAddressSelected(value, { resolve, reject });
+        })
+            .then(fullData => {
+                onSelect(fullData);
+                setFormattedData([]);
+                return;
+            })
+            .catch(reason => {
+                setErrorMessage(reason);
+            });
```
moaead, line 99: Can we add a loading here?

moaead (changes requested):

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
m1aw: We have investigated this, yet we couldn't come up with a good solution. For example, Google own address solution doesn't have a loading indicator.
I will give the feedback to design and product.

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -55,17 +73,35 @@ export default function AddressSearch({ onAddressLookup, onSelect, onManualAddre
         setErrorMessage(externalErrorMessage);
     }, [externalErrorMessage]);
 
-    const onChange = event => {
+    const onChange = async event => {
```
ribeiroguilherme, line 76: What is the difference between `onInput` and `onChange`?

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -33,9 +52,9 @@ export default function AddressSearch({ onAddressLookup, onSelect, onManualAddre
...
 
     const onInput = useCallback(
-        async event => {
+        async (inputValue: string) => {
             new Promise<Array<AddressLookupItem>>((resolve, reject) => {
-                onAddressLookup(event, { resolve, reject });
+                onAddressLookup(inputValue, { resolve, reject });
             })
                 .then(data => {
                     setOriginalData(data);
```
ribeiroguilherme, line 60: What is the original data in this case? Is the data formatted according to the `AddressLookupItem` type? or is it just `any` data?

Adding types can help understand the logic better. It can be nice to add the type definitions to the `useState` too if possible, so it is clear what is being stored

packages/lib/src/components/internal/Address/components/AddressSearch.tsx
```diff
@@ -15,15 +16,33 @@ export type OnAddressLookupType = (
...
+    onAddressSelected?: OnAddressSelectedType;
+    onSelect: (addressItem: AddressData) => void;
     onManualAddress: any;
     externalErrorMessage: string;
     hideManualButton: boolean;
+    addressSearchDebounceMs?: number;
 }
 
-export default function AddressSearch({ onAddressLookup, onSelect, onManualAddress, externalErrorMessage, hideManualButton }: AddressSearchProps) {
+export default function AddressSearch({
```
ribeiroguilherme, line 37: This component could be unit tested - you could test the callbacks and passing back values to them.
It might be worth to do it, as it has potential to be used in multiple places.

m1aw: Done!

ribeiroguilherme (approved):

PR #2409, opened 2023-11-06
[WIP] Click to Pay - 3rd party cookies

ribeiroguilherme:
## Tested scenarios

**Fixed issue**:


packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPSaveCookiesCheckbox/CtPSaveCookiesCheckbox.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.adyen-checkout-ctp__otp-checkbox-container {
+  padding: 12px;
+  background-color: $color-gray-lighter;
+  border-radius: $border-radius-medium;
+  margin-bottom: 16px;
+}
+
+.adyen-checkout-ctp__otp-checkbox-container--checked {
+  background-color: #F2F8FF;
```
m1aw, line 11: Could we not approximate this color to something that exists?

packages/lib/src/components/internal/ClickToPay/context/ClickToPayContext.ts
```diff
@@ -37,8 +39,10 @@ const ClickToPayContext = createContext<IClickToPayContext>({
     configuration: null,
     isStandaloneComponent: null,
     isCtpPrimaryPaymentMethod: null,
+    isStoringCookies: null,
```
m1aw, line 42: Should this be false? Type says `boolean`. Otherwise `undefined` might be better.

packages/playground/src/pages/Cards/Cards.js
```diff
@@ -188,7 +188,7 @@ getPaymentMethods({ amount, shopperLocale }).then(async paymentMethodsResponse =
                     mcSrcClientId: '6d41d4d6-45b1-42c3-a5d0-a28c0e69d4b1'
                 },
                 clickToPayConfiguration: {
-                    shopperEmail: 'shopper@example.com',
+                    shopperEmail: 'shopper@email.com',
```
m1aw, line 191: Why? 😢 This is a real email.

packages/lib/src/components/internal/FormFields/Checkbox/Checkbox.scss
```diff
@@ -7,7 +7,6 @@
         position: relative;
         padding-left: 24px;
         cursor: pointer;
-        display: inline-block;
```
m1aw, line 10: I tested and this causes a change in OpenInvoice component.

m1aw (changes requested):

packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPSaveCookiesCheckbox/CtPSaveCookiesCheckbox.scss
ribeiroguilherme: We have that defined for bento on v6 branch. So, as soon as I merge this back to v6 , I will change these values to use the bento variables

packages/lib/src/components/internal/FormFields/Checkbox/Checkbox.scss
ribeiroguilherme: Can you highlight what is the change? Looking at the Open Invoices page in the old playground, I can't spot any problem

packages/lib/src/components/internal/ClickToPay/context/ClickToPayContext.ts
ribeiroguilherme: Sure. Somehow Typescript does not complain about it, even when informing it that this value is of type `boolean`.

packages/lib/src/components/internal/FormFields/Checkbox/Checkbox.scss
m1aw: https://github.com/Adyen/adyen-web/assets/6471825/37125f4b-f1c2-492c-974c-9319e8da5325

The top margin shifts but also the [ ] checkbox square itself shifts by 1px.

ribeiroguilherme: The checkbox shift is fine I believe, as the text gets a better vertical alignment next to the Checkbox. Regarding the margin shift, I will take a look with Arjen what is the correct pixels there

ribeiroguilherme: I updated some styles - the CSS should be ok now. Margin between elements is 16px, and from the pay button is 24px.

m1aw (approved):

PR #2422, opened 2023-11-08
fix: correct the Riverty consent links

longyulongyu:
## Summary
Correct the consent urls for Riverty / Afterpay
Translations are committed another [PR](https://github.com/Adyen/adyen-web/pull/2409/commits/2daf0c7c90bf366ab38df836cf52037cd865a38f)
## Tested scenarios
All tests passed.

**Internal ticket**:  COWEB-1289


packages/lib/src/components/AfterPay/utils.ts
```diff
@@ -1,10 +1,8 @@
-import { AFTERPAY_CONSENT_URL_BE, AFTERPAY_CONSENT_URL_EN, AFTERPAY_CONSENT_URL_NL } from './config';
+import { rivertyConsentUrlMap } from './config';
 
 function getConsentLinkUrl(countryCode: string, locale: string): string {
     const languageCode = locale?.toLowerCase().slice(0, 2);
-    if (languageCode === 'en') return AFTERPAY_CONSENT_URL_EN;
-    if (countryCode?.toLowerCase() === 'be') return AFTERPAY_CONSENT_URL_BE;
-    return AFTERPAY_CONSENT_URL_NL;
+    return rivertyConsentUrlMap[countryCode?.toLowerCase()]?.[languageCode];
```
ribeiroguilherme, line 5: It seems that the countryCode is an optional field . Should we make it mandatory in this case?
What happens if the `rivertyConsentUrlMap` mapping fails and undefined is returned?

longyulongyu: The problem is `countryCode` comes from the merchant's configuration and it's [optional](https://github.com/Adyen/adyen-web/blob/47bb84386abbd8d5e76ce5d2cfb84bc62a92b4c2/packages/lib/src/components/internal/OpenInvoice/types.ts#L25), so it can be optional and it would return `undefined`. 
Riverty also doesn't have a generic consent url, it always ties with country code and language code, so i think it's an expected case to have an `undefined` value as we just don't know which url to return..

ribeiroguilherme: I think we should still return a default value, as we used to do it before. It doesn't sound good to me to not provide the consent link.

Did you check with product if it is fine to not provide the consent link? What happens in the UI when we don't provide the URL?

longyulongyu: For the dropin if we don't specify the country code, the payment method would not render, so it's not an issue.

For using the component, the `countryCode` needs to be provided in the `paymentMethods` call and passed down to the `AdyenCheckout`, so in reality i think the `countryCode` will always be there. 

Nevertheless, also checked with the product, they are fine with no consent link, because if a wrong link is equivalent of no link. On the UI side, the shopper can see the link but cannot open anything.

ribeiroguilherme: Does it make sense to at least then trigger a console warning if the Component is initialized without country code or if the consent link is not defined properly due to this? Showing a link that does not open anything is a bug IMO. 

Furthermore, there are merchants that don't do `paymentMethods`,  call neither `sessions` , because they always use the same payment methods data. It is not recommended of course, but they do that. They can even omit the countryCode in this case if I am not mistaken.

longyulongyu: Hmm would it get rejected on the BE side, i think Riverty is enabled only for NL and BE? It's just we don't know which default country code to use, and product also doesn't have a 'default' consent link. Throwing a console warning definitely makes sense!

longyulongyu: I updated to console a warning when no consent url is found.

ribeiroguilherme (approved):

sponglord (approved):

PR #2425, opened 2023-11-13
v6 - Synchronous submit with order tracking

ribeiroguilherme:
### onSubmit/onAdditionalDetails requires data to be passed back

The payments response data must be passed back to the lib. Internally we handle the action; trigger the Partial payment flow; or check the payload and decide if the payment is completed or not.

```js
 onSubmit: async (state, component, actions) => {
    try {
        const { action, order, resultCode, donationToken } = await makePayment(state.data);

        if (!resultCode) actions.reject();

        actions.resolve({
            resultCode,
            action,
            order,
            donationToken
        });
    } catch (error) {
        console.error('## onSubmit - critical error', error);
        actions.reject();
    }
},
```

### onPaymentCompleted and onPaymentFailed

To have more granular control of what is happening:
- `onPaymentCompleted` is called when the payment is successfull 
- `onPaymentFailed` (new callback) is called when the payment fails
- Both callbacks work on Advanced Flow and Sessions Flow

### Partial payments
- `onOrderCreated` renamed to `onOrderUpdated` to be more concise
- Added new callback `onPaymentMethodsRequest` which is triggered when there is a partial order pending. The merchant then can request the new payment methods in order to update the UI accordingly

### onAuthorized

onAuthorized now is triggered **before** the 'onSubmit' . It returns an object with the **authorizedEvent** (raw data from ApplePay/GooglePay/PayPal) along with the formatted **billingAddress** and **deliveryAddress**. 
- The `onAuthorized` must be resolved in order to trigger the payment flow (onSubmit) 
- It can be reject, then the payment flow is cancelled
- The `onAuthorized` for Paypal happens **before** the onAdditionalDetails, as the authorization data is available at that point

```js
onAuthorized({ authorizedEvent, billingAddress, deliveryAddress }, actions) {
  actions.resolve();
},
```

### ApplePay onOrderTracking

Adding support for ApplePay order tracking.
- Once the 'onSubmit' is resolved, we check if the `onOrderTrackingRequest` callback is available. If it is, it is called and then the merchant must create and pass the order details back to the Component.
- The merchant can also 'reject' inside the 'onOrderTrackingRequest'. That is used in case something fails. It won't stop the payment to be sucessfull

```js
onOrderTrackingRequest: async (resolve, reject) {
  const order = await createOrderUsingMerchantBackend(...);

  const orderDetails = {
      orderTypeIdentifier: order.orderTypeIdentifier,
      orderIdentifier:  order.orderIdentifier,
      webServiceURL: order.webServiceURL,
      authenticationToken: order.authenticationToken
  };
  resolve(orderDetails);
}
```

### AVS improvements for GooglePay and ApplePay
If the information about the billing address and shipping address is available in the GooglePay/Applepay components when the authorization happens, we populate the `state.data` with the **billingAddress** and **deliveryAddress** fields. This improves the AVS checks.

### ApplePay improvements
If the payment fails during the `onSubmit` call, or if the merchant rejects the payment during the `onAuthorized` callback, the ApplePay payment overlay will display 'Payment Failed' message

The error can also be customized in the 'onSubmit' by doing:
```js
actions.resolve({
  resultCode: 'Refused',
  error: {
    applePayError: new ApplePayError( /**  props */ )
  }
});
```
or in the `onAuthorized` by doing:
```js
actions.reject(new ApplePayError( /**  props */ ));
```

### GooglePay improvements
If the payment fails during the `onSubmit` call, or if the merchant rejects the payment during the `onAuthorized` callback, the GooglePay overlay will display 'Payment Failed' message. 

The error can also be customized in the 'onSubmit' by doing:
```js
actions.resolve({
  resultCode: 'Refused',
  error: {
    googlePayError: 'Insufficient funds. Please try with another Card'
  }
});
```
or in the `onAuthorized` by doing:
```js
actions.reject('Network not supported');
```
**WIP**:
- Fix ThreeDS and Card `onError` callback .(Align with Core `onError`)
- Partial order types. Amazon pay flow

## Tested scenarios

**Fixed issue**:


packages/lib/storybook/stories/wallets/AmazonPayExample.tsx
```diff
@@ -0,0 +1,132 @@
...
+                                alert(response.resultCode);
+                            } else {
+                                // Try handling the decline flow
+                                // This will redirect the shopper to select another payment method
+                                component.handleDeclineFlow();
+                            }
+                        })
+                        .catch(error => {
+                            throw Error(error);
+                        });
```
ribeiroguilherme, line 85: @m1aw It is still using the old syntax for the `onSubmit`: handling action part with `component.handleAction`;  checking the resultCode manually instead of passing back to the library; I believe it also does not trigger the new callbacks onPaymentCompleted/onPaymentFailed. The goal is to make it work with the new syntax at least, so all payment methods follow the new flow.

Perhaps this needs an adjustment like it was done with GooglePay/ApplePay, as it seems that AmazonPay has an special flow to decline the payment (`component.handleDeclineFlow`)

Would that be possible?

m1aw: To the first part. I have changed it to use `actions.resolve` as part of the tests. I will pull those changes in.
On the second part, I have not made modifications to the way AmazonPay `onSubmit` works and also the way handleDeclineFlow works.

I thought that was the objective, not make more changes then the ones necessary here. I will take a look to also adjust it, I because I think I misunderstood it.

ribeiroguilherme: Adjusting the callback to behave in the new way is a necessary one I'd say. We don't want merchants to have to implement two different submit flows depending on the payment method that they use.

Let's then merge this PR as it is blocking other tasks, and then you can iterate on this in a new PR along with the e2e tests.

sponglord (approved):

PR #2428, opened 2023-11-14
Fix/v6 dropping TxVariants prop for Cards

sponglord:
## Summary
Removing the `static` property `txVariants` for Card related components.
Rationalising how `type` and `brand` are passed down from the various `Card` components to their `CardInputs` and on to the `SecuredFieldProvider` (covers `Card`, `StoredCard`, `CustomCard` & `Bancontact`)

## Tested scenarios
All card related e2e tests pass


packages/lib/src/components/Ach/components/AchInput/AchInput.tsx
```diff
@@ -204,6 +204,7 @@ export default AchInput;
 
 const extractPropsForSFP = (props: ACHInputProps) => {
     return {
+        allowedDOMAccess: props.allowedDOMAccess,
```
ribeiroguilherme, line 207: Looks like the `allowedDOMAccess` got added again to the codebase. Is it intentional?

packages/lib/src/components/Card/Card.tsx
```diff
@@ -66,7 +65,7 @@ export class CardElement extends UIElement<CardElementProps> {
             // billingAddressRequired only available for non-stored cards
             billingAddressRequired: props.storedPaymentMethodId ? false : props.billingAddressRequired,
             // ...(props.brands && !props.groupTypes && { groupTypes: props.brands }),
-            type: props.type === 'scheme' ? 'card' : props.type,
+            type: TxVariants.card,
```
ribeiroguilherme, line 68: Do we still need the `TxVariant.card` ? What is it used for? I expected that we would need only `TxVariant.scheme`

packages/lib/src/components/Ach/components/AchInput/AchInput.tsx
sponglord: I think there was a merge error. It should be fixed now

packages/lib/src/components/Card/Card.tsx
sponglord: `this.props.type` (with value `TxVariant.card`) is an internal property that propagates down to the securedFields to tell them what "type" they are.
All the securedFields code is built around being a "card"

ribeiroguilherme: Ah, ok.. seems to be a massive change then 😅

PR #2458, opened 2023-11-22
Klarna B2B - Fixing logo on Drop-in

ribeiroguilherme:
## Summary
The logo on UIElement currently is being build using the `this.constructor[type]` . This implies that it will always use the main tx variant defined in the Component, although there are components that have multiple TxVariants (for ex: Klarna, DragonPay, etc).

So instead of getting the main tx variant to use as icon, we are now pulling the tx variant from the `type()` getter which will decide if we take from the `type` property or from the `this.constructor[type]`.

## Tested scenarios
- Proper icon is rendered
- Unit test added to make sure the `icon` method pulls the txVariant from `type` getter

### DIFF
--- a/packages/lib/src/components/UIElement.tsx
+++ b/packages/lib/src/components/UIElement.tsx
@@ -239,7 +239,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
      * Get the element icon URL for the current environment
      */
     get icon(): string {
-        return this.props.icon ?? this.resources.getImage()(this.constructor['type']);
+        return this.props.icon ?? this.resources.getImage()(this.type);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/UIElement.tsx
```diff
@@ -239,7 +239,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
      * Get the element icon URL for the current environment
      */
     get icon(): string {
-        return this.props.icon ?? this.resources.getImage()(this.constructor['type']);
+        return this.props.icon ?? this.resources.getImage()(this.type);
```
sponglord, line 242: Do all `UIElement`s have `this.type` set?

sponglord (approved):

PR #2467, opened 2023-11-24
Fix/v6 improved integration warning

sponglord:
## Summary
Created a more informative warning for when Dropin detects that not all the supported PMs have been imported. The warning matches `txVariants` to the _actual_ Class that will be required e.g.
> You support the payment method 'sepadirectdebit' but this component has not been configured. Make sure to import the Class  'Sepa' and then pass it in the Dropin's 'paymentMethodComponents' config property if you wish to offer this payment method.

## Tested scenarios
Warnings given for "non-imported" PM classes


ribeiroguilherme (approved):
I believe that we can delete the [core.registry warning](https://github.com/Adyen/adyen-web/blob/v6/packages/lib/src/core/core.registry.ts#L43) and the [supportedTxVariables](https://github.com/Adyen/adyen-web/blob/v6/packages/lib/src/core/core.registry.ts#L28) variable from it. Then we keep only the Drop-in warning.

If I am not mistaken, the reason that the `core.registry` had its own `console.warn` is because previously the idea was to perform the component registration even with standalone component, and we also initially started the POC with using `create` function. Ex:

```js
  components: [Card],
  ...
});
checkout.create('card')..
```
The code above would trigger a console.warn from the core.registry.

As the standalone components auto-register themselves now , it seems like we just need this warning for Drop-in

PR #2519, opened 2024-01-16
Adds Gitpod support to the playground

jlengrand:
## Summary

Avoids people having to setup a local environment and instead runs fully in the browser.
This has been requested by our IMs (and some merchants) who are not allowed to install external repositories on their laptop. Using Gitpod (or codespaces) allows them to work without contaminating their local machine. I also allows for a better developer experience, since the `yarn` process is automated by the `.gitpod.yml` config.

I had to slightly modify the playground webpack dev server config to go around [that issue](https://stackoverflow.com/questions/43619644/i-am-getting-an-invalid-host-header-message-when-connecting-to-webpack-dev-ser). Given that we are talking about a playground here, I would argue that this causes not large security threat.
Additional hosts might have to be added if we want to support, say Codespaces. 

Happy to answer other questions :).

## Tested scenarios

Not relevant

**Fixed issue**:  Internal, DR-520


m1aw:
Hey @jlengrand do you know how I can test this? My only concern is the playground being served on a *.gitpod.com address which means we would need to access it via HTTPS otherwise secure fields won't work.

PR #2521, opened 2024-01-18
Feature/using checkoutanalytics mvp

sponglord:
## Summary
Adding the analytics events defined as the MVP

| Data point | Component | Status |
| --- | --- | --- |
|A Component/Dropin was displayed to the shopper |Component/Dropin | ✅  |
| iDeal featured issuer tile selected | iDeal | TODO |
|List opened | iDeal | TODO |
|List item selected | iDeal | TODO |
|Search usage (dropdown or list) | iDeal |TODO |
| Focus on an Input Field | Card| ✅ |
|Unfocus from an Input Field | Card | ✅ |
|Invalid Field | Card |  ✅ |
|Unsupported card entered | Card |  ✅ |
|Component Submitted (Pay button click) | All | ✅ |
|Action Was Handled| All | ✅ |
| File Download | QR codes| TODO |
| Instant payment method button click | GooglePay/ApplePay| ✅  |
| 3DS2 challenge displayed | Card |TODO |
| 3DS2 challenge ended | Card | TODO |

NOTE: TODOs will be done in separate PR(s) to keep the changelist minimal

## Tested scenarios
Current unit tests all pass

### DIFF
--- a/packages/lib/src/components/ApplePay/ApplePay.tsx
+++ b/packages/lib/src/components/ApplePay/ApplePay.tsx
@@ -10,6 +10,7 @@ import { preparePaymentRequest } from './payment-request';
 import { resolveSupportedVersion, mapBrands } from './utils';
 import { ApplePayElementProps, ApplePayElementData, ApplePaySessionRequest, OnAuthorizedCallback } from './types';
 import AdyenCheckoutError from '../../core/Errors/AdyenCheckoutError';
+import { ANALYTICS_SELECTED_STR } from '../../core/Analytics/constants';
 
 const latestSupportedVersion = 14;
 
@@ -53,6 +54,11 @@ class ApplePayElement extends UIElement<ApplePayElementProps> {
     }
 
     submit() {
+        // Analytics
+        if (this.props.isInstantPayment) {
+            this.submitAnalytics({ type: ANALYTICS_SELECTED_STR, target: 'instant_payment_button' });
+        }
+
         return this.startSession(this.props.onAuthorized);
     }
 
--- a/packages/lib/src/components/Card/Card.tsx
+++ b/packages/lib/src/components/Card/Card.tsx
@@ -5,18 +5,19 @@ import CoreProvider from '../../core/Context/CoreProvider';
 import collectBrowserInfo from '../../utils/browserInfo';
 import { BinLookupResponse, CardElementData, CardElementProps } from './types';
 import triggerBinLookUp from '../internal/SecuredFields/binLookup/triggerBinLookUp';
-import { CbObjOnBinLookup } from '../internal/SecuredFields/lib/types';
+import { CbObjOnBinLookup, CbObjOnFocus } from '../internal/SecuredFields/lib/types';
 import { reject } from '../internal/SecuredFields/utils';
 import { hasValidInstallmentsObject } from './components/CardInput/utils';
 import createClickToPayService from '../internal/ClickToPay/services/create-clicktopay-service';
 import { ClickToPayCheckoutPayload, IClickToPayService } from '../internal/ClickToPay/services/types';
 import ClickToPayWrapper from './components/ClickToPayWrapper';
-import { PayButtonFunctionProps, UIElementStatus } from '../types';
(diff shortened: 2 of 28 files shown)


packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -53,6 +54,11 @@ class ApplePayElement extends UIElement<ApplePayElementProps> {
     }
 
     submit() {
+        // Analytics
+        if (this.props.isInstantPayment) {
+            this.submitAnalytics({ type: ANALYTICS_SELECTED_STR, target: 'instant_payment_button' });
+        }
+
```
ribeiroguilherme, line 61: Is this supposed to be 'selected' or 'clicked'? 
Here we are tracking when the button is clicked.

packages/lib/src/components/Card/Card.tsx
```diff
@@ -162,6 +165,53 @@ export class CardElement extends UIElement<CardElementProps> {
...
+
+        // Call merchant defined callback
+        if (ALL_SECURED_FIELDS.includes(obj.fieldType)) {
+            this.props.onBlur?.(obj.event);
+        } else {
+            this.props.onBlur?.(obj);
+        }
+    };
+
+    private onErrorAnalytics = obj => {
```
ribeiroguilherme, line 206: Is it possible to add proper typescript types to the functions signatures and objects? 
We are doing big improvements on v6 regarding this, so it would be nice to try keep improving the types of the codebase.

packages/lib/src/components/GooglePay/GooglePay.tsx
```diff
@@ -45,6 +46,11 @@ class GooglePay extends UIElement<GooglePayProps> {
     }
 
     public submit = () => {
+        // Analytics
+        if (this.props.isInstantPayment) {
+            this.submitAnalytics({ type: ANALYTICS_SELECTED_STR, target: 'instant_payment_button' });
+        }
```
ribeiroguilherme, line 52: Same question that I asked on the ApplePay file applies here

packages/lib/src/components/UIElement.tsx
```diff
@@ -101,13 +110,40 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
...
+                const { validationErrorCode, validationErrorMessage } = analyticsObj;
+                this.props.modules?.analytics.createAnalyticsEvent({
+                    event: 'info',
+                    data: { component, type, target, validationErrorCode, validationErrorMessage }
+                });
+                break;
+            }
+
+            case ANALYTICS_SELECTED_STR:
+                this.props.modules?.analytics.createAnalyticsEvent({
```
ribeiroguilherme, line 139: Isn't possible to move all this analytics logic to the analytics module?

The UIElement could just pass the required information back to the module, and all the logic about verifying events, and creating them, etc,  would be located there.

That would keep the logic in the UIElement more simple, and most of the analytics code would be in their right place, easy to find and easier to be tested too.

ribeiroguilherme (approved):

packages/lib/src/components/ApplePay/ApplePay.tsx
sponglord: 'selected' is correct. That's what's been decided between the SDKs - we want to know when an instant PM has bee n chosen by the shopper

packages/lib/src/components/GooglePay/GooglePay.tsx
sponglord: See previous answer

packages/lib/src/components/UIElement.tsx
sponglord: I'll give it a try. There will still always be component related logic that will have to be applied (what is the component's `type`, is it a stored Card?) - things a UIElement "knows". So we will always need this `submitAnalytics` function. But, I agree the generation of the analytics objects, based on type, should be part of the Analytics Module

packages/lib/src/components/Card/Card.tsx
sponglord: Will do

PR #2522, opened 2024-01-19
Feature: Use backend label for stored Blik

m1aw:
## Summary

For Blik we want to display a backend label to distinguish between stored payment methods.

## Tested scenarios

See tests.

**Fixed issue**:


ribeiroguilherme (approved):
Don't forget to merge this into v6 after merging into master. 
You will need to fix the test cases there

longyulongyu (approved):

PR #2531, opened 2024-01-24
Checkoutanalytics mvp with 3DS2 events

sponglord:
## Summary
- Adding analytics events for the 3DS2 process (data sent, iframe loaded, fingerprint/challenge complete)
- Moved app specific analytics logic to a separate processing function (`analyticsPreProcessor`) in order to separate concerns and make testing easier
- Added `"configured"` type to info event - generate when a component's securedFields are all configured and ready for use
- Added "focus" & "unfocus" info events for the Custom Card component

## Tested scenarios
All existing unit tests run
Added unit tests for all the analytics events that we currently generate

<hr>

**NOTE:** extends changes made in #2521

### DIFF
--- a/packages/lib/src/components/ThreeDS2/components/Challenge/types.ts
+++ b/packages/lib/src/components/ThreeDS2/components/Challenge/types.ts
@@ -2,11 +2,13 @@ import { ChallengeData, ThreeDS2FlowObject } from '../../types';
 import { ChallengeResolveData } from '../utils';
 import { ThreeDS2ChallengeProps } from '../../ThreeDS2Challenge';
 import { ActionHandledReturnObject } from '../../../types';
+import { SendAnalyticsObject } from '../../../../core/Analytics/types';
 
 export interface DoChallenge3DS2Props extends ChallengeData {
     onCompleteChallenge: (resolveObject: ThreeDS2FlowObject) => void;
     onErrorChallenge: (rejectObject: ThreeDS2FlowObject) => void;
     onActionHandled: (rtnObj: ActionHandledReturnObject) => void;
+    onFormSubmit: (msg: string) => void;
 }
 
 export interface DoChallenge3DS2State {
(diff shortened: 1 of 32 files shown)


packages/lib/src/components/UIElement.tsx
```diff
@@ -78,8 +78,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
 
         const { type } = analyticsObj;
 
-        // let storedCardIndicator;
-        if (type === ANALYTICS_RENDERED_STR) {
+        if (type === ANALYTICS_RENDERED_STR || type === ANALYTICS_CONFIGURED_STR) {
             // Check if it's a storedCard
```
longyulongyu, line 82: Would it be possible to let the corresponding children components (e.g. `scheme` and `bcmc`) send the correct data `component` and `analyticsObj`?

I think the UIElement doesn't need to know anything about the children component, whether it's a stored card or not.

packages/lib/src/components/ThreeDS2/components/Challenge/types.ts
```diff
@@ -7,6 +7,7 @@ export interface DoChallenge3DS2Props extends ChallengeData {
     onCompleteChallenge: (resolveObject: ThreeDS2FlowObject) => void;
     onErrorChallenge: (rejectObject: ThreeDS2FlowObject) => void;
     onActionHandled: (rtnObj: ActionHandledReturnObject) => void;
+    onFormSubmit: (w) => void;
```
longyulongyu, line 10: Could we make the type and name of `w` more specific, i suppose it's an analytic message and the type is `string`?

I see it's being defined in many places, maybe we can create a share `type` for it.

sponglord: Done

packages/lib/src/components/UIElement.tsx
sponglord: I agree this is Card specific logic... I'll make the change

longyulongyu (approved):

PR #2540, opened 2024-02-07
Feature/v6 enforce country code

sponglord:
## Summary
Making `countryCode` a mandatory property when instantiating Checkout

## Tested scenarios
all unit tests pass


packages/lib/src/core/core.ts
```diff
@@ -212,11 +212,17 @@ class Core implements ICore {
      * @internal
      * Create or update the config object passed when AdyenCheckout is initialised (environment, clientKey, etc...)
      */
-    private setOptions = (options: CoreConfiguration): void => {
+    private setOptions = (options: CoreConfiguration, isUpdate = false): void => {
         if (hasOwnProperty(options, 'paymentMethodsConfiguration')) {
             console.warn('WARNING:  "paymentMethodsConfiguration" is supported only by Drop-in.');
         }
 
+        if (!isUpdate && !options.session) {
```
longyulongyu, line 220: I think putting the logic separately in the `update` and `initialize` would be better. Because it keeps the `setOptions`'s single responsibility.

sponglord: But then there's actually _three_ places the logic has to be - which seems like unnecessary duplication to me

packages/lib/src/core/core.ts
```diff
@@ -212,11 +212,17 @@ class Core implements ICore {
...
+            if (!hasOwnProperty(options, 'countryCode')) {
+                throw new AdyenCheckoutError(IMPLEMENTATION_ERROR, 'You must specify a countryCode when initializing checkout');
+            }
+        }
+
```
ribeiroguilherme, line 225: This check does not cover all scenarios. Example: creating a session using the `setup` call without passing the countryCode into it. 

I believe that ideally this check could go after the `initialize` function. I don't think we need to add the checks on the setOptions function nor keep watching if it is an update or not.

A draft of how it could look like:

```js
 public initialize(): Promise<ICore> {
        return new Promise<ICore>((resolve, reject) => {
            if (this.session) {
                this.session
                    .setupSession(this.options)
                    .then(sessionResponse => {
                        const { amount, shopperLocale, countryCode, paymentMethods, ...rest } = sessionResponse;

                        this.setOptions({
                            ...rest,
                            amount: this.options.order ? this.options.order.remainingAmount : amount,
                            locale: this.options.locale || shopperLocale,
                            countryCode: this.options.countryCode || countryCode
                        });

                        this.createPaymentMethodsList(paymentMethods);
                        this.createCoreModules();
                        resolve(this);
                    })
                    .catch(error => {
                        if (this.options.onError) this.options.onError(error);
                        reject();
                    });
            } else {
                this.createPaymentMethodsList();
                this.createCoreModules();
                resolve(this);
            }
        }).then((checkout: ICore) => {
            if (!this.options.countryCode) {
                throw new AdyenCheckoutError(IMPLEMENTATION_ERROR, 'You must specify a countryCode when initializing checkout');
            }
            return checkout;
        });
    }
```

In this way, once the SDK is fully initialized with sessions or non-sessions flow, we perform the check for the countryCode.

packages/lib/src/core/core.ts
longyulongyu: I think duplication would be okay in some scenarios. I feel adding `isUpdate` flag and checking upon `option.session` adds more complexity to a very simple setter function😄. 
As a result, it does more than just setting the option, it needs to know if it's an update operation and it needs to know if it's a session flow.

packages/lib/src/core/core.ts
sponglord: Can you create a session without a countryCode? I didn't think that was possible?

packages/lib/src/core/core.ts
sponglord: done

packages/lib/src/core/core.ts
ribeiroguilherme: I tried here. It seemed to work

sponglord: From the API team "You are allowed, but really shouldn't"

sponglord: > I don't think we need to add the checks on the setOptions function nor keep watching if it is an update or not.

They've already been removed

ribeiroguilherme: > From the API team "You are allowed, but really shouldn't"

Shouldn't we then move the check to the `initialize` block, as I suggested above? If merchants are allowed .. 🤷‍♂️

sponglord: So if we _do_ need to check the session for a countryCode (it's apparently not enforced for legacy reasons, it might be in the future but that's TBD)..
your solution seems very complicated, wrapping a promise around the whole process that has to be resolved/rejected.
Why don't we just to this:
```
public initialize(): Promise<this> {
        if (this.session) {
            if (!hasOwnProperty(this.options.session, 'countryCode')) {
                throw new AdyenCheckoutError(IMPLEMENTATION_ERROR, 'You must specify a countryCode when creating a session');
            }
            return this.session
                .setupSession(this.options)
                .then(sessionResponse => {
                    const { amount, shopperLocale, countryCode, paymentMethods, ...rest } = sessionResponse;

                    this.setOptions({
                        ...rest,
                        amount: this.options.order ? this.options.order.remainingAmount : amount,
                        locale: this.options.locale || shopperLocale,
                        countryCode: this.options.countryCode || countryCode
                    });

                    this.createPaymentMethodsList(paymentMethods);
                    this.createCoreModules();

                    return this;
                })
                .catch(error => {
                    if (this.options.onError) this.options.onError(error);
                    return this;
                });
        }

        this.createPaymentMethodsList();
        this.createCoreModules();

        return Promise.resolve(this);
    }
```

Which works in conjunction with the changes already suggested by Yu

ribeiroguilherme: Fair enough, we can indeed perform the check on the line 3 in the code above.  

For some reason I wanted to check countryCode property of the following line:
```js
const { amount, shopperLocale, countryCode, paymentMethods, ...rest } = sessionResponse;
```
But there is no need for that indeed.

ribeiroguilherme (approved):

PR #2542, opened 2024-02-07
v6 Feature: AmazonPay updated flow and onOrderCancel for advance flow

m1aw:
## Summary

The branch is ready to be merged. 

The main objective of this PR is to add support for the "new" onSubmit flow to AmazonPay.
The main challenge is that AmazonPay only trigger the onSubmit flow once, in the last step of the payment.

`handleDeclineFlow` should now be triggered in the in the `onPaymentFailed`.

Changes were made to `handleAdvanceFlowPaymentMethodsUpdate` to ensure the amount is also updated when reseting the dropin back to it's original state.

Also `onOrderCancel` can be a Chekcout property instead of a dropin property to be in line with the other callbacks. However it actually only works in the dropin.

Future discussion notes:
- The `handleAdvanceFlowPaymentMethodsUpdate` is now handling logic that I think should be moved to .initialize()
- We should probably take a look into unifying the types of payment methods so when tese callbacks are used UIElement is not returned, but instead the actual type of the element

## Tested scenarios

- Amazon pay was tested in the advance flow as standalone component
  - onPaymentFailed and onPaymentCompleted were called
  - payments were sucesfull 
- onOrderCancel was tested in the playground

**Fixed issue**:


packages/lib/src/components/Dropin/types.ts
```diff
@@ -95,14 +95,23 @@ export interface DropinConfiguration extends UIElementProps {
...
 export interface DropinComponentProps extends DropinConfiguration {
     core: ICore;
     onCreateElements: any;
     onChange: (newState?: object) => void;
-    onOrderCancel?: (data: onOrderCancelData) => void;
+    onOrderCancel?: (
+        data: onOrderCancelData,
+        actions: {
+            resolve: (data: { amount: PaymentAmount }) => void;
+            reject: () => void;
```
ribeiroguilherme, line 112: I believe the `reject ` must accept an `error?: string` as parameter in the function, as it seems that the merchant can reject the cancel order step with a custom error  by looking at the `getOnOrderCancel` implementation

packages/lib/src/core/types.ts
```diff
@@ -184,14 +192,7 @@ export interface CoreConfiguration {
...
-        state: any,
-        element: UIElement,
-        actions: {
-            resolve: (response: CheckoutAdvancedFlowResponse) => void;
-            reject: () => void;
-        }
-    ): void;
+    onSubmit?: onSubmitGenericType<UIElement>;
 
     /**
```
ribeiroguilherme, line 197: The type `onOrderCancel` in Core is not updated to reflect its new interface

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -97,15 +99,26 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
     private getOnOrderCancel = () => {
         if (this.props.onOrderCancel) {
             return (data: onOrderCancelData) => {
-                this.props.onOrderCancel(data);
+                const order = sanitizeOrder(data.order);
+                new Promise<{ amount: PaymentAmount }>((resolve, reject) => {
+                    this.props.onOrderCancel({ order }, { resolve, reject });
+                })
+                    .then(({ amount }) => this.props.elementRef.handleAdvanceFlowPaymentMethodsUpdate(null, amount))
```
ribeiroguilherme, line 106: Wouldn't be possible in this case to get the initial amount direct from the core? 
Then we could skip the step of asking the merchant to pass the amount back to us.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -97,15 +99,26 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
             };
         }
         if (this.props.session) {
             return (data: onOrderCancelData) =>
                 this.props.session
                     .cancelOrder(data)
                     .then(() => this.props.core.update({ order: null }))
-                    .catch(error => this.setStatus(error?.message || 'error'));
+                    .catch(error => {
+                        console.error(error);
```
ribeiroguilherme, line 119: Is it intentional to leave `console.error` ?

packages/lib/src/core/types.ts
```diff
@@ -47,6 +47,14 @@ export interface ICore {
...
 export type AdyenEnvironment = 'test' | 'live' | 'live-us' | 'live-au' | 'live-apse' | 'live-in' | string;
 
+export type onSubmitGenericType<T> = (
+    state: any,
+    element: T,
+    actions: {
+        resolve: (response: CheckoutAdvancedFlowResponse) => void;
+        reject: () => void;
+    }
+) => void;
```
ribeiroguilherme, line 57: I like the idea of trying to narrow the scope of the `onSubmit` function to the respective component, although we should investigate how can we perform this not only for the `onSubmit` but for all other callbacks as well that return an UIElement.

I suggest to start tackling this generic types in another PR.

packages/playground/src/pages/DropinUMD/manual.js
```diff
@@ -63,9 +63,9 @@ export async function initManual() {
         onOrderRequest: async (resolve, reject) => {
             resolve(await createOrder({ amount }));
         },
-        onOrderCancel: async order => {
+        onOrderCancel: async (order, actions) => {
             await cancelOrder(order);
-            checkout.update({ paymentMethodsResponse: await getPaymentMethods({ amount, shopperLocale }), order: null, amount });
+            actions.resolve({ amount });
```
ribeiroguilherme, line 68: What happens if the `actions.reject('Something went wrong') ` gets called? Do we display this error in the UI?

ribeiroguilherme (changes requested):

packages/lib/src/components/Dropin/components/DropinComponent.tsx
m1aw: For now we decided to not change this behaviour since it is similar on how the implementation already works (in v5) for the advance flow.
In v5 when the .update() call the merchant needs to pass the amount.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
m1aw: Change this to onError and NETWORK_ERROR.

packages/lib/src/core/types.ts
m1aw: Check if this change needs to be in this PR.

packages/playground/src/pages/DropinUMD/manual.js
m1aw: No error is going to be displayed in the UI. There is no UI design for it, if it's well implemented it shouldn't fail besides a failed network request. So we just throw network error now.

packages/lib/src/components/Dropin/types.ts
```diff
@@ -101,14 +101,25 @@ export interface DropinConfiguration extends UIElementProps {
...
+        orderData: string;
+        pspReference: string;
+    };
 }
 
+export type onOrderCancelType = (
+    data: onOrderCancelData,
+    actions: {
+        resolve: (data: { amount: PaymentAmount }) => void;
+        reject: (error: string) => void;
```
ribeiroguilherme, line 114: Not sure if we discussed this, but does it make sense to pass an error string here? what will happen in this case? is it just forwarded back as `AdyenCheckoutError` ?

ribeiroguilherme (approved):

PR #2545, opened 2024-02-08
v6 build pipeline update

ribeiroguilherme:
## Summary
- Upgraded to Rollup 4
- Upgraded other dev dependencies as well
- Set Node to run on version 18.18

**Fixed issue**:


.github/workflows/.size-limit.yml
```diff
@@ -8,14 +8,15 @@ jobs:
 
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
sponglord, line 11: Out of interest what does this `v4` refer to?

sponglord (approved):

packages/lib/.stylelintrc.json
```diff
@@ -4,6 +4,7 @@
         "no-descending-specificity": null,
         "selector-class-pattern": null,
         "scss/no-global-function-names": null,
-        "declaration-block-no-redundant-longhand-properties": null
+        "declaration-block-no-redundant-longhand-properties": null,
+        "media-feature-range-notation": null
```
sponglord, line 8: What does this new rule do?

.github/workflows/.size-limit.yml
ribeiroguilherme: Version 4 of the Github actions/checkout : https://github.com/actions/checkout

longyulongyu (approved):

packages/lib/.stylelintrc.json
ribeiroguilherme: https://stylelint.io/user-guide/rules/media-feature-range-notation/

As I updated stylelint, it started complaining about some media rules. I disabled the rule as it didn't seem like a big deal. Let me know if you think the opposite

PR #2546, opened 2024-02-09
v6 All UIElements can receive defaultProps based on regulations

sponglord:
## Summary
Finnish regulations state that no payment methods can be open by default...

This PR allows all `UIElements` to receive a set of `defaultProps` based on regulatory requirements (currently retrieved based on `countryCode`)
The functionality is now in place should this logic ever need to be expanded (based on other country's requirements &/or different logic entirely)

It should be noted that the merchant can still override these regulatory defaults if they want to. That's up to them. 
But at least CheckoutWeb will be compliant "out of the box"

## Tested scenarios
All current unit tests pass
Add new ones specific to this rule

**Relates to issue**: COAPI-293

### DIFF
--- a/packages/lib/src/components/Dropin/Dropin.test.ts
+++ b/packages/lib/src/components/Dropin/Dropin.test.ts
@@ -38,12 +38,14 @@ describe('Dropin', () => {
         });
     });
 
+    // TODO - this test doesn't do anything
     describe('closeActivePaymentMethod', () => {
         test('should close active payment method', async () => {
             const dropin = new Dropin(checkout);
             const component = await mount(dropin.render());
             await component.update();
 
+            // re. TODO: dropin.dropinRef.state.activePaymentMethod = null, which is defined, so this assertion passes
             expect(dropin.dropinRef.state.activePaymentMethod).toBeDefined();
             dropin.closeActivePaymentMethod();
             expect(dropin.dropinRef.state.activePaymentMethod).toBeNull();
@@ -153,6 +155,26 @@ describe('Dropin', () => {
             expect(await screen.findByRole('radio')).toBeTruthy();
             dropin.setStatus('error');
             expect(await screen.findByText(/An unknown error occurred/i)).toBeTruthy();
+        });
+    });
+
+    describe('Complying with local regulations', () => {
+        test('when countryCode is Finland openFirstPaymentMethod & openFirstStoredPaymentMethod should be false', async () => {
+            checkout.options.countryCode = 'FI';
+
+            const dropin = new Dropin(checkout);
+
+            expect(dropin.props.openFirstPaymentMethod).toBe(false);
+            expect(dropin.props.openFirstStoredPaymentMethod).toBe(false);
+        });
+
+        test('if openFirstPaymentMethod & openFirstStoredPaymentMethod are set by merchant then these values should be used', async () => {
+            checkout.options.countryCode = 'FI';
+
+            const dropin = new Dropin(checkout, { openFirstPaymentMethod: true, openFirstStoredPaymentMethod: true });
+
+            expect(dropin.props.openFirstPaymentMethod).toBe(true);
+            expect(dropin.props.openFirstStoredPaymentMethod).toBe(true);
         });
     });
 });
--- a/packages/lib/src/components/Dropin/Dropin.test.ts
+++ b/packages/lib/src/components/Dropin/Dropin.test.ts
@@ -159,7 +159,14 @@ describe('Dropin', () => {
     });
 
     describe('Complying with local regulations', () => {
-        test('when countryCode is Finland openFirstPaymentMethod & openFirstStoredPaymentMethod should be false', async () => {
+        test('Default values for openFirstPaymentMethod & openFirstStoredPaymentMethod are true', () => {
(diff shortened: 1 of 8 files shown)


packages/lib/src/core/utils.ts
```diff
@@ -11,3 +11,16 @@ export function processGlobalOptions(globalOptions) {
         return r;
     }, {});
 }
+
+export function getDefaultPropsByCountryCode(countryCode: string): Record<string, any> {
```
ribeiroguilherme, line 15: I think we should distinguish if it is Drop-in or not in here. Otherwise, all payment methods from Finland will get these default props, although it must be currently used only in Drop-in.

It is not a big deal for now, but my point is that I'd rather just pass props if they are supposed to be passed and used by the Component. We can avoid pass props that won't be used.

Sending a flag `isDropin` and then perform the switch based on that would work I guess:

```js
export function getDefaultPropsByCountryCode(countryCode: string, isDropin: boolean): Record<string, any> {
  switch (countryCode) {
    case 'FI': {
       return isDropin ? { openFirstPaymentMethod: false, openFirstStoredPaymentMethod: false } : {}
    }
    default: 
      return {};
  }
}
...
```

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -83,7 +83,11 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
             ...componentProps
         };
 
-        this.props = this.formatProps({ ...this.constructor['defaultProps'], ...finalProps });
+        this.props = this.formatProps({
+            ...this.constructor['defaultProps'], // component defaults
+            ...this.core.getRegulatoryDefaults(), // regulatory defaults
```
ribeiroguilherme, line 88: Does the core need to know about this `getRegulatoryDefaults` function? 
We could skip adding an extra public function in the core. We could pull this directly from here: 

```js
this.props = this.formatProps({
  ...this.constructor['defaultProps'], // component defaults
  ...this.getRegulatoryDefaults({ countryCode: this.core.options.contryCode, isDropin: this.type === 'dropin'  }), 
  ...finalProps
});
```

sponglord: I had it in a separate, standalone function before, but then thought since we do things like `this.core.getPropsForComponent`  - isn't this retrieving and compiling the correct props for any component part of the core's responsibility?

packages/lib/src/core/utils.ts
sponglord: For this particular rule it makes sense, yes

packages/lib/src/components/internal/UIElement/UIElement.tsx
ribeiroguilherme: I'd say that the UIElement is responsible to build the properties of the UI component and retrieve from where it is necessary.

On core we have the `getCorePropsForComponent`, which basically returns the properties that belongs only to the core module: the initialized modules; the properly parsed loadingContext, cdnContext; createFromAction method, etc.

From my view, in this case, it is fine to just pull from an standalone function directly into the UIElement. Core does not need to know about these regulatory props. Another benefit is that we do not have to expose another public function of the core, as exposing public functions can lead to typescript merchants trying to use it for something else.

sponglord: Made changes after our discussion

packages/lib/src/components/Dropin/Dropin.test.ts
```diff
@@ -38,12 +38,14 @@ describe('Dropin', () => {
         });
     });
 
+    // TODO - this test doesn't do anything
```
ribeiroguilherme, line 41: Shall we remove this test then? 😅

packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -12,10 +12,11 @@ import { hasOwnProperty } from '../../utils/hasOwnProperty';
 import SRPanelProvider from '../../core/Errors/SRPanelProvider';
 import splitPaymentMethods from './elements/splitPaymentMethods';
 import type { ICore } from '../../core/types';
+import { IDropin } from './types';
```
ribeiroguilherme, line 15: Not a big deal, but I'd recommend to start using `import type { isDropin } from './types';` instead of direct import when importing typescript types. 

It helps making sure we are always importing only the 'type' and that the imported values will be eliminated during compilation. ([more here](https://medium.com/@quizzesforyou/import-vs-import-type-in-typescript-8e5177b62bea))

ribeiroguilherme (approved):

packages/lib/src/components/Dropin/Dropin.test.ts
sponglord: This was just something I found whilst making my other tests.
I think the correct action would be to actually make it work! But that's out of scope for this PR - I just left this as a reminder

packages/lib/src/components/Dropin/Dropin.tsx
sponglord: Do you mean `IDropin` rather than `isDropin`? :-P

PR #2550, opened 2024-02-13
Feature/using checkoutanalytics mvp qr codes

sponglord:
## Summary
Adding the analytics events for QRCodes (defined as the MVP)

| Data point | Component | Status |
| --- | --- | --- |
| Clicking the button to copy/download a QR code | QR Codes | ✅  | 

## Tested scenarios

------
**Relates to issue:** #2521


packages/lib/src/components/helpers/QRLoaderContainer.tsx
```diff
@@ -49,6 +50,10 @@ class QRLoaderContainer<T extends QRLoaderContainerProps = QRLoaderContainerProp
         };
     }
 
+    protected submitAnalytics = (aObj: SendAnalyticsObject) => {
```
longyulongyu, line 53: I think we can remove the override `submitAnalytics` since it doesn't have other logic except for calling the parent one.

sponglord: yup

longyulongyu (approved):

PR #2551, opened 2024-02-13
PayPal Express Flow

ribeiroguilherme:
## Summary
Implementing new callbacks to let PayPal work as an express payment method:
- https://developer.paypal.com/sdk/js/reference/#onshippingaddresschange
- https://developer.paypal.com/sdk/js/reference/#onshippingoptionschange
- Added the flag `userAction` , so merchant can configure the PayPal lightbox to display 'Continue to review page' label

## Tested scenarios
- Manual testing and some unit tests
- Added examples on Storybook

**Fixed issue**:


packages/playground/src/pages/Dropin/session.js
```diff
@@ -46,6 +46,28 @@ export async function initSession() {
...
+                    };
+
+                    if (data.shippingAddress.countryCode === 'US') {
+                        const { paymentData } = await patchPaypalOrder(patch);
+                        component.updatePaymentData(paymentData);
+                        return;
+                    }
+
+                    return actions.reject();
+                }
```
sponglord, line 70: Would it also be good to have an example of the `onShippingOptionsChange` callback? Or is the process the same?

packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -31,8 +35,11 @@ class PaypalElement extends UIElement<PayPalElementProps> {
         const intent: Intent = isZeroAuth ? 'tokenize' : props.intent || intentFromConfig;
         const vault = intent === 'tokenize' || props.vault;
 
+        const displayContinueToReviewPageButton = props.userAction === 'continue';
+
         return {
             ...props,
+            commit: displayContinueToReviewPageButton ? false : props.commit,
```
sponglord, line 42: Would it be clearer to pass the `displayContinueToReviewPageButton` through to the component as a boolean rather than having to re-evaluate/re-create `displayContinueToReviewPageButton` in `PayPalButtons.tsx`?

packages/playground/src/pages/Dropin/session.js
ribeiroguilherme: I focused mainly on documenting this in the Storybook playground. I should remove this code from here

packages/lib/src/components/PayPal/Paypal.tsx
ribeiroguilherme: We can. I wanted to avoid more prop drilling but it is ok to do that here

PR #2555, opened 2024-02-14
Fix the storybook showing result after redirection

longyulongyu:
## Summary
Fix the storybook showing result after redirection


packages/lib/storybook/stories/helpers/RedirectResultContainer.tsx
```diff
@@ -1,21 +1,33 @@
...
+        if (!redirectResult && !sessionId) {
             message = 'There is no redirectResult / sessionId provided';
             return;
         }
 
         AdyenCheckout({
             clientKey: process.env.CLIENT_KEY,
             environment: process.env.CLIENT_ENV,
-            session: { id: sessionId },
+            ...(sessionId && { session: { id: sessionId, countryCode } }),
```
longyulongyu, line 20: Here we already have the session id but we have to pass the country code otherwise it won't work. Failed at `initialize` function.

ribeiroguilherme (approved):

PR #2568, opened 2024-02-21
Porting latest v5 analytics changes to v6

sponglord:
## Summary
Porting latest v5 analytics changes, namely:
- Adding `platform:"Web" prop to all calls
- All events have a unique id

Also fixed issue where after a failed setup call we were still trying to send subsequent analytics events


packages/lib/src/core/Services/analytics/collect-id.ts
```diff
@@ -67,8 +67,8 @@ const collectId = ({ analyticsContext, clientKey, locale, analyticsPath }: Colle
                 return undefined;
             })
             .catch(() => {
-                console.debug(FAILURE_MSG);
-                return FAILURE_MSG;
+                console.warn(FAILURE_MSG);
+                return undefined;
```
longyulongyu, line 71: Can we return `Promise.reject('xxx')` here? Because it will be caught by `Analytics.ts` line 82, and `capturedCheckoutAttemptId` will not be assigned with `undefined` value.

sponglord: Alright, if you like.

longyulongyu: I think the `undefined` value is not a problem, it's more about utilizing the benefit of `try...catch` block. If `collectId` failed to get a valid id (no id or a network error), we could just reject the promise so that they all end up in the `catch` block. So it's guaranteed that we always have a valid id in the try block (non error flow).

PR #2581, opened 2024-02-28
Feature/v6 new error codes

sponglord:
## Summary
Aligning errorCodes generated for validation errors so that a common set of values can be shared between all the SDKs.
This has involved changing some of the translation keys used by securedFields.

Mostly however this PR revolves around defining constants and types - which is why it's so big!

**Still TODO**: 
- ~~wait for sf v5.0.0 to be on Test~~
- ~~run new translations job~~

## Tested scenarios
All unit tests pass


packages/lib/src/components/internal/SecuredFields/lib/securedField/SecuredField.ts
```diff
@@ -97,6 +98,10 @@ class SecuredField extends AbstractSecuredField {
         // Set result back onto config object
         this.sfConfig.iframeUIConfig.ariaConfig = processedAriaConfig;
 
+        // console.log('### SecuredField::init:: processedAriaConfig', processedAriaConfig);
+
+        console.log('### SecuredField::init:: processedAriaConfig', processedAriaConfig);
```
ribeiroguilherme, line 103: Is it intentional to leave the console.log here?

ribeiroguilherme (approved):
Is it missing the other translations files? 
I see that we updated some key-values on `en-us.json` ..

packages/lib/src/components/internal/SecuredFields/lib/securedField/SecuredField.ts
sponglord: Nope!

sponglord:
> Is it missing the other translations files? I see that we updated some key-values on `en-us.json` ..

Yeah. I won't actually merge until the translations job has run

PR #2586, opened 2024-03-04
Feature/using checkoutanalytics_with_base64 refactor

sponglord:
## Summary
Created greater granularity when detect errors in the `base64` decode process.
Ground work for the next PR when we use increase the granularity in detecting errors in the 3DS2 process

## Tested scenarios
All unit tests pass


packages/lib/src/utils/base64.ts
```diff
@@ -35,7 +50,11 @@ const base64 = {
             return false;
         }
 
-        return window.btoa(window.atob(pDataStr)) === pDataStr;
+        try {
+            return window.btoa(window.atob(pDataStr)) === pDataStr;
+        } catch (e) {
+            return false;
+        }
```
m1aw, line 57: I believe this check will fail 16 bit UTF char. This makes that we will never reach `b64DecodeUnicode` when calling in `decode`.

m1aw (approved):

PR #2596, opened 2024-03-11
fix(Address): only emit the Address form data for valid schema fields

longyulongyu:
## Summary
`getAddressSchemaForCountryFlat` returns the defined schema fields (for a specific country or default one).
When emitting the Address form data, we should filter out fields that are not in the specified schema.

## Tested scenarios
Add unit tests.


packages/lib/src/components/internal/Address/Address.tsx
```diff
@@ -120,7 +120,11 @@ export default function Address(props: AddressProps) {
 
     useEffect((): void => {
         const optionalFields = specifications.getOptionalFieldsForCountry(data.country);
+        // Country specific / default schema fields, include both required and optional fields.
+        const validSchemaFields = specifications.getAddressSchemaForCountryFlat(data.country);
         const processedData = ADDRESS_SCHEMA.reduce((acc, cur) => {
+            if (!validSchemaFields?.includes(cur)) return acc;
```
m1aw, line 126: Don't we need to also add the value of `firstName` to acc?
We just want to avoid line 134?

longyulongyu: No we don't want to add `firstName` and `lastName` for all the schemas, because that caused the issue [here](https://github.com/Adyen/adyen-web/pull/2591). 
Only if the schema explicitly mentions to show `firstName` & `lastName` for instance [Riverty](https://github.com/Adyen/adyen-web/blob/3ce12ef069ba39c4ae9c5f794261f5bbffb87e9a/packages/lib/src/components/Riverty/config.ts#L7) we put the key and value to the `acc`.

m1aw: Makes sense sorry. In my head I was just trying to avoid the `N/A` default scenario. If you don't pass things that are not in the schema also works.

PR #2597, opened 2024-03-12
v6 - Adding isExpress flag for GooglePay and ApplePay

ribeiroguilherme:
## Summary
- Adding `isExpress` flag . This flag makes the component add the `subtype: 'express'` to the payment data. This impacts the reports on CA, showing to merchants the payments that are being done using express flow.

## Tested scenarios
- Added tests


packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -35,7 +46,7 @@ class ApplePayElement extends UIElement<ApplePayConfiguration> {
     /**
      * Formats the component props
      */
-    protected formatProps(props) {
+    protected override formatProps(props) {
```
m1aw, line 49: It might be worth in the future fixing this everywhere in new PR, and adding `noImplicitOverride` to TS config.

m1aw (approved):

packages/lib/src/components/ApplePay/ApplePay.tsx
ribeiroguilherme: Yea. That would be nice. 
Actually I think we should start removing the disabled strict features that we have on `tsconfig.json` until the point the we get the 'strict: true' mode . That will make us have a more reliable codebase

packages/lib/src/components/GooglePay/GooglePay.test.ts
```diff
@@ -55,6 +55,21 @@ beforeEach(() => {
...
+            expect(googlepay.data.paymentMethod).toHaveProperty('subtype', 'express');
+        });
+        test('should not add subtype: express when isExpress is omitted', () => {
+            const googlepay = new GooglePay(global.core);
+            expect(googlepay.data.paymentMethod).not.toHaveProperty('subtype', 'express');
+        });
+
+        test('should throw error when express callbacks are passed but isExpress flag is not set', () => {
+            expect(() => new GooglePay(global.core, { paymentDataCallbacks: { onPaymentDataChanged: jest.fn() } })).toThrow();
+        });
```
asumaran, line 70: It would be great to see this documented somewhere. There's no mention about it on your documentation.

ribeiroguilherme: Hi @asumaran - we do mention in the [release notes](https://github.com/Adyen/adyen-web/releases/tag/v6.0.0), but indeed we are not mentioning it in the payment method page itself. I shared the feedback with the docs team. Thanks for flagging it

PR #2598, opened 2024-03-12
v6 - showPayButton clean up

ribeiroguilherme:
## Summary
- When we updated the library to have default `showPayButton: true` for all components , we missed in many different Components to remove setting the default prop `showPayButton: true`.  
- There were also some interfaces extending UIElement with `showPayButton` property, which is irrelevant


packages/lib/src/components/internal/IbanInput/IbanInput.tsx
```diff
@@ -11,7 +11,7 @@ interface IbanInputProps {
     holderName?: boolean;
     placeholders?: Omit<IbanData, 'countryCode'>;
     countryCode?: string;
-    showPayButton?: any;
+    showPayButton?: boolean;
```
m1aw, line 14: Why is it still set to optional here?

m1aw (approved):

PR #2604, opened 2024-03-18
Fix/v5 sr panel issues for open invoices

sponglord:
## Summary
Fixing `SRPanel` issues in `OpenInvoice` comp now that (delivery) addresses can have first & last names
Ensuring no first & last name errors get overwritten and that the `SRPanel` has the correct translations for every field

## Tested scenarios
All unit tests pass
Added Playwright e2e test for `Riverty` component and check every field in the `SRPanel`

### DIFF
--- a/packages/lib/src/components/internal/OpenInvoice/OpenInvoice.tsx
+++ b/packages/lib/src/components/internal/OpenInvoice/OpenInvoice.tsx
@@ -59,7 +59,7 @@ export default function OpenInvoice(props: OpenInvoiceProps) {
         fieldTypeMappingFn: mapFieldKey
     });
 
-    const specifications = useMemo(() => new Specifications(), []);
+    const specifications = useMemo(() => new Specifications(props.deliveryAddressSpecification), []);
     /** end SR stuff */
 
     const initialActiveFieldsets: OpenInvoiceActiveFieldsets = getInitialActiveFieldsets(visibility, props.data);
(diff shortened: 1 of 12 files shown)


packages/lib/src/components/internal/OpenInvoice/OpenInvoice.tsx
```diff
@@ -59,7 +59,7 @@ export default function OpenInvoice(props: OpenInvoiceProps) {
         fieldTypeMappingFn: mapFieldKey
     });
 
-    const specifications = useMemo(() => new Specifications(), []);
+    const specifications = useMemo(() => new Specifications(props.deliveryAddressSpecification), []);
```
longyulongyu, line 62: Why use `props.deliveryAddressSpecification` particularly 🤔? I would assume delivery address and billing address specifications can be different.

sponglord: This is an easy way to enhance the Billing Address specifications with with those needed by the deliveryAddress i.e. first and last name.
The other solution would be to create a separate specifications object just for DeliveryAddress. Is this what you are proposing?

longyulongyu: I don't think it's a problem so far, but it might get problematic when deliveryAddress and billingAddress schema differ too much, e.g (an extreme case): deliveryAddress schema contains postal code, street and country, billingAddress contains first name, last name. Then the layout would be completely wrong.

I like your first approach which decouples delivery address and billing address 💯.

sponglord: ok, I'll go with the first approach

longyulongyu (approved):

PR #2606, opened 2024-03-20
V6: remove country flag & add Riverty to the tx variant list

longyulongyu:
## Summary
- Remove the country flag
- Add Riverty to the tx variant list

## Tested scenarios
All tests passed

**Fixed issue**: COWEB-1346

### DIFF
--- a/packages/lib/src/components/internal/Address/components/CountryField.tsx
+++ b/packages/lib/src/components/internal/Address/components/CountryField.tsx
@@ -5,16 +5,14 @@ import useCoreContext from '../../../../core/Context/useCoreContext';
 import getDataset from '../../../../core/Services/get-dataset';
 import { CountryFieldProps, CountryFieldItem } from '../types';
 import Select from '../../FormFields/Select';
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/internal/Address/components/CountryField.tsx
```diff
@@ -5,16 +5,14 @@ import useCoreContext from '../../../../core/Context/useCoreContext';
 import getDataset from '../../../../core/Services/get-dataset';
 import { CountryFieldProps, CountryFieldItem } from '../types';
 import Select from '../../FormFields/Select';
-import { getFlagEmoji } from '../../../../utils/getFlagEmoji';
```
ribeiroguilherme, line 8: Should we remove this function as well?

ribeiroguilherme (approved):

packages/lib/src/components/internal/Address/components/CountryField.tsx
longyulongyu: It's already removed 😄

PR #2617, opened 2024-03-26
Chore/v5 refactor how wallet p ms call submit analytics

sponglord:
## Summary
Refactor how wallet PMS call `submitAnalytics` when they are used as express.PMS. 
We move the logic to `analyticsPreProcessor` so the wallet PMs can keep their `submitAnalytics` override as simple as possible

## Tested scenarios
All unit tests pass
Added new unit tests to test `analyticsPreProcessor` and how it handles the `"rendered"` event - which is the one that need to set the `isExpress` and `expressPage` properties for express PMs

### DIFF
--- a/packages/lib/src/components/UIElement.tsx
+++ b/packages/lib/src/components/UIElement.tsx
@@ -67,7 +67,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
      *  In some other cases e.g. 3DS2 components, this function is overridden to allow more specific analytics actions to be created
      */
     /* eslint-disable-next-line */
-    protected submitAnalytics(analyticsObj: SendAnalyticsObject) {
+    protected submitAnalytics(analyticsObj: SendAnalyticsObject, uiElementProps?) {
         /** Work out what the component's "type" is:
          * - first check for a dedicated "analyticsType" (currently only applies to custom-cards)
          * - otherwise, distinguish cards from non-cards: cards will use their static type property, everything else will use props.type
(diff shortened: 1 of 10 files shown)


packages/lib/src/components/UIElement.tsx
```diff
@@ -77,7 +77,7 @@ export class UIElement<P extends UIElementProps = any> extends BaseElement<P> im
             component = this.constructor['type'] === 'scheme' || this.constructor['type'] === 'bcmc' ? this.constructor['type'] : this.props.type;
         }
 
-        this.props.modules?.analytics.sendAnalytics(component, analyticsObj);
+        this.props.modules?.analytics.sendAnalytics(component, analyticsObj, uiElementProps);
```
ribeiroguilherme, line 80: What about always propagating the UIElement props to `submitAnalytics`? 
Then we do not need to override this method in the AmazonPay/ApplePay etc..

In this case, we would:

```js
this.props.modules?.analytics.sendAnalytics(component, analyticsObj, this.props);
```

And then in the analytics we would use the prop if it is available or not. We would not need to  care about adding this method overload in the express PM's

ribeiroguilherme (approved):

packages/lib/src/components/UIElement.tsx
sponglord: At the moment, instead of just throwing all of a component's props at the analytics call, all the time - I want a way to identify those components that _need_ to pass some of their props to the analytics endpoint, since I think this will be the exception rather than the rule.
If this looks to change in the future, if more components need to pass elements from their props, then I'll revisit this decision

PR #2618, opened 2024-03-26
fix: display date and time in some vouchers

m1aw:
## Summary

For some voucher like Econtext and Dragon pay we shuold display expiration time together with the date.
Created a new i18n function to format dateTime. Also make a tittle change to make sure we are not creating a new formatter every time as it's a slow operation (according MDN).

## Tested scenarios

- Tested with the mock data in the playground
- Still needs a test with the 2 payment methods test data

**Fixed issue**:  
COWEB-1297, COWEB-1359

### DIFF
--- a/packages/lib/src/components/Dragonpay/components/DragonpayVoucherResult/DragonpayVoucherResult.tsx
+++ b/packages/lib/src/components/Dragonpay/components/DragonpayVoucherResult/DragonpayVoucherResult.tsx
@@ -25,7 +25,7 @@ export default function DragonpayVoucherResult(props: DragonpayVoucherResultProp
             surcharge={surcharge && i18n.amount(surcharge.value, surcharge.currency)}
             voucherDetails={
                 [
-                    { label: i18n.get('voucher.expirationDate'), value: i18n.date(expiresAt) },
+                    { label: i18n.get('voucher.expirationDate'), value: i18n.dateTime(expiresAt) },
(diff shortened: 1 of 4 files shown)


sponglord:
How did you test that this works with expiration dates/times from different time zones?

packages/lib/src/components/Dragonpay/components/DragonpayVoucherResult/DragonpayVoucherResult.tsx
```diff
@@ -25,7 +25,7 @@ export default function DragonpayVoucherResult(props: DragonpayVoucherResultProp
             surcharge={surcharge && i18n.amount(surcharge.value, surcharge.currency)}
             voucherDetails={
                 [
-                    { label: i18n.get('voucher.expirationDate'), value: i18n.date(expiresAt) },
+                    { label: i18n.get('voucher.expirationDate'), value: i18n.dateTime(expiresAt) },
```
sponglord, line 28: Are we sure that the `expiresAt` will always be a date & time value? What happens if it's still just a date only?

sponglord (approved):

packages/lib/src/components/Dragonpay/components/DragonpayVoucherResult/DragonpayVoucherResult.tsx
m1aw: It will always be there since the backend is sending a ISO date. It could be wrong but then it's something on backend/acquiring side.

PR #2619, opened 2024-03-26
[WIP] v6 - Loading translations from CDN

ribeiroguilherme:
WIP

## Tested scenarios

**Fixed issue**:


packages/lib/src/core/core.ts
```diff
@@ -121,22 +123,40 @@ class Core implements ICore {
         return Promise.resolve(this);
     }
 
+    private async fetchLocaleTranslations(): Promise<Translations> {
+        try {
+            const translation = await getTranslations(
```
sponglord, line 128: Small point: when passing so many arguments - it's better to clean up the function signature a bit by passing a single object and extract the props in `get-translations.ts`

packages/lib/src/core/Errors/utils.test.ts
```diff
@@ -3,9 +3,10 @@ import { mapFieldKey as mapFieldKeyCC } from '../../components/Card/components/C
 import Language from '../../language';
 import { SortErrorsObj } from './types';
 import { sortErrorsByLayout } from './utils';
+import enUS from '../../../../server/translations/en-US.json';
 
 describe('Tests for Errors/utils', () => {
-    const i18n = new Language('en-US');
+    const i18n = new Language({ locale: 'en-US', translations: enUS });
```
sponglord, line 9: This could just be `const i18n = global.i18n;`
...since you've done such a nice job tidying up! :-)

packages/lib/src/components/internal/OpenInvoice/utils.test.ts
```diff
@@ -62,7 +63,7 @@ describe('OpenInvoice utils', () => {
         });
     });
     describe('mapFieldKey functionality', () => {
-        const i18n = new Language('en-US');
+        const i18n = new Language({ locale: 'en-US', translations: enUS });
```
sponglord, line 66: `const i18n = global.i18n` would work here too

packages/lib/package.json
```diff
@@ -53,13 +53,14 @@
...
         "dist/**",
+        "translations/**",
         "LICENSE",
         "README"
     ],
     "scripts": {
         "start": "npm run dev-server",
         "story": "storybook dev --port 3020 --disable-telemetry",
-        "dev-server": "cross-env NODE_ENV=development rollup --watch --config config/rollup.dev.js",
+        "dev-server": "rollup --watch --config config/rollup.dev.js",
```
sponglord, line 63: I need this for testing securedFields locally, I believe

packages/lib/src/core/core.ts
```diff
@@ -121,22 +123,40 @@ class Core implements ICore {
...
+                this.cdnContext,
+                Core.metadata.version,
+                this.options.locale,
+                this.options._translationEnvironment,
+                this.options.translations
+            );
```
sponglord, line 134: With so many arguments - you could clean up the function signature a bit by passing a single object and extracting the props in `get-translations.ts`

sponglord (approved):

packages/lib/src/core/Errors/utils.test.ts
ribeiroguilherme: I will try

packages/lib/src/components/internal/OpenInvoice/utils.test.ts
ribeiroguilherme: I will try

packages/lib/src/core/core.component.props.test.ts
```diff
@@ -4,6 +4,11 @@ import { PaymentMethodsResponse } from '../types';
 import { PayPal, Card, Dropin, Redirect } from '../components';
 import { CoreConfiguration } from './types';
 
+import getTranslations from './Services/get-translations';
+jest.mock('./Services/get-translations');
+const mockedGetTranslations = getTranslations as jest.Mock;
+mockedGetTranslations.mockResolvedValue({});
+
```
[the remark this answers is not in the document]
ribeiroguilherme: Not sure since this is not simply an object that is used around, but it is actually mocking an import . 
Perhaps it should be possible to mock it for all tests, although sometimes it is nice to make things more explicit.. I will take a look if it can be improved

packages/lib/src/core/core.ts
ribeiroguilherme: I prefer to be more strict and pass each prop individually in this case.

Otherwise I'd end up passing the whole `CoreConfiguration` object to the function without any reason. Besides that, I'd have to add 'similar' piece of code in the `get-translations` function itself, to extract the values correctly

PR #2628, opened 2024-04-03
feat: remove the country flag in the Phone select, change the label for country/region

longyulongyu:
## Summary
Remove the country flag in the Phone select, change the label to `Country/Region` for the country/region field.

## Tested scenarios
Fixed unit tests

**Fixed issue**:  COWEB-1346


packages/lib/src/components/internal/PhoneInput/PhoneInput.tsx
```diff
@@ -61,8 +61,6 @@ export function PhoneInput(props) {
                                 />
 
                                 <div className="adyen-checkout__phoneNumber">
-                                    <div>{data.phonePrefix}</div>
```
longyulongyu, line 64: The reasons to remove this line are:
- We already show the phone prefix as the selected label due to the changes in `packages/lib/src/components/QiwiWallet/utils.ts`, showing another `phonePrefix` is a duplication and doesn't look good.
- `packages/lib/src/components/internal/PhoneInput/PhoneInput.tsx` is only used by qiwi wallet, so it's safe to remove the line as well

m1aw (approved):

PR #2629, opened 2024-04-03
fix: show the instant payments when they are the only supported PMs

longyulongyu:
## Summary
Fix the bug that the drop-in is not rendered when there are only instant payments.

## Tested scenarios
To reproduce:
- For the manual flow, pass the following to the checkout config
```javascript
paymentMethodsResponse: {
            paymentMethods: [
                {
                    configuration: {
                        merchantId: '12345678',
                        gatewayMerchantId: 'testMerchant'
                    },
                    name: 'Google Pay',
                    type: 'paywithgoogle'
                }
            ]
        },
```
- Specify the `instantPaymentTypes: ['paywithgoogle']` in the drop-in config
- The drop-in is not rendered

### DIFF
--- a/packages/lib/src/components/Dropin/components/DropinComponent.tsx
+++ b/packages/lib/src/components/Dropin/components/DropinComponent.tsx
@@ -114,6 +114,7 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
     render(props, { elements, instantPaymentElements, status, activePaymentMethod, cachedPaymentMethods }) {
         const isLoading = status.type === 'loading';
         const isRedirecting = status.type === 'redirect';
+        const hasPayments = elements?.length > 0 || instantPaymentElements?.length > 0;
 
         switch (status.type) {
             case 'success':
@@ -130,7 +131,7 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
                     <div className={`adyen-checkout__dropin adyen-checkout__dropin--${status.type}`}>
                         {isRedirecting && status.props.component && status.props.component.render()}
                         {isLoading && status.props && status.props.component && status.props.component.render()}
-                        {elements && !!elements.length && (
+                        {hasPayments && (
                             <PaymentMethodList
                                 isLoading={isLoading || isRedirecting}
                                 isDisablingPaymentMethod={this.state.isDisabling}
--- a/.changeset/smart-tables-cry.md
+++ b/.changeset/smart-tables-cry.md
@@ -1,0 +1,5 @@
+---
+"@adyen/adyen-web": patch
+---
+
+Fix the bug that the drop-in is not rendered when there are only instant payments. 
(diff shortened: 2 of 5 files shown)


packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -114,6 +114,7 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
     render(props, { elements, instantPaymentElements, status, activePaymentMethod, cachedPaymentMethods }) {
         const isLoading = status.type === 'loading';
         const isRedirecting = status.type === 'redirect';
```
ribeiroguilherme, line 116: Shouldn't we also check for `storedPaymentElements` as done here? 
https://github.com/Adyen/adyen-web/blob/v6/packages/lib/src/components/Dropin/components/DropinComponent.tsx#L131

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -130,7 +131,7 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
                     <div className={`adyen-checkout__dropin adyen-checkout__dropin--${status.type}`}>
                         {isRedirecting && status.props.component && status.props.component.render()}
                         {isLoading && status.props && status.props.component && status.props.component.render()}
-                        {elements && !!elements.length && (
```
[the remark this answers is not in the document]
longyulongyu: Sure will add.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
ribeiroguilherme: If I am not mistaken they are part of the Component state. We set them here https://github.com/Adyen/adyen-web/blob/v6/packages/lib/src/components/Dropin/components/DropinComponent.tsx#L36

longyulongyu: The link you shared is from v6 😄, we put stored and regular payments together in v5: https://github.com/Adyen/adyen-web/blob/9902c457f47549623b744c2428a1f44b445cf96d/packages/lib/src/components/Dropin/components/DropinComponent.tsx#L32

ribeiroguilherme: Ouch!  Then you are totally correct 😄

packages/lib/src/components/Dropin/components/DropinComponent.tsx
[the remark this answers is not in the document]
longyulongyu: Added tests. Still need to mock something because of this line https://github.com/Adyen/adyen-web/blob/ffd80012b8d6495f82aadf8e22d646552ea99b50/packages/lib/src/components/Dropin/components/DropinComponent.tsx#L30
It's waiting for promise to resolve, otherwise the `instantPaymentElements` array stays empty.

PR #2637, opened 2024-04-08
Feature/v6 analytics with 3DS2 error events

sponglord:
## Summary
- Improvements to the 3DS2 Fingerprint & Challenge processes in terms of analysing the structure of received data in order to provide finer granularity in detecting when this data is malformed.
- At each fail point we now submit an analytics `error` event (and in the Challenge process also provide a corresponding UI error).
- No longer call the `onError` callback within the fingerprinting flow (this was a cause of misunderstanding with merchants who were treating all errors as 'fatal'). However it is still possible to call this callback when in the MDFlow (where we have control over what it does).
- Have also aligned the processes within both components so that they follow a similar way of doing things
- Improved typings

## Tested scenarios
Extensive manual checking to see that the expected analytics are generated

## TODO
Automate some of the manual testing

## notes for reviewers
Due to some file & constants renaming this PR looks bigger than it is.
The main functional changes are only in `PrepareFingerprint3DS2.tsx` and `PrepareChallenge3DS2.tsx`


packages/lib/src/components/ThreeDS2/components/Challenge/types.ts
```diff
@@ -18,10 +18,16 @@ export interface DoChallenge3DS2State {
 export interface PrepareChallenge3DS2Props extends ThreeDS2ChallengeConfiguration {
     onComplete?: (data: ChallengeResolveData) => void;
     onSubmitAnalytics: (aObj: SendAnalyticsObject) => void;
+    isMDFlow: boolean;
 }
 
 export interface PrepareChallenge3DS2State {
-    challengeData?: ChallengeData;
+    challengeData?: ChallengeData | ErrorObject;
     status?: string;
```
longyulongyu, line 26: I think we could even narrow down the status to `status?: 'performingChallenge' | 'error' | 'init' | 'complete';`

packages/lib/src/components/ThreeDS2/components/DeviceFingerprint/PrepareFingerprint3DS2.tsx
```diff
@@ -55,18 +64,90 @@ class PrepareFingerprint3DS2 extends Component<PrepareFingerprint3DS2Props, Prep
...
+                }
+            );
+
+            console.debug('### PrepareFingerprint3DS2:: token does not exist or could not be base64 decoded &/or JSON.parsed');
+        }
     }
 
-    setStatusComplete(resultObj: ResultObject) {
+    setStatusComplete(resultObj: ResultObject, errorCodeObject: ErrorCodeObject = null) {
         this.setState({ status: 'complete' }, () => {
```
longyulongyu, line 151: Just out of curiosity, why writing it inside the `setState` callback?

packages/lib/src/core/Analytics/analyticsPreProcessor.ts
```diff
@@ -17,13 +17,18 @@ import {
...
-import { THREEDS2_ERROR, THREEDS2_FULL } from '../../components/ThreeDS2/config';
+import { THREEDS2_ERROR, THREEDS2_FULL } from '../../components/ThreeDS2/constants';
+import AdyenCheckoutError, { SDK_ERROR } from '../Errors/AdyenCheckoutError';
 
 export const analyticsPreProcessor = (analyticsModule: AnalyticsModule) => {
     // return function with an analyticsModule reference
     return (component: string, analyticsObj: SendAnalyticsObject, uiElementProps = {} as any) => {
         const { type, target } = analyticsObj;
 
+        if (type == null) {
```
longyulongyu, line 28: Maybe just `if (!type) ` since `type` is string type, it can be an empty string, `''==null` will evaluate to `false`.

longyulongyu (approved):

packages/lib/src/components/ThreeDS2/components/DeviceFingerprint/PrepareFingerprint3DS2.tsx
sponglord: Good question. Don't remember -it's just always been done this way. I presume there's some subtle difference in timing - doing it this way means you change state, _then_ run the callback, and _then_ do the re-render (triggered by the fact that state has changed)?

packages/lib/src/components/ThreeDS2/components/Challenge/types.ts
sponglord: Yes, indeed

PR #2667, opened 2024-04-26
v6 - Enabling typescript strict mode 

ribeiroguilherme:
## Summary
- Adjusted `tsconfig.json` to use `strict: true` and fixed some lint issues on try-catch blocks

We can start enforcing the strict rules by tackling the deactivated rules from tsconfig file one by one (Example: we can remove a rule, then fix the lint issues and create a PR for that specific fix)

### DIFF
--- a/packages/lib/src/components/CashAppPay/components/CashAppComponent.tsx
+++ b/packages/lib/src/components/CashAppPay/components/CashAppComponent.tsx
@@ -66,7 +66,8 @@ export function CashAppComponent({
 
             setStatus('ready');
         } catch (error) {
-            onError(error);
+            if (error instanceof AdyenCheckoutError) onError(error);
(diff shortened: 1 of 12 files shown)


packages/lib/src/components/CashAppPay/components/CashAppComponent.tsx
```diff
@@ -66,7 +66,8 @@ export function CashAppComponent({
 
             setStatus('ready');
         } catch (error) {
```
sponglord, line 68: so should this be `error:unknown`?

sponglord (approved):

packages/lib/src/components/CashAppPay/components/CashAppComponent.tsx
ribeiroguilherme: I will add it in my follow up pr

PR #2671, opened 2024-05-02
v6 - Removing 'auto' from 'sideEffects' config

ribeiroguilherme:
## Summary
- Bumping to alpha7
- Removing auto files from `sideEffect` array
- Removed PURE tag from core.registry (it seems to not be needed)


longyulongyu (approved):
I ran it locally, it looks good to me!

PR #2673, opened 2024-05-07
feat: add Upi intent flow

longyulongyu:
## Summary
UPI with the `upi_intent` flow: 

For mobile:
 - We should show upi_collect or upi_intent depending on if `apps` are returned in /paymentMethods response
If the `apps` list is returned, we show them in a radio group, we always show the vpa input as the last option.
If the `apps` list is not returned or empty, we fall back to the old vpa flow.
- Submitting one of the `upi_intent` apps triggers `await` action. We should redirect the shopper to the corresponding apps. The redirect url comes from the BE action response.
- The upi_qr should always be on the second tab

 For non-mobile:
 - We should never show the upi_intent (ignore `apps` in /paymentMethods response)
 - The upi_qr should be on the first tab and the upi_collect should be on second tab

For UPI in the drop-in:
- The drop-in component shows only one upi component if the parent `upi` tx_variant is found.

## Test UPI component locally
- Run Storybook locally and go to https://localhost:3020/?path=/story/components-upi--upi
- Toggle the device tool bar to see different UI, refer to the `Summary` section for more details

## Tested scenarios
Added tests

**Fixed issue**:  COWEB-1353


ribeiroguilherme (approved):

packages/lib/storybook/stories/components/UPI.stories.tsx
```diff
@@ -18,9 +18,22 @@ export const UPI: UpiStory = {
...
+                },
+                {
+                    id: 'gpay',
+                    name: 'Google Pay'
+                },
+                {
+                    id: 'phonepe',
+                    name: 'Phone Pe'
+                }
+            ]
```
ribeiroguilherme, line 36: Can this be removed, since the Backend now returns the 'apps'?

ribeiroguilherme (approved):
Seems to be working well. Looks good 💯

packages/lib/storybook/stories/components/UPI.stories.tsx
longyulongyu: Will do.

packages/lib/src/components/UPI/types.ts
```diff
@@ -2,18 +2,35 @@ import { UIElementProps } from '../types';
...
+    UpiCollect = 'upi_collect',
+    UpiQr = 'upi_qr',
+    UpiIntent = 'upi_intent'
+}
+
+export type App = { id: string; name: string; type?: TX_VARIANT };
+
 export interface UPIElementProps extends UIElementProps {
-    defaultMode: UpiMode;
+    defaultMode?: UpiMode;
```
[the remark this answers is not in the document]
longyulongyu: > It is better to go with Template Literal Types rather than `enum` when exposing this as configuration.
> 
> ```js
> defaultMode: 'vpa' | 'qrCode' | 'intent'
> ```
> 
> By using enums, we force merchants to import the enum itself in order to configure the Component, which is not ideal.

The only problem is we are using the `UpiMode` in multiple places, it's a bit difficult to do the code changes as i need to replace them with the string literal. 
Since we were already declaring `enum` for `UpiMode` for old upi flows anyway, i think we can keep it for v5, and i can do it for v6.

PR #2680, opened 2024-05-13
v5 Adding subType to 3DS2 analytics events

sponglord:
## Summary
3DS2 log events have new, `subType` property, describing the nature of the event e.g. `fingerprintIframeLoaded` or `fingerprintCompleted`
Whilst the `fingerprintCompleted` also comes with a new `result` property describing the type of completion e.g. `success`, `failed`, `timeout` etc

This is a cherry-pick of the functionality from the v6 branch

## Tested scenarios
Manual testing that all expected `subTypes` & `results` are added (v6 version has actual unit tests covering this)
All other unit tests pass

### DIFF
--- a/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
+++ b/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
@@ -77,7 +77,7 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
         });
     };
 
-    setStatusComplete(resultObj) {
+    setStatusComplete(resultObj, isTimeout = false) {
         this.setState({ status: 'complete' }, () => {
             /**
              * Create the data in the way that the /details endpoint expects.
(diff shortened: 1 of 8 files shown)


packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
```diff
@@ -60,19 +60,24 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
...
         this.props.onSubmitAnalytics({
             type: THREEDS2_FULL,
-            message: msg
+            message: msg,
+            subtype: Analytics3DS2Events.CHALLENGE_DATA_SENT
         });
     };
 
-    setStatusComplete(resultObj) {
+    setStatusComplete(resultObj, isTimeout = false) {
```
longyulongyu, line 80: It feels like this function does too many things beside setting status to `complete`. 
It's not a blocker but we could put all the analytic logic inside the `componentDidUpdate` hook, and it would react to different state changes (potentially easy to extend the analytics logic in the future).

longyulongyu (approved):

PR #2688, opened 2024-05-16
v6 Allowing 3DS2 in checkoutshopper demo to work on localhost

sponglord:
## Summary
Recent changes in internal data verification within the 3DS2 components mean we don't recognise `http://` domains as valid for the related acs/issuer URLs: `threeDSMethodURL` & `acsURL`.
_However_ the checkoutshopper demo, when run on `localhost:8080` generates `http` protocols for these urls

Not wanting to have to create new config properties for this internal "edge case" - this PR adds an exception to our url validation rule base on the existing `environment`, and new, `environmentUrls` config properties

## Tested scenarios
Manual UMD build, tested in checkoutshopper demo, at localhost:8080,

### DIFF
--- a/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
+++ b/packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
@@ -74,14 +74,20 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
         const hasChallengeData = !isErrorObject(this.state.challengeData);
 
         if (hasChallengeData) {
+            const shouldAllowHttpDomains =
+                /** Allow http urls if in development and testing against localhost:8080 */
+                (process.env.NODE_ENV === 'development' && process.env.__CLIENT_ENV__?.indexOf('localhost:8080') > -1) ||
+                /**
+                 * Allows the checkoutshopper demo on localhost:8080 to work -
+                 *  requires a configuration in localhost of environment: 'test', _environmentUrls: {api: 'http://localhost:8080/'}
+                 */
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/ThreeDS2/components/Challenge/PrepareChallenge3DS2.tsx
```diff
@@ -74,14 +74,20 @@ class PrepareChallenge3DS2 extends Component<PrepareChallenge3DS2Props, PrepareC
         const hasChallengeData = !isErrorObject(this.state.challengeData);
 
         if (hasChallengeData) {
+            const shouldAllowHttpDomains =
+                /** Allow http urls if in development and testing against localhost:8080 */
+                (process.env.NODE_ENV === 'development' && process.env.__CLIENT_ENV__?.indexOf('localhost:8080') > -1) ||
```
longyulongyu, line 79: I am not entirely sure whether we should relax the rule. Technically, the checkout shopper demo can use `https`, a different host name other than `localhost`, and a different port. If these aspects are changed, we would also need to adjust our code, which is not ideal.

sponglord: This is not a new rule. This rule has always been there - I just moved the line it was on.
Without this rule backend devs cannot do 3DS2inMDFlow tests.

The _new_ rule is concerned with the `this.props._environmentUrls?.api` check, which when it's run by developers locally, is always run on localhost:8080

longyulongyu: I missed that the first rule was already there. It's more of a concern that when `this.props._environmentUrls?.api` suddenly changed to  some other local / dev origins, we may need to adapt to it, definitely not a big deal at this moment.

longyulongyu (approved):

PR #2694, opened 2024-05-27
Click to Pay - Fixed card list sorting and single card view image size

ribeiroguilherme:
## Summary

Bugfix:
- Fixed bug where newly created Click to Pay card was taking priority of used cards in the Card list view

Improvements:
- Improved CSS specificity for the Card image size when displaying single Card (sometimes the merchant `img` CSS was affecting the size of the Card logos)

## Tested scenarios
- Added unit test to make sure the sorting is working as expected


packages/lib/src/components/internal/ClickToPay/services/utils.ts
```diff
@@ -47,21 +47,31 @@ function sortCardByLastTimeUsed(card1: ShopperCard, card2: ShopperCard) {
     return new Date(card2.dateOfCardLastUsed).getTime() - new Date(card1.dateOfCardLastUsed).getTime();
 }
 
+function sortCardByLastTimeCreated(card1: ShopperCard, card2: ShopperCard) {
```
longyulongyu, line 50: Maybe it can be simplified into one function.
```ts
function sortByDate(key: string) {
    return (card1: ShopperCard, card2: ShopperCard) => new Date(card2[key]).getTime() - new Date(card1[key]).getTime();
}
```
Usage:
```ts
function createShopperCardsList(srcProfiles: SrcProfileWithScheme[]): ShopperCard[] {
// the rest
    return [
        ...usedCards.sort(sortByDate('dateOfCardLastUsed')),
        ...unusedCards.sort(sortByDate('dateOfCardCreated')),
        ...expiredCards.sort(sortByDate('dateOfCardLastUsed'))
    ];
}
```

ribeiroguilherme: It could! 

I thought about it, but I wanted to be more strict in terms of accessing the property directly without using an possible random string. (Ex: `new Date(card2.dateOfCardCreated)`).

Issue #2554, opened 2024-02-14
`showPayButton` not working for stored ACH


jaredtbates:
**Describe the bug**
Hi there! We are using the Adyen drop-in and have an external button to trigger payment. However, the `showPayButton` property is not working when showing stored ACH payment methods. It seems to be showing the "Continue to ..." button no matter what we pass in.

**To Reproduce**

Here's a rudimentary code example:

```typescript
AdyenCheckout({
  paymentMethodsResponse: {
    paymentMethods: [
      {
        brand: null,
        brands: null,
        configuration: null,
        fundingSource: null,
        group: null,
        inputDetails: null,
        issuers: null,
        name: 'ACH Direct Debit',
        type: 'ach',
      },
    ],
    storedPaymentMethods: [
      {
        bankAccountNumber: '011000138',
        bankLocationId: '011000138',
        brand: null,
        expiryMonth: null,
        expiryYear: null,
        holderName: null,
        iban: null,
        id: 'ASDFASDFASDFASDF',
        label: null,
        lastFour: null,
        name: 'ACH Direct Debit',
        networkTxReference: null,
        ownerName: 'Jared Bates',
        shopperEmail: null,
        supportedRecurringProcessingModels: [
          'CardOnFile',
          'Subscription',
          'UnscheduledCardOnFile',
        ],
        supportedShopperInteractions: ['Ecommerce', 'ContAuth'],
        type: 'ach',
      },
    ],
  } as any,
  showPayButton: false,
}).then((checkout) => checkout.create('dropin').mount('#dropin-container'));
```

**Expected behavior**
I expect the `showPayButton` property to hide the "Continue to ..." button in the drop-in.

**Screenshots**
<img width="359" alt="image" src="https://github.com/Adyen/adyen-web/assets/2905521/db3dac41-1e32-4f1d-8bc2-c0b5d494e15f">

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Edge (Chromium)
 - Version: 121.0.2277.112

m1aw:
@jaredtbates this indeed seems to be an issue. Since it regards a few more payment methods we are going to issue a proper fix.

jaredtbates:
Hi @m1aw! Just curious, is there a plan to fix this in the next few weeks possibly? Thanks!

m1aw:
I have created a fix for this. Probably shipping this next week.

ribeiroguilherme:
[...]

PR #2696, opened 2024-05-28
Fix: ACH not respecting showPayButton

m1aw:
## Summary

ACH and some other stored payment methods do not respect `showPayButton` config.
This hopefully fixes that by moving the responsibility to the RedirectButton component.  Together with that there's a small CSS fix to ensure we don't display extra bottom margins if there's no content.

## Tested scenarios

* Set Dropin to `showPayButton` false

**Fixed issue**:  
#2554

### DIFF
--- a/packages/lib/src/components/internal/RedirectButton/RedirectButton.tsx
+++ b/packages/lib/src/components/internal/RedirectButton/RedirectButton.tsx
@@ -1,8 +1,22 @@
 import { h, Fragment } from 'preact';
 import useCoreContext from '../../../core/Context/useCoreContext';
 import { useState } from 'preact/hooks';
+import { PaymentAmount } from '../../../types';
 
-function RedirectButton({ label = null, icon = null, payButton, onSubmit, amount = null, name, ...props }) {
+// TODO this should ideally be remove but we need let prop propagate down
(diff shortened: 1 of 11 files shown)


packages/lib/src/components/internal/RedirectButton/RedirectButton.tsx
```diff
@@ -16,6 +16,10 @@ function RedirectButton({ label = null, icon = null, payButton, onSubmit, amount
         return `${i18n.get('continueTo')} ${name}`;
     };
 
+    if (!props.showPayButton) {
```
longyulongyu, line 19: I don't see the redirect button in some payments anymore in the playground dropin, such as stored cashapppay, stored ach, and wechat pay (just by checking it out and running playground). 
I think it's because not all pms are using the `RedirectButton` the same way, some PMs pass the `showPayButton` prop to the `RedirectButton` and some are not.

m1aw: Indeed that was the case. Thank you for checking.
I have added some types, it shouldn't be possible to happen again.

longyulongyu (approved):

Issue #1510, opened 2022-02-24
A way to specify payment method when mounting Dropin


geext-dk:
**Is your feature request related to a problem? Please describe.**
We have a case when our payment amount may change after initializing the Dropin. The documentation [advices ](https://docs.adyen.com/online-payments/web-drop-in/integrated-before-5-0-0#methods) to remount (unmount and mount) the widget in case it changes. But payment methods configuration (some of which also require `amount`) is specified in the global configuration only, so it also should be updated. That's why I use `AdyenCheckout.update` method to update the configuration and remount dropin automatically. Yet it introduces another problem - after remounting, the dropin doesn't open the payment method that was opened before unmounting.

Without updating the amount, Google Pay and Apple Pay show outdated amount in their widgets.

To workaround that, I get the current payment method type (`this.dropin.activePaymentMethod?.type`) and I find the needed payment method in the DOM via this selector `.adyen-checkout__payment-method--${type}` and call `.click()` to simulate choice. And all of that should be done in the `onReady` dropin hook, because when `AdyenCheckout.update` resolves dropin is not yet fully mounted. I think that's not how it should be done.

Of course it's possible I'm doing it the totally wrong way.

**Describe the solution you'd like**
Add a public method for selecting a payment method by its type

**Describe alternatives you've considered**
* Preserve selected payment option when remounting Dropin
* Add a public method or other way to update amount without remounting Dropin
* A configuration value that specifies which payment method to open upon mounting the Dropin

**Additional context**
Our dropin is integrated the old way (before 5.0.0).
Package version we use (`@adyen/adyen-web`): 5.3.1.

ankarhem:
Hey,

I have a similar problem. I think the approach klarna has for solving this problem is quite good. They allow you to lock the UI with a suspend function when performing external actions that affect the session. And then a resume function that unlocks the UI, that also refetches the internal amount in klarnas' session.

You could provide a way to patch the session via api and then mimick this behaviour in the mount / remount function. That provides a good DX which abstracts much of the logic.

camil-adyen:
Hi all!

A BETA version of Web v6 is now available on Github. Feel free to take a look and give us feedback! We aim to release the official v6 in the coming weeks. More information can be found [here](https://github.com/Adyen/adyen-web/releases/tag/v6.0.0-beta.0).

ribeiroguilherme:
Hi all,

We release v6.0.0 this week which exposes the functionality to pre-select payment method when mounting Drop-in.

[...]

PR #2701, opened 2024-06-05
v6 - Drop-in: configuring payment method to be pre-selected

ribeiroguilherme:
## Summary
- Added new configuration  property `openPaymentMethod` for Drop-in which can be used to pre-select an specific PM.

```js
new Dropin(checkout, {
  openPaymentMethod: {
    type: 'paypal',
  }
});
```

- Also squeezed in an small change for WeChat, changing default timer to 30min

## Tested scenarios
- Added tests

**Fixed issue**:  #1510


longyulongyu (approved):

chrisneven:
Hey @ribeiroguilherme, would it also be possible to support stored payment methods? Maybe to identify we can use the ID.

ribeiroguilherme:
Hi @chrisneven ,

Stored payment method is a bit more complicated, since we can have multiple payments with the same type and name.

What is your use-case?

chrisneven:
For selected payment methods, we store the name in the URL, e.g., www.example.com/checkout/payment?paymentId=Ideal. This allows us to preselect the payment method again when the user refreshes or navigates back from the payment provider.

Issue #1881, opened 2022-12-12
Submit the form by pressing the Enter key with the custom card integration 


pguilbert:
**Is your feature request related to a problem? Please describe.**
I'm doing a [custom card integration](https://docs.adyen.com/payment-methods/cards/custom-card-integration). As a user, I would like to be able to submit the form by pressing the Enter key (default browser behavior for native forms).

**Describe the solution you'd like**
Add an `onSubmit` callback (next to `onFieldValid` / `onChange`, etc...) that is called when the user presses enter in the iframe fields.

**Describe alternatives you've considered**
AFAIK, there is no other way to do this (?)

ribeiroguilherme:
Hi @pguilbert ,

Does this answer your question? 

> We do not capture "Enter" key presses on the credit card fields. We expect users to Tab between fields and when they have Tabbed onto the Pay button, pressing Enter will then validate/submit the form

from #1242

pguilbert:
Hi,

Not really. I would like to offer the user the same experience as in an html form. 
For that, the only thing that is missing is the ability to submit the form with the "Enter".

ribeiroguilherme:
Hi @pguilbert ,

The `onSubmit` callback already exists and it is used to perform the `/payments` call . It would make more sense to provide something like `onEnterPress` for example.

Regardless of that, I discussed internally with the team and we created a ticket to check this out. We have other priorities at the moment, but we will make sure to keep this ticket updated with the progress. Cheers

obouchari:
I have a custom button using components, when calling submit() the form doesn't seem to show any validation errors! Am I missing something?

ondrakub:
hi, any progress on this issue? we would also like to submit the form using enter. This is a common behavior that users are used to and would improve the experience.

sponglord:
The major release, `v6.0.0`, was launched today!
See the documentation here:
- [release notes](https://docs.adyen.com/online-payments/release-notes/#releaseNote=2024-07-24-web-componentsdrop-in-6.0.0)
- [migration guide](https://docs.adyen.com/online-payments/update-your-integration/migrate-to-web-v6/)

See the "New" section of the Release notes...
> Shoppers can now confirm the payment by pressing the Enter key...

For a `CustomCard` you will need to implement your own `onEnterKeyPressed: (activeEl, component) => {}` function and use it to validate/submit your form e.g. by triggering `click()` on your Pay button

Issue #1242, opened 2021-08-10
Pressing Enter should trigger form submission


mogelbrod:
**Describe the bug**
Pressing <kbd>Enter</kbd> while a credit card input is focused doesn't trigger submission, instead requiring the user to manually focus the submit button and press it. Is this intentional?

**To Reproduce**
1. Go to https://www.mystoredemo.io/#/
2. Add any item to the cart
3. Enter the checkout
4. Focus any of the credit card form inputs
5. Press <kbd>Enter</kbd>
6. Nothing happens

**Expected behavior**
<kbd>Enter</kbd> triggers form submission, which in turn highlights any remaining errors.

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser: Chrome
 - Version: Version 92.0.4515.131 (Official Build) (x86_64)

**Additional context**
Related issue: https://github.com/Adyen/adyen-web/issues/610

sponglord:
Hi @mogelbrod - this is intentional.
We do not capture "Enter" key presses on the credit card fields. We expect users to Tab between fields and when they have Tabbed onto the Pay button, pressing Enter will then validate/submit the form

gabrielf:
But why? HTML forms by default submit on enter. Have you experienced that this causes problems and in that case, what problems?

Would it be possible to let integrators configure this behaviour? By default it could be off but with a setting you could capture enter presses and submit the form?

PR #2704, opened 2024-06-10
[v6] Feature Enter key presses trigger validation

sponglord:
## Summary
This adds new functionality to Components - the ability to submit the PM by pressing the Enter key.

Upon the shopper pressing the Enter key, we first look for the presence of an `onEnterKeyPressed` callback (defined at `Checkout` config level), and if this is not defined, we trigger validation/submission of the PM form. 

If an `onEnterKeyPressed` callback _is_ defined - it gets passed an object containing a reference to the PM component (`UIElement`) ; and a reference to the active element (`document.activeElement`) from which the "Enter" key press was triggered

## Tested scenarios
Tested:
- Card inc. securedFields & non-securedFields (name, address, kcp); storedCards
- CustomCard
- Giftcards
- Dropin
- ANCV
- Ach
- OpenInvoices, Affirm
- Vouchers, Bacs

New unit tests added

**Relates to issue**:  COWEB-1181, #1881, #1242


packages/lib/src/components/Dropin/Dropin.test.ts
```diff
@@ -217,4 +219,24 @@ describe('Dropin', () => {
             expect(dropin.props.openFirstStoredPaymentMethod).toBe(true);
         });
     });
+
+    describe('Detecting Enter key presses', () => {
+        test('should see merchant defined onEnterKeyPressed callback fired', done => {
+            new Dropin(checkout).mount('body');
```
longyulongyu, line 225: I think we could use the `render` utility function from the `@testing-library/preact` - `render(dropin.render());`. The reason is it cleans up the dom automatically, and more aligned with the rest of the tests.

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -178,6 +189,9 @@ abstract class BaseElement<P extends BaseElementProps> implements IBaseElement {
      * Unmounts a payment element from the DOM
      */
     public unmount(): this {
+        // Remove listener
+        off(this._node, 'keypress', this.handleKeyPress);
```
longyulongyu, line 193: Could the secured field remove all the event listeners when unmounting itself?

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -384,6 +395,31 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+     * @param obj
+     */
+    protected onEnterKeyPressed(obj: OnKeyPressObj, compRef = this) {
+        // Add component ref here, rather than in handleKeyPress in case this function has been called directly from a securedField
+        obj.component = compRef;
+
+        if (this.props.onEnterKeyPressed) {
+            this.props.onEnterKeyPressed(obj);
+        } else {
+            compRef?.payButtonRef?.buttonElRef?.focus();
```
longyulongyu, line 418: Couldn't we call `this.submit();` directly when the `showPayButton` is `true` 🤔?

ribeiroguilherme:
Regarding a11y, taking the following Drop-in scenario into account: 
1. Card payment is open and active
2. Shopper presses tab to navigate to another payment method in the payment method list, therefore making another payment method focused
3. Shopper presses Enter to open the payment method, then Card component attempts to submit the payment and focus is shifted to the Card fields

Is that the desired behavior? Or should the focused payment method be opened instead?

ribeiroguilherme:
Another a11y remark - I noticed that for Card component, we are properly shifting the focus to the first invalid field once the Pay button is pressed 👍  On some of the other PM's, that does not seem to be the case.

I tested with Giftcards, ACH, UPI, and the focus stays in the Pay button. For Affirm for example, it seems works fine.

packages/lib/src/components/internal/BaseElement/BaseElement.ts
sponglord: This listener doesn't apply to securedFields - this is for UIElements within Checkout

packages/lib/src/components/Dropin/Dropin.test.ts
sponglord: `render(dropin.render())` didn't work for me. The element, within the unit test, didn't fully render - so I could not access the element that I needed to, within the PM

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: No, for reasons I tried to explain in the Tech talk, last week. (It's probably easier to demo the "why" rather than explain it!)

sponglord:
> Another a11y remark - I noticed that for Card component, we are properly shifting the focus to the first invalid field once the Pay button is pressed 👍 On some of the other PM's, that does not seem to be the case.
> 
> I tested with Giftcards, ACH, UPI, and the focus stays in the Pay button. For Affirm for example, it seems works fine.

This is correct! Full SRPanel integration has only been added to the Card and OpenInvoices PMs. It's a long standing TODO

sponglord:
> Regarding a11y, taking the following Drop-in scenario into account:
> 
> 1. Card payment is open and active
> 2. Shopper presses tab to navigate to another payment method in the payment method list, therefore making another payment method focused
> 3. Shopper presses Enter to open the payment method, then Card component attempts to submit the payment and focus is shifted to the Card fields
> 
> Is that the desired behavior? Or should the focused payment method be opened instead?

This is a good question and needs to be investigated further

packages/lib/src/components/Dropin/Dropin.test.ts
longyulongyu: Maybe we need to wait for the pay button to be presented.

packages/lib/src/components/internal/UIElement/UIElement.tsx
ribeiroguilherme: This would also benefit merchants that are programmatically calling the `component.submit()`, no?

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -8,6 +8,7 @@ import type { ICore } from '../../../core/types';
 import type { BaseElementProps, IBaseElement } from './types';
 import type { PaymentData } from '../../../types/global-types';
 import type { AnalyticsInitialEvent, SendAnalyticsObject } from '../../../core/Analytics/types';
+import { off, on } from '../SecuredFields/lib/utilities/dom';
```
ribeiroguilherme, line 11: We could move these functions out of the SecuredFields folder and place somewhere more top-level, since they are being used not only in the secured fields context anymore.

packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -184,6 +185,10 @@ class DropinElement extends UIElement<DropinConfiguration> implements IDropin {
         this.dropinRef.closeActivePaymentMethod();
     }
 
+    protected onEnterKeyPressed(obj: OnKeyPressObj) {
+        super.onEnterKeyPressed(obj, this.activePaymentMethod || this);
+    }
```
ribeiroguilherme, line 190: I think `super.onEnterKeyPress` might not be needed here. You could call the `this.activePaymentMethod.onEnterKeyPress` directly

sponglord: yup, good point

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -384,6 +395,31 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+     * Handle Enter key pressed either from a UIElement (in which case this function has been called from handleKeyPress)
+     * OR from a securedField (in which case this function has been called 'directly')
+     * @param obj
+     */
+    protected onEnterKeyPressed(obj: OnKeyPressObj, compRef = this) {
+        // Add component ref here, rather than in handleKeyPress in case this function has been called directly from a securedField
+        obj.component = compRef;
+
+        if (this.props.onEnterKeyPressed) {
+            this.props.onEnterKeyPressed(obj);
```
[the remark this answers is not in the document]
sponglord: `action` maybe not - but I think the `fieldType` could be useful for `CustomCard`

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -384,6 +395,31 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+
+            this.debounceEnterKeyPress({ component: null, fieldType: (e.target as HTMLInputElement).name, action: 'enterKeyPressed' });
+        }
+    }
+
+    /**
+     * Handle Enter key pressed either from a UIElement (in which case this function has been called from handleKeyPress)
+     * OR from a securedField (in which case this function has been called 'directly')
+     * @param obj
+     */
```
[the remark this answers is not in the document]
sponglord: ok. Nice idea. I'll implement it

packages/lib/src/core/types.ts
```diff
@@ -275,6 +276,8 @@ export interface CoreConfiguration {
 
     onOrderRequest?: onOrderRequestCallbackType;
 
+    onEnterKeyPressed?(o: OnKeyPressObj): void;
```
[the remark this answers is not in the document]
sponglord: sure

packages/lib/src/components/Dropin/Dropin.test.ts
sponglord: I got it working it working with `render(dropin.render())` and `wiatfor` but then the event won't fire programatically!
If you want to refactor this test later be my guest!!

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: Fixed! No need for a `payButtonRef` anymore

packages/lib/src/components/internal/BaseElement/BaseElement.ts
sponglord: Was on my TODO list. Now it's done!

sponglord: Done

packages/lib/src/components/internal/UIElement/UIElement.tsx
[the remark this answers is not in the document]
ribeiroguilherme: What would be the use-case for the fieldType?

packages/lib/src/components/internal/UIElement/types.ts
```diff
@@ -100,6 +100,12 @@ export type UIElementProps = BaseElementProps &
          * @internal
          */
         paymentMethodType?: string;
+
+        /**
+         * Called when a Component detects, or is told by a SecuredField, that the Enter key has been pressed.
+         * - merchant set config option
+         */
+        onEnterKeyPressed?: (elements: OnKeyPressedObject) => void;
```
ribeiroguilherme, line 108: Minor thing - perhaps we can change the interface of this method to be better aligned with the other callbacks? (We generally pass `component` as last parameter instead of wrapping within an object). Something like:

```js
onEnterKeyPressed?: (activeInputElement: Element, component: UIElement) => void;
```

ribeiroguilherme (approved):
Good stuff. This is a great feature for v6 👏

packages/lib/src/components/internal/UIElement/types.ts
sponglord: ok

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -383,9 +385,31 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+            this.onEnterKeyPressed({ component: this, activeElement: document?.activeElement });
+        }
+    }
+
+    /**
+     * Handle Enter key pressed from a UIElement (called via handleKeyPress)
+     * @param obj
+     */
+    protected onEnterKeyPressed(obj: OnKeyPressedObject) {
+        console.log('### UIElement::onEnterKeyPressed::obj ', obj);
```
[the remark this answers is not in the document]
sponglord: yup

PR #2705, opened 2024-06-10
[V6] use bento alias tokens

longyulongyu:
## Summary
- use the `token` sass function to generate css variables with default value sourcing from the bento

## Tested scenarios
Unit tests passed


ribeiroguilherme (approved):
I tested locally -  The UI looks consistent 👍

PR #2711, opened 2024-06-12
fix google pay catch error message

strublos:
## Summary
google pay cancel event propagate [object object] instead of statusMessage

## Tested scenarios
catched error is string not [object object]

### DIFF
--- a/packages/lib/src/components/GooglePay/GooglePay.tsx
+++ b/packages/lib/src/components/GooglePay/GooglePay.tsx
@@ -73,9 +73,9 @@ class GooglePay extends UIElement<GooglePayProps> {
             })
             .catch((error: google.payments.api.PaymentsError) => {
                 if (error.statusCode === 'CANCELED') {
-                    this.handleError(new AdyenCheckoutError('CANCEL', error.toString(), { cause: error }));
+                    this.handleError(new AdyenCheckoutError('CANCEL', error.statusMessage.toString(), { cause: error }));
                 } else {
-                    this.handleError(new AdyenCheckoutError('ERROR', error.toString(), { cause: error }));
+                    this.handleError(new AdyenCheckoutError('ERROR', error.statusMessage.toString(), { cause: error }));
                 }
             });
     };


ribeiroguilherme:
@strublos thanks for contributing to the repository.

Can you share step-by-step how to get the  `[object object]` as part of the error message?
I am not managing to get this on my environment. Ex: When closing the pop-up, I get proper error message in the `onError` callback.

strublos:
@ribeiroguilherme Hi I not able to reproduce this error, but we are logging these type of error and get only `[object object]`. If you look at `interface PaymentsError` there is statusMessage, so log error with error.toString() is nonsense, it is object.

ribeiroguilherme:
Hi @strublos ,

Sorry for the delay getting back to you.

We didn't really manage to reproduce the error being logged as `[object object]` . We tested by closing/cancelling the GooglePay pop-up,  and we are getting the `CANCEL AbortError: User closed the Payment Request UI` message.

I also tested applying your proposed changes but the `statusMessage` message is being returned `undefined`  (at least in the CANCEL scenario), so this fix would not work unfortunately.

In this case, I can suggest you to access the error object that comes within the `cause ` property. You can access it and perhaps report back the `error.message` and see if it works for you.

Thanks for reaching out!

strublos:
Hi @ribeiroguilherme ,
you have right, all object structure is in cause property. But it is very short only `{"errorMessage":{"statusCode": "CANCELED"}} `
We have information from customers, that they are not able pay with google pay, but had no clue, where was the problem and they were confused. It is possible get more information about this behaviour?

ribeiroguilherme:
@strublos I don't think that it is possible to get more information than that. 

If you got "CANCELED", that means the shopper closed the Google Pay payment sheet without doing the transaction. I am not sure what information you are looking for besides this one

strublos:
@ribeiroguilherme  Yes, thats is true, transaction is not finished. But we have feedback from customers, that is not possible  finish transaction via google pay. So, I don`t think so, they close payment dialog themselves, there have to be another problem. If is dialog closed by user, there is different message in log.

PR #2716, opened 2024-06-18
[v6] card input has custom hook for sr panel

sponglord:
## Summary
Similar to how `OpenInvoice' works, the `CardInput` component now uses a custom hook for setting and clear errors in the `SRPanel`

## Tested scenarios
All tests pass

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/useSRPanelForCardInputErrors.ts
+++ b/packages/lib/src/components/Card/components/CardInput/useSRPanelForCardInputErrors.ts
@@ -9,27 +9,24 @@ import { setFocusOnFirstField } from './handlers';
 import { getArrayDifferences } from '../../../../utils/arrayUtils';
 
 const useSRPanelForCardInputErrors = ({ errors, props, isValidating, retrieveLayout, specifications, billingAddress, sfp }) => {
-    //
-    /** SR stuff */
+    // Extract fns from context
     const { setSRMessagesFromObjects, setSRMessagesFromStrings, clearSRPanel, shouldMoveFocusSR } = useSRPanelContext();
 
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/Card/components/CardInput/useSRPanelForCardInputErrors.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const [sortedErrorList, setSortedErrorList] = useState<SortedErrorObject[]>(null);
+
+    // Get the previous list of errors
+    const previousSortedErrors = usePrevious(sortedErrorList);
+
+    const sfStateErrorsObj = sfp.current?.mapErrorsToValidationRuleResult();
+    const mergedErrors = { ...errors, ...sfStateErrorsObj };
+
+    useEffect(() => {
```
longyulongyu, line 30: Maybe one improvement we could do is to wrap everything inside try catch block. So that any exceptions inside here won't block card rendering.

To demo we can try this code, it blocks rendering.
```js
    useEffect(() => {
            // the rest of the code
            throw new Error('dummy error');
    }, [errors]);
```
However, if we handle the error, it doesn't block rendering.
```js
    useEffect(() => {
        try {
            // all the code
            throw new Error('dummy error');
        } catch (e) {
// do nothing or send to analytics or log as warnings
            console.log({ e });
        }
    }, [errors]);
```

longyulongyu (approved):

packages/lib/src/components/Card/components/CardInput/useSRPanelForCardInputErrors.ts
sponglord: I'l add it as a defensive-coding measure

PR #2723, opened 2024-06-24
[v6] Fix(klarna): remove token for the makeAdditionalDetailsCall

longyulongyu:
## Summary
- Remove `token` for the makeAdditionalDetailsCall
- Add the `klarna_b2b` back

## Tested scenarios
All tests passed and we are able to submit payments successfully.

### DIFF
--- a/packages/lib/src/components/components-map.ts
+++ b/packages/lib/src/components/components-map.ts
@@ -182,6 +182,7 @@ export const ComponentsMap = {
     [TxVariants.klarna]: Klarna,
     [TxVariants.klarna_account]: Klarna,
     [TxVariants.klarna_paynow]: Klarna,
+    [TxVariants.klarna_b2b]: Klarna,
     /** Klarna */
 
     /** QRLoader */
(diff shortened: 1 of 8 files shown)


packages/lib/src/components/components-map.ts
```diff
@@ -182,6 +182,7 @@ export const ComponentsMap = {
     [TxVariants.klarna]: Klarna,
     [TxVariants.klarna_account]: Klarna,
     [TxVariants.klarna_paynow]: Klarna,
+    [TxVariants.klarna_b2b]: Klarna,
```
ribeiroguilherme, line 185: Was `klarna_b2b` a redirect component before?

ribeiroguilherme (approved):

packages/lib/src/components/components-map.ts
longyulongyu: No I think it should link to `Klarna` component ([v5](https://github.com/Adyen/adyen-web/blob/aa18474e2788adb8a86abbe853a4904ee2e31d28/packages/lib/src/components/index.ts#L201)).

PR #2724, opened 2024-06-24
Feature/v5 card adds analytics config data

sponglord:
## Summary
Adding a `configData` object to the  `info` event (`type:'rendered')` for `/checkoutanalytics`
This object will contain details of the configuration that the component has been set up with.
We're starting out with the config for the `Card` component, but will later expand it to `Dropin`, other components and even some of the top level `Checkout` configuration

## Tested scenarios
Unit tests added for the generation of card related `configData`

**Relates to issue**:  COWEB-1351 (checkoutanalytics mvp)


packages/lib/src/core/Analytics/utils.ts
```diff
@@ -63,3 +70,133 @@ export const processAnalyticsData = (analyticsData: AnalyticsData): AnalyticsDat
         return acc;
     }, {});
 };
+
+export const getCardConfigData = (cardProps: CardElementProps): CardConfigData => {
+    // Extract props from cardProps - mostly setting a default value, if prop not found
+    const {
+        autoFocus = CardDefaultProps.autoFocus,
```
longyulongyu, line 77: Why not using the `cardProps` directly? I believe the Card element should already merged the default values with the user defined value (if provided)?

packages/lib/src/components/SecuredFields/SecuredFields.tsx
```diff
@@ -40,6 +41,10 @@ export class SecuredFieldsElement extends UIElement {
         };
     }
 
+    protected submitAnalytics(analyticsObj: SendAnalyticsObject) {
```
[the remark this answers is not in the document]
sponglord: This has come up before...
For now I think this is the exception (passing all the props) rather than the rule - so I've been explicitly extending the `submitAnalytics` function so we can keep a track of where we are doing it (basically in Wallet & Card related comps.
If it turns out that we end up doing this more often (so it becomes the rule) - then we can make the refactor you suggest

packages/lib/src/core/Analytics/types.ts
```diff
@@ -120,3 +122,55 @@ export type FieldErrorAnalyticsObject = {
...
+    showBrandIcon: boolean;
+    showBrandsUnderCardNumber: boolean; // (v5 only)
+    showInstallmentAmounts: boolean;
+    showKCPType: 'none' | 'auto' | 'atStart';
+    showPayButton: boolean;
+    socialSecurityNumberMode: SocialSecurityMode;
+    srPanelEnabled: boolean;
+    srPanelMoveFocus: boolean;
+    // callbacks
+    onAllValid: boolean;
```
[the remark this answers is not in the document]
sponglord: yes

packages/lib/src/core/Analytics/utils.ts
sponglord: But all the default values haven't been set in the `Card` element. 
The default values you see being used here are actually being pulled from the `CardInput` component and are "filling in" the gaps of any missing config

sponglord: This doesn't work out the box... but I can _make_ it work

longyulongyu (approved):

PR #2732, opened 2024-07-04
Add other types of template

sponglord:


.github/ISSUE_TEMPLATE/custom.md
```diff
@@ -0,0 +1,10 @@
+---
+name: Custom issue template
```
longyulongyu, line 2: What does this file do? Feels like we don't need this one as we already have the issue template.

sponglord: Yeah, maybe. It's just a default they offer for "other" issues that are not bug reports or feature requests

longyulongyu: Indeed i think we can delete it, it's more of an Open Source team's service for other teams.

longyulongyu (approved):

PR #2733, opened 2024-07-05
[v6]Cherry pick new upi flow

longyulongyu:
## Summary
- Fixed tests
- New bento design
- Refactor some reusable components


packages/lib/src/components/Dropin/elements/createElements.ts
```diff
@@ -20,13 +20,10 @@ const createElements = (
...
 ): Promise<UIElement[]> => {
-    const elements = paymentMethods
+    const elements = optionallyFilterUpiSubTxVariants(paymentMethods)
         .map(paymentMethod => {
-            const paymentMethodConfigurationProps = getComponentConfiguration(
-                paymentMethod.type,
-                paymentMethodsConfiguration,
-                paymentMethod.isStoredPaymentMethod
-            );
+            const isStoredPaymentMethod = 'isStoredPaymentMethod' in paymentMethod && paymentMethod.isStoredPaymentMethod;
```
ribeiroguilherme, line 25: Is it needed to perform this check here? Just wondering that since it wasn't here before

packages/lib/src/components/Dropin/elements/filters.ts
```diff
@@ -1,5 +1,5 @@
 import promiseTimeout from '../../../utils/promiseTimeout';
-import { UIElement } from '../../../types';
+import { PaymentMethod, type StoredPaymentMethod, UIElement } from '../../../types';
```
ribeiroguilherme, line 2: I am not entirely sure the difference, but usually the import types is used this way:

```
import types { PaymentMethod, StoredPaymentMethod, UIElement } from '../../../types';
```

packages/lib/src/components/tx-variants.ts
```diff
@@ -143,6 +143,7 @@ export enum TxVariants {
     upi = 'upi', // also QR
     upi_qr = 'upi_qr', // also QR
     upi_collect = 'upi_collect', // also QR
+    upi_intent = 'upi_intent', // also QR
```
ribeiroguilherme, line 146: I believe the `upi_intent` is missing in the `components-map.ts` and `components-name-map.ts` , no?

ribeiroguilherme (approved):

packages/lib/src/components/Dropin/elements/filters.ts
longyulongyu: I think it's indeed a mistake, the Webstorm auto imported it. I'll fix it.

packages/lib/src/components/Dropin/elements/createElements.ts
longyulongyu: I got a TS error `TS2339: Property isStoredPaymentMethod does not exist on type PaymentMethod | StoredPaymentMethod
Property isStoredPaymentMethod does not exist on type PaymentMethod` so I added this extra check.

packages/lib/src/components/tx-variants.ts
longyulongyu: Indeed nice find, will add.

PR #2745, opened 2024-07-17
v6 - Fetching any locale passed to checkout

ribeiroguilherme:
## Summary
- Accepting any locale as part of the `AdyenCheckout` configuration
- If the request to fetch that locale fails, we fallback fetching the 'en-US' locale


longyulongyu:
I was testing a random non-compliant locale `aaa`, i saw we are preventing the sdk rendering. Do you think we should just output a warning, fallback to English and still let the sdk render? I refer to the function `formatLocale`.
The rest works and looks good!

longyulongyu (approved):

PR #2761, opened 2024-07-24
v6 release

ribeiroguilherme:
## Summary
- Bumped version
- Updated pipelines and changelog


longyulongyu (approved):

sponglord (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -2,9 +2,9 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ main, v6 ]
+    branches: [ main ]
```
sponglord, line 5: Do we not need to add v5 here now? We will still need to make v5 releases

.github/workflows/codeql-analysis.yml
```diff
@@ -2,9 +2,9 @@ name: "CodeQL"
...
   pull_request:
-    branches: [ main, v6 ]
+    branches: [ main ]
```
sponglord, line 7: same comment as above

.github/workflows/codeql-analysis.yml
ribeiroguilherme: we will do that in the v5 branch

PR #2781, opened 2024-07-31
[v5] adds new privacy link to Riverty

m1aw:
## Summary

This feature adds a new privacy link to RIverty disclaimer checkbox. 

Also in this PR we rework the way we build this string since it seems like there's different implementations for different components.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Riverty/Riverty.tsx
+++ b/packages/lib/src/components/Riverty/Riverty.tsx
@@ -22,10 +22,6 @@ export default class Riverty extends OpenInvoiceContainer {
         ...OpenInvoiceContainer.defaultProps
     };
 
-    constructor(props) {
-        super(props);
-    }
-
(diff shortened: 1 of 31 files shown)


packages/lib/src/components/Riverty/Riverty.tsx
```diff
@@ -20,15 +22,27 @@ export default class Riverty extends OpenInvoiceContainer {
         ...OpenInvoiceContainer.defaultProps
     };
 
+    constructor(props) {
```
longyulongyu, line 25: Do we need this constructor? It doesn't do anything special.

m1aw: Well, spotted. We don't it was just there because at some point it was needed during dev.

longyulongyu (approved):

PR #2794, opened 2024-08-16
Analytics config data dropin

longyulongyu:
## Summary
- For card config data, do not use `JSON.stringify` for `billingAddressAllowedCountries`, `billingAddressRequiredFields`, `brands`
- Omit the field `billingAddressAllowedCountries` when empty
- Send dropin config data: `openFirstStoredPaymentMethod` and `showStoredPaymentMethods` 

## Tested scenarios
Added unit test

**Fixed issue**:  COWEB-1406

### DIFF
--- a/packages/lib/src/core/Analytics/analyticsPreProcessor.ts
+++ b/packages/lib/src/core/Analytics/analyticsPreProcessor.ts
@@ -37,7 +37,7 @@ export const analyticsPreProcessor = (analyticsModule: AnalyticsModule) => {
             // Called from BaseElement (when component mounted) or, from DropinComponent (after mounting, when it has finished resolving all the PM promises)
             // &/or, from DropinComponent when a PM is selected
             case ANALYTICS_RENDERED_STR: {
-                const { isStoredPaymentMethod, brand } = analyticsObj;
+                const { isStoredPaymentMethod, brand, configData: originalConfigData } = analyticsObj;
 
                 // Expected from Wallet PMs
                 const { isExpress, expressPage } = uiElementProps;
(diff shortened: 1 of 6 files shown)


packages/lib/src/core/Analytics/analyticsPreProcessor.ts
```diff
@@ -59,7 +59,8 @@ export const analyticsPreProcessor = (analyticsModule: AnalyticsModule) => {
                     ...(brand && { brand }),
                     ...(typeof isExpress === 'boolean' && { isExpress }),
                     ...(isExpress === true && hasExpressPage && { expressPage }), // We only care about the expressPage value if isExpress is true
-                    ...(configData && { configData })
+                    ...(configData && { configData }),
+                    ...(originalConfigData && { configData: originalConfigData })
```
sponglord, line 63: You should add a comment here to explain what "originalConfigData" represents i.e. config data extracted from the Dropin - since it is not clear form the code itself

packages/lib/src/core/Analytics/utils.ts
```diff
@@ -140,18 +140,17 @@ export const getCardConfigData = (cardProps: CardConfiguration): CardConfigData
     }
 
     // Probably just for development - in real life we wouldn't expect the number of supported brands to push the endpoint limit on 128 chars
-    let brandsStr = JSON.stringify(brands);
+    let brandsStr = brands.toString();
     if (brandsStr.length > 128) {
-        brandsStr = brandsStr.substring(0, 124) + '...]';
+        brandsStr = brandsStr.substring(0, 124);
```
sponglord, line 145: I added the `+ '...]'` so it was clear from looking at the string that it had been deliberately truncated and wasn't accidentally malformed

longyulongyu: Hmm ok i will check with other teams, i don't think they are aware of it, and it looks difficult for the data team to program on it, because they cannot just simply use `split(',')` to split the string and get the array.

packages/lib/src/core/Analytics/analyticsPreProcessor.ts
longyulongyu: Sure.

packages/lib/src/core/Analytics/utils.ts
sponglord: ok... It's not vital, and is probably an edge case in development (that the brands array could be so big) - if it's causing difficulty for the data team then keep your change

sponglord (approved):

PR #2798, opened 2024-08-19
fix(analytic-util): convert to string when exists

longyulongyu:
## Summary
Minor fix to optionally convert array data to string.

No need to add changeset as it was a patch for #2794

### DIFF
--- a/packages/lib/src/core/Analytics/utils.ts
+++ b/packages/lib/src/core/Analytics/utils.ts
@@ -9,6 +9,7 @@ import CardInputDefaultProps from '../../components/Card/components/CardInput/de
 import { DEFAULT_CARD_GROUP_TYPES } from '../../components/internal/SecuredFields/lib/constants';
 import { notFalsy } from '../../utils/commonUtils';
 
+const MAX_LENGTH = 128;
 export const getUTCTimestamp = () => Date.now();
 
 /**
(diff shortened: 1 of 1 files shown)


packages/lib/src/core/Analytics/utils.ts
```diff
@@ -139,19 +140,17 @@ export const getCardConfigData = (cardProps: CardConfiguration): CardConfigData
...
-    let brandsStr = brands.toString();
-    if (brandsStr.length > 128) {
-        brandsStr = brandsStr.substring(0, 124);
-    }
     // @ts-ignore commenting out props until endpoint is ready
     const configData: CardConfigData = {
         autoFocus,
-        ...(billingAddressAllowedCountries?.length > 0 && { billingAddressAllowedCountries: billingAddressAllowedCountries.toString() }),
+        ...(billingAddressAllowedCountries?.length > 0 && {
+            billingAddressAllowedCountries: billingAddressAllowedCountries.toString().substring(0, MAX_LENGTH)
```
sponglord, line 147: Does this also need the optional chaining operator (`?`) after `toString()`
(You used it for `billingAddressRequiredFields` and `brands`)

sponglord (approved):

packages/lib/src/core/Analytics/utils.ts
longyulongyu: It should be fine. The engine checks `billingAddressAllowedCountries?.length > 0 ` first, if it failed, it won't execute `billingAddressAllowedCountries: billingAddressAllowedCountries.toString().substring(0, MAX_LENGTH)`, so it shouldn't cause any exceptions.

PR #2800, opened 2024-08-19
fix stored card holder name

m1aw:
## Summary

Fixes `props.holderName` not being used in stored cards as `paymetMethod.holderName`. 

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Card/Card.test.tsx
+++ b/packages/lib/src/components/Card/Card.test.tsx
@@ -1,6 +1,6 @@
 import { h } from 'preact';
 import { CardElement } from './Card';
-import { render, screen } from '@testing-library/preact';
+import { render, screen, waitFor } from '@testing-library/preact';
 import CoreProvider from '../../core/Context/CoreProvider';
 import Language from '../../language';
 import { Resources } from '../../core/Context/Resources';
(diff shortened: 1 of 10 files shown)


packages/lib/src/components/Card/Card.test.tsx
```diff
@@ -128,6 +128,47 @@ describe('Card', () => {
...
+
+            expect(card.formatData().paymentMethod.holderName).toContain('');
+        });
+
+        test('should NOT echo back holderName if is not a stored card', () => {
+            const card = new CardElement({ ...coreProps, holderName: 'Test Holder' });
+            render(card.render());
+
+            expect(card.formatData().paymentMethod.holderName).toBeUndefined();
+        });
```
sponglord, line 169: Should we add a test that if it is not a storedCard, but `holderName` is passed in a _data_ object, then we _should_ see that in the data?

sponglord (approved):

packages/lib/src/components/Card/Card.test.tsx
m1aw: Yeah, I think that should be covered too.

m1aw: Tests added.

PR #2805, opened 2024-08-21
Fix/change publish pipeline

m1aw:
## Summary

Creates a new job `Build UMD and Translations` that does exactly that. Builds the UMD bundle and bundles translations, publish the artifact. 

The job should be executed after closing the Release PR.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -1,14 +1,13 @@
 # This workflow will run tests using node and then publish a package when a release is created
 # For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
 
-name: Publish Package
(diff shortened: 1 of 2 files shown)


.github/workflows/release-build.yml
```diff
@@ -0,0 +1,38 @@
+# This workflow will run tests using node and then publish a package when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Build UMD and Translations
```
longyulongyu, line 4: Just realized that this pipeline can be triggered when the branch `changeset-release/main` merged to `main`, but we can do it later.

longyulongyu (approved):

PR #2811, opened 2024-08-23
Add missing currency information for Iceland and Bulgaria

sponglord:
## Summary
In v6 we added new locales: Iceland and Bulgaria.
This PR also adds the necessary currency information for these two, non-Euro, countries

## Tested scenarios
Manually tested in Chrome, Firefox  & Safari that currency is identified and formatted

### DIFF
--- a/packages/lib/src/utils/constants/currency-decimals.ts
+++ b/packages/lib/src/utils/constants/currency-decimals.ts
@@ -10,6 +10,7 @@ const CURRENCY_DECIMALS = {
     GHC: 1,
     GNF: 1,
     KMF: 1,
+    ISK: 1,
     PYG: 1,
     RWF: 1,
     UGX: 1,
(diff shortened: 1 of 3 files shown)


m1aw (approved):

packages/lib/src/utils/constants/currency-decimals.ts
```diff
@@ -10,6 +10,7 @@ const CURRENCY_DECIMALS = {
     GHC: 1,
     GNF: 1,
     KMF: 1,
+    ISK: 1,
```
jureperak, line 13: Hey, are you sure that ISK needs to be set as zero decimal currency? In docs it clearly states that even do ISK is zero decimal, it will be used with two decimals https://docs.adyen.com/development-resources/currency-codes/#note

I'm asking this because, after this version, when i'm using dropin with CustomPay button, amount that i provide in ApplePay configuration for ISK is not same as in the version before this release. I will try to make a test on my sandbox environment and create PR, but if you could verify this before me even better.

Issue #2362, opened 2023-10-17
Postal code validation error messages don't match with regex


ignas-seskauskas2:
Postal code validation error message doesn't show correct information. It says that 99999 format should be correct, but it's not correct, error message is still shown.

![image](https://github.com/Adyen/adyen-web/assets/103568359/6aa69544-7657-4544-9ed0-88aef163edc9)

In https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/internal/Address/validate.ts:
language LT pattern is "/^(LT-\d{5})$/"
But in https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/internal/Address/validate.formats.ts error message for language LT is: '9999 or 99999 or LT-99999'

There may be a similiar issue for other languages too

m1aw:
Hello @ignas-seskauskas2 

This issue has been fixed on the latest version `5.68.1` and `6.1.1`.

PR #2822, opened 2024-08-29
Fixes LT postal code

m1aw:
## Summary

Makes postal code validation rules same as description. Also add AddressElement to Storybook.  Adds some test to validate.

### Discussion

As I was writing the test I decided to write some additional tests for some more countries, and quickly found out that are some scenarios that we might not be validating correctly (with the validator alone). 

It's okay because these scenarios get validated correctly at the end but I left the tests in there so we can maybe decide to fix this in the future.

Also for future reference there's [this giant list from the European Central Bank](https://www.ecb.europa.eu/stats/money/aggregates/anacredit/shared/pdf/List_postal_code_formatting_rules_and_regular_expressions.xlsx) that has a pretty comprehensive list of regex and validation rules.

## Tested scenarios

* Added tests
* Tested in storybook the LT-XXXX, XXXX, XXXXX variations.

**Fixed issue**:  
#2362

### DIFF
--- a/packages/lib/src/components/internal/Address/validate.test.ts
+++ b/packages/lib/src/components/internal/Address/validate.test.ts
@@ -1,0 +1,140 @@
+import { validatePostalCode } from './validate';
+import { ValidatorRules } from '../../../utils/Validator/types';
+
+describe('validatePostalCode', () => {
+    const validatorRules: ValidatorRules = {
+        postalCode: {
+            modes: ['blur'],
+            validate: jest.fn(),
+            errorMessage: ''
+        }
+    };
+
+    // LT
+    it('should return true for a valid LT postal code without the LT prefix', () => {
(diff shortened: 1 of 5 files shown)


packages/lib/src/components/internal/Address/validate.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        const result = validatePostalCode('LT-1235', 'LT', validatorRules);
+        expect(result).toBe(false);
+    });
+
+    it('should return false for an invalid LT postal code with fewer than 4 digits', () => {
+        const result = validatePostalCode('123', 'LT', validatorRules);
+        expect(result).toBe(false);
+    });
+
+    it('should return false for an invalid LT postal code with fewer than 6 digits', () => {
```
sponglord, line 39: "with more than 6 digits" :-p

sponglord (approved):

PR #2823, opened 2024-08-30
adding storedCard story

sponglord:
## Summary
Added story for storedCard

### DIFF
--- a/packages/lib/storybook/stories/cards/Card.stories.tsx
+++ b/packages/lib/storybook/stories/cards/Card.stories.tsx
@@ -5,6 +5,7 @@ import { Card } from '../../../src';
 import { Container } from '../Container';
 import { searchFunctionExample } from '../../../../playground/src/utils';
 import { CardWith3DS2Redirect } from './cardStoryHelpers/CardWith3DS2Redirect';
+import './cardStoryHelpers/storedCard.style.scss';
 
 type CardStory = StoryConfiguration<CardConfiguration>;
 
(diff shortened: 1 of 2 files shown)


packages/lib/storybook/stories/cards/Card.stories.tsx
```diff
@@ -20,6 +21,53 @@ const createComponent = (args: PaymentMethodStoryProps<CardConfiguration>, conte
     return <Container element={card} />;
 };
 
+const createStoredCardComponent = (args: PaymentMethodStoryProps<CardConfiguration>, context) => {
+    const { componentConfiguration } = args;
+    const checkout = getStoryContextCheckout(context);
+
+    if (checkout.paymentMethodsResponse.storedPaymentMethods && checkout.paymentMethodsResponse.storedPaymentMethods.length > 0) {
+        // We are only interested in card based storedPaymentMethods that support Ecommerce  - a quick way to distinguish these is if they have a brand property
+        let storedCardData;
```
m1aw, line 30: Just an idea, could this be something we make an helper for the merchants to use? Or is this something they should do on the API before returning to it to the library?

m1aw (approved):

packages/lib/storybook/stories/cards/Card.stories.tsx
sponglord: I don't know if this is much use for a merchant. They're expected to just pass all the `storedPaymentMethods` into the sdk and we present all of them (filtered on if they support `Ecommerce` etc)

This example for the story is a contrived one because, for simplicities sake, we just want to present a single storedPM

PR #2836, opened 2024-09-06
ApplePaySession is not always available in Window

rasmusbe:
## Summary
Since it's declared globally we can not expect ApplePaySession to always be available in the Window interface.

### DIFF
--- a/.changeset/swift-emus-complain.md
+++ b/.changeset/swift-emus-complain.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Change ApplePaySession type to optional in window interface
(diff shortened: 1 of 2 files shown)


ribeiroguilherme:
Thanks @rasmusbe for the contribution. Your change does make sense 👍  I am curious to know though - did you face any issues with the way that it is typed right now?

rasmusbe:
Yes, I found it when I activated the lint rule [@typescript-eslint/no-unnecessary-condition](https://typescript-eslint.io/rules/no-unnecessary-condition/) and it warned about us checking if ApplePaySession exists in our code.

rasmusbe:
We're using 6.0.3 right now

ribeiroguilherme:
@rasmusbe can you add the changeset for your change? Then I can approve and merge it.
Run `yarn changeset` in the root, flag as 'patch', and add a description of why this change. thanks

ribeiroguilherme (approved):

PR #2837, opened 2024-09-09
Fixes regression and allows Enter key to validate/submit

sponglord:
## Summary
Fixes regression caused by #2750 and allows Enter key once again to trigger validation or submission on the Card component

## Tested scenarios
Added new e2e test

**Relates to issue**:  #2750

### DIFF
--- a/packages/lib/src/components/Card/Card.tsx
+++ b/packages/lib/src/components/Card/Card.tsx
@@ -57,7 +57,7 @@ export class CardElement extends UIElement<CardConfiguration> {
         _disableClickToPay: false,
         doBinLookup: true,
         // Merge most of CardInput's defaultProps
-        ...reject(['type', 'setComponentRef']).from(CardInputDefaultProps)
+        ...reject(['type', 'setComponentRef', 'onEnterKeyPressed']).from(CardInputDefaultProps)
     };
 
     public setStatus(status: UIElementStatus, props?): this {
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/Card/Card.tsx
```diff
@@ -57,7 +57,7 @@ export class CardElement extends UIElement<CardConfiguration> {
         _disableClickToPay: false,
         doBinLookup: true,
         // Merge most of CardInput's defaultProps
-        ...reject(['type', 'setComponentRef']).from(CardInputDefaultProps)
+        ...reject(['type', 'setComponentRef', 'onEnterKeyPressed']).from(CardInputDefaultProps)
```
ribeiroguilherme, line 60: It does not make sense to have 'onEnterKeyPress' as default props, since it is not actually a default prop and it shouldn't be. Can't we remove it from there?

ribeiroguilherme (approved):

packages/lib/src/components/Card/Card.tsx
sponglord: Remove it from where? `CardInputDefaultProps`?

ribeiroguilherme: yes

sponglord: Done

PR #2847, opened 2024-09-11
Send browser info on setup

m1aw:
## Summary

Forward browserInfo from the SDK onto the internal /setup request, so it can be passed on by the backend to the Checkout.

## Tested scenarios

**Fixed issue**:  
COWEB-1391


ribeiroguilherme (approved):
Don't forget to add the changeset and cherry-pick

PR #2850, opened 2024-09-11
Removes deprecated request package from internal server

m1aw:
## Summary
Basically what the title says, we had some security warnings about the deprecated package `request`. It has been deprecated for 5 years, so I think is a good moment to refactor it for something more modern.

## Tested scenarios

- Played around in the playground

**Fixed issue**:

### DIFF
--- a/packages/server/api/ordersCancel.js
+++ b/packages/server/api/ordersCancel.js
@@ -1,4 +1,4 @@
-const makePostRequest = require('../utils/getPostParameters');
+const makePostRequest = require('../utils/makePostRequest');
 const handleCallback = require('../utils/handleCallback');
 const { MERCHANT_ACCOUNT: merchantAccount } = require('../utils/config');
 


packages/server/api/originKeys.js
```diff
@@ -1,10 +1,8 @@
...
 
-module.exports = (res, request) => {
+module.exports = async (res, request) => {
     const originDomains = [`${request.protocol}://${request.headers.host}`];
-    const params = getPostParameters('originKeys', { originDomains });
-
-    post(params, (err, response, body) => handleCallback({ err, response, body }, res));
+    const response = await makePostRequest('originKeys', { originDomains });
+    handleCallback(response, res);
 };
```
ribeiroguilherme, line 8: Let's delete this endpoint. Nobody is using that currently

ribeiroguilherme (approved):

PR #2855, opened 2024-09-13
v6 - Fixing styles import path for Webpack4 projects

ribeiroguilherme:
## Summary
Webpack 4 does not use the 'exports' field in the 'package.json' for the module resolution. This leads to projects using webpack4  not being able to import the styles from `@adyen/adyen-web/styles/adyen.css` . 
The temporary solution is to import directly using the file path `@adyen/adyen-web/dis/es/adyen.css`.

To solve that, the prepublish pipeline was adjusted to create a folder 'styles' and then push the CSS file there. 
By doing that, Webpack 4 projects can directly import by doing  `@adyen/adyen-web/styles/adyen.css` , as it is a valid file path.

## What was done:
- Added the 'styles' folder to the package.json 'files' entry, so it gets added to the tarball file
- In the prepublish script, the project is build and then the styles folder is created and the file is placed there

### DIFF
--- a/.changeset/rotten-flies-bathe.md
+++ b/.changeset/rotten-flies-bathe.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixing adyen.css file exposure for Webpack 4 projects
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,6 +1,3 @@
-# This workflow will run tests using node and then publish a package when a release is created
-# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
-
 name: Publish Package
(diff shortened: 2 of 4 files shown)


d4nielchaves:
I'm encountering the exactly same issue with a brand new Vite (`vite/5.4.5`) project.

m1aw (approved):

ribeiroguilherme:
Hi @d4nielchaves. We are happy to investigate further with you provide a reproducible example.

Here you can see brand new Vite (5.4.5) without the issue you mentioned: https://stackblitz.com/edit/vitejs-vite-7p5a4f?file=main.js

d4nielchaves:
@ribeiroguilherme I could successfully proof your sandbox. The issue is on my side. Thank you so much for your kind attention.

PR #2861, opened 2024-09-18
Whitelabel Pay by Bank in the US

m1aw:
## Summary

Make the Pay by Bank solution in the US whitelabel instead of using branding specific open banking provider.

This PR also adds the ability to make brands always visible in Dropin, even when the payment method is selected.

The PR is waiting translations and logo changes but for now is ready for review.

## Tested scenarios

**Fixed issue**:


packages/lib/src/components/PayByBankUS/PayByBankUS.tsx
```diff
@@ -0,0 +1,75 @@
...
+            // paymentMethodBrands configuration
+            keepBrandsVisible: true,
+            showOtherInsteafOfNumber: true,
+            ...props
+        };
+    }
+
+    get displayName() {
+        return this.props.name || this.constructor['type'];
+    }
```
ribeiroguilherme, line 33: Why do we need to override the `this.displayName` that is defined in the `UIElement`?  Can't we reuse it?

packages/lib/src/components/PayByBankUS/PayByBankUS.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    /*
+    Hardcode US brands 
+    */
+    get brands(): { icon: any; name: string }[] {
```
ribeiroguilherme, line 38: Can we be more strict here and add proper type instead of  `any`?

packages/lib/src/components/PayByBankUS/PayByBankUS.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    /* TODO
+    DONE render component over PbB US
+    DONE - show correct disclaimer messages
+    TODO - upload translations  
+    TODO - fix logo
+    DONE - display brands
+    DONE - deduplicate css classes
+    TODO - check if needs clean/refactor
+    */
```
ribeiroguilherme, line 20: Is it missing just the translations for now or is there more stuff?

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodBrands/PaymentMethodBrands.tsx
```diff
@@ -30,7 +41,11 @@ const PaymentMethodBrands = ({ brands, excludedUIBrands = [], isPaymentMethodSel
             {visibleBrands.map(brand => (
                 <PaymentMethodIcon key={brand.name} altDescription={getFullBrandName(brand.name)} type={brand.name} src={brand.icon} />
             ))}
-            {leftBrandsAmount !== 0 && <span className="adyen-checkout__payment-method__brand-number">+{leftBrandsAmount}</span>}
+            {showOtherInsteafOfNumber ? (
+                <span className="adyen-checkout__payment-method__brand-number">+ {i18n.get('paymentMethodBrand.other')}</span>
+            ) : (
+                leftBrandsAmount !== 0 && <span className="adyen-checkout__payment-method__brand-number">+{leftBrandsAmount}</span>
+            )}
```
ribeiroguilherme, line 48: Minor thing, but maybe you can add a simple test case to validate this part

packages/lib/src/components/PayByBankUS/PayByBankUS.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+            i18n: global.i18n,
+            loadingContext: 'test',
+            modules: { resources: global.resources },
+            showPayButton: false
+        });
+
+        render(pbb.render());
+        expect(screen.queryByRole('button')).not.toBeInTheDocument();
+    });
+});
```
ribeiroguilherme, line 39: Can you add a test case here just to make sure that the 'submit' flow is triggered once the payment button is pressed (or .submit is called) ?

We are kinda used to test mainly the UI look, but we aren't actually testing the payment part which is very important.

I believe that you can just validate that the `onSubmit` callback is called, then we are sure that the payment flow works.

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodBrands/PaymentMethodBrands.tsx
m1aw: Yeah, the only problem of this is that we need to validate it in the Dropin, since is not really a PM logic.

packages/lib/src/components/PayByBankUS/PayByBankUS.scss
```diff
@@ -0,0 +1,20 @@
...
+        color: token(color-label-primary);
+    }
+
+    &__description-body {
+        list-style-type: disc;
+        color: token(color-label-secondary);
+        margin: 0;
+        font-size: token(text-body-font-size);
+        line-height: 1.5;
+    }
```
ribeiroguilherme, line 19: The `font-weight` of the body needs to be set to `400` - right now it is bold. 

And there is some margin pixels missing between the header and body - the header should push the body down by 6px.

(I checked with Arjen)

packages/lib/src/components/components-name-map.ts
```diff
@@ -114,6 +114,7 @@ const ComponentsNameMap = {
     [TxVariants.twint]: 'Twint',
     [TxVariants.vipps]: 'Vipps',
     [TxVariants.trustly]: 'Trustly',
+    [TxVariants.paybybank_AIS_DD]: 'PayByBank US',
```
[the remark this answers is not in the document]
m1aw: Oh, I thought it was used for error description. I see it's meant to be the class name, we probably can write this logic in typescript.

packages/lib/src/components/PayByBankUS/PayByBankUS.scss
m1aw: How do you mean it's bold? It isn't bold?

m1aw: Screenshot 2024-09-18 at 14 24 55

packages/lib/src/components/PayByBankUS/PayByBankUS.tsx
m1aw: Translations and logos, however it's getting stored payment added too.

packages/lib/src/components/PayByBankUS/PayByBankUS.scss
ribeiroguilherme: I am getting this:

image

m1aw: Also about the spacings, I will double check as we didn't have figma.

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodBrands/PaymentMethodBrands.tsx
ribeiroguilherme: There is already a `PaymentmethodBrands.test.tsx` - file , perhaps you can just add this there? Just a new test case passing the new props and making sure it renders ok

packages/lib/src/components/components-name-map.ts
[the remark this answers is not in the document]
ribeiroguilherme: It is used for this:
image

packages/lib/src/components/PayByBankUS/PayByBankUS.scss
m1aw: I kinda just want to fix the playground here. What you think?
1. Option - We just fix this in the playground. If the merchant has such rule we should follow what the merchant has.
2. Option - We make a defensive rule, but then we need to do it in the base css styles and not specifically for this component has this can happen again.

ribeiroguilherme: I believe it is better to have a defensive rule on our side. It does not have to be global rule - I guess it is better to make the CSS of this Component more explicit in this case (by setting the font-weight instead of inheriting from somewhere else)

packages/lib/src/components/PayByBankUS/PayByBankUS.tsx
m1aw: FIxed

packages/lib/src/components/PayByBankUS/PayByBankUS.scss
m1aw: Fixed.

packages/lib/src/components/components-name-map.ts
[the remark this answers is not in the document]
m1aw: Fixed.

packages/lib/src/components/PayByBankUS/PayByBankUS.tsx
m1aw: Fixedx.

packages/lib/src/components/PayByBankUS/PayByBankUS.test.tsx
m1aw: Fixed.

ribeiroguilherme (approved):

PR #2889, opened 2024-10-04
PayNow improvements

ribeiroguilherme:
## Summary
- Added `Timeline` component which display an instructions timeline
- PayNow component now renders the `Timeline` at the top in case it is mobile view. The breakpoint for mobile view is width < 1024px 
- Adjusted QR code component CSS according to our design (margins, paddings, font-sizes)
- Added PayNow, PayMe and Timeline storybook stories
- Added translations

## Tested scenarios
- Added unit tests
- Tested PayNow with desktop and mobile view

This PR will be added to the `v6-qr-code-improvements` branch which will contain other changes for other QR code payment methods as well, so we release them all together


packages/lib/src/components/PayNow/PayNow.test.ts
```diff
@@ -21,4 +28,53 @@ describe('PayNow', () => {
...
+            loadingContext: 'checkoutshopper.com/',
+            modules: { resources, analytics: global.analytics, srPanel },
+            i18n: global.i18n,
+            paymentData: 'Ab02b4c0!BQABAgBH1f8hqfFxOvbfK..',
+            qrCodeImage: '',
+            paymentMethodType: 'paynow',
+            qrCodeData: '00020126580009SG...'
+        });
+
+        render(paynow.mount('body'));
```
longyulongyu, line 64: Wouldn't `render(paynow.render())` not work?

packages/lib/src/components/internal/Timeline/Timeline.scss
```diff
@@ -0,0 +1,73 @@
+@import 'styles/variable-generator';
+
+.adyen-checkout-timeline {
+  scrollbar-color: #c0c5cc transparent;
```
longyulongyu, line 4: Don't we have a semantic bento token for it?

packages/lib/src/components/internal/Timeline/Timeline.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.adyen-checkout-timeline__items {
+  all: unset;
+}
+
+.adyen-checkout-timeline-item {
+  scrollbar-color: var(--b-color-background-quaternary) transparent;
```
longyulongyu, line 12: Maybe use `token(color-background-quaternary)`?

packages/lib/src/components/internal/Timeline/Timeline.scss
```diff
@@ -0,0 +1,73 @@
...
+  max-width: token(spacer-070);
+  margin-top: token(spacer-010);
+}
+
+.adyen-checkout-timeline-item__separator {
+  flex-grow: 1;
+  position: relative;
+
+  &::before {
+    background-color: #c9cdd3;
```
longyulongyu, line 47: Same here to use a token?

longyulongyu (approved):

packages/lib/src/components/PayNow/PayNow.test.ts
ribeiroguilherme: It doesn't, there are errors about undefined props of SRPanel. I guess I need to properly mock the object for this test-case. I will try to fix it in the other PR

PR #2892, opened 2024-10-09
Feature/calling onActionHandled for all actions

sponglord:
## Summary
We have the `onActionHandled` callback - designed to be called as soon as we know a specific `action` has been handled.
But it was only being called for `threeds2`, `await` and `qrcode` actions.

This PR also ensures it is also called for `redirect`, `voucher`, `bankTransfer` and `sdk` actions.

Also, for all types, _except_ `threeds2`, the original `action` object, as returned from the `/payments` call, is now passed to the callback (as the property `originalAction`)

## Tested scenarios
Extensive manual testing for each of the action _types_ (`sdk`, `voucher` etc); including testing all 8 types of `Voucher`, to ensure the `onActionHandled` callback is both called _and_ passed the `originalAction` object.

All unit tests pass

**Relates to issue**:  COWEB-1438


ribeiroguilherme:
Wouldn't it work to use the `onActionHandled` on more top level like [here at the core.ts](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/core.ts#L251) or [here in the uielement](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/internal/UIElement/UIElement.tsx#L318) ? Then we would avoid many file changes and prop-drilling.

sponglord:
> Wouldn't it work to use the `onActionHandled` on more top level like [here at the core.ts](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/core.ts#L251) or [here in the uielement](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/internal/UIElement/UIElement.tsx#L318) ? Then we would avoid many file changes and prop-drilling.

I'm working on the principle that this action handling isn't all _synchronous_.
We don't just want to know that we are calling `createFromAction` 'cos we don't know, for sure, that this will lead to anything. Rather we want to know that the action has _actually_ been handled.
So we get as close to the "handling" as possible.
- For 3DS2 this is when the iframe has loaded something.
- For the Redirect this is just before we change window location
- For the qrCode this is when the `` has fired it's `onLoad` handler
- For the Await this is just as we start polling for the first time
- For the 'sdk' this is when the KlarnaWidget fires it's onLoad handler

packages/lib/src/components/internal/Voucher/types.ts
```diff
@@ -47,10 +48,26 @@ export interface VoucherProps {
...
 }
 
 export interface VoucherConfiguration extends UIElementProps {
     reference?: string;
     url?: string;
     paymentMethodType?: string;
+    originalAction?: PaymentAction;
+}
+
+export interface CommonVoucherProps {
```
m1aw, line 65: If you allow my comment here, not necessary but maybe would be nice to do:
```
export interface CommonVoucherProps = Pick<VoucherProps, "paymentMethodType", "introduction", "imageUrl", "onActionHandled", etc..>
```

longyulongyu (approved):

Issue #2824, opened 2024-09-02
Cannot read properties of undefined (reading 'origin')


rchl:
Our production systems receive a relatively large number (8K a month) of client-side errors due to this code:

https://github.com/Adyen/adyen-web/blob/2f59a65b0e272676164a317c668c5293b870c8a9/packages/lib/src/utils/get-process-message-handler.ts#L19

The code crashes because `event.originalEvent` seems to be `undefined`.

99% of crashes are triggered from Chrome Mobile devices on Android 10.

Suggested fix (adding optional chaining operator):

```
const origin = event.origin || event.originalEvent?.origin;
```

ribeiroguilherme:
Hi @rchl ,

Which version of the adyen-web are you using? Is it happening **only** on Android 10 or also in another versions? And do you know the which browser by any chance?

rchl:
Happen to be using `5.43.0` now but upgraded to `5.68.0` earlier today. Will monitor if it makes any difference but looking at the relevant code, it seems like it will still trigger.

Android 10 collects majority of the crashes but it's also occasionally triggered on other versions (not that not all of that can be trusted since UA can be spoofed):

![Screenshot 2024-09-02 at 14 44 58](https://github.com/user-attachments/assets/9d50b055-9ebe-407b-99d3-cec66c39469a)

As far as user agent it's typically something like

```
Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Mobile Safari/537.36
```

ribeiroguilherme:
@rchl I can see that there are two situations where this code is executed: 
- when the library is initially rendered
- when the 3DS flow is triggered
 
Do you know by any chance in which part of the flow this error happens? And do you have visibility if it is a critical error that crashes the UI/checkout or if it is an uncaught error only?

rchl:
Sorry, I don't have visibility into those details and not sure if it's a critical error that prevents people from completion the transaction.

ribeiroguilherme:
Hi @rchl -  Did you notice any difference in the reports since you upgraded your version?

I ran some tests on Samsung Galaxy S20 with Android 10 using Chrome and unfortunately I didn't manage to reproduce the problem.

sponglord:
@rchl 
[...]

PR #2894, opened 2024-10-10
Added optional chaining operator

sponglord:
## Summary
Added optional chaining operator to postMessage handler function

**Fixed issue**:  #2824

### DIFF
--- a/.changeset/rotten-camels-think.md
+++ b/.changeset/rotten-camels-think.md
@@ -2,4 +2,4 @@
 '@adyen/adyen-web': patch
 ---
 
-Added optional chaining operator to postMessage handler function
+Added optional chaining operator to postMessage handler function. Fixes issue where 'get-process-message.handler.js' caused exceptions due to stray postMessages without properly formed events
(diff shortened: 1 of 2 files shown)


m1aw (approved):

.changeset/rotten-camels-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Added optional chaining operator to postMessage handler function
```
ribeiroguilherme, line 5: This ends up in the release notes. Maybe we can be more clear about which problem this solves?

ribeiroguilherme: In my opinion, not really. I think it is better to describe the bug that is fixed rather than the fix itself or point it somewhere else. This also helps the docs team to write better docs and avoid back and forth with questions, and  also improves our release notes by making clear what problem got fixed.

More explanatory sentence for example:
`Fixed issue where 'get-process-message.handler.js' caused exceptions due to stray postMessages`

PR #2898, opened 2024-10-10
Add ACH Checking an Savings

m1aw:
## Summary

This PR adds `bankAccountType` property to ACH. This will allow the shopper to chose between Saving and Checking account.

Together, in this PR we also remove billing address.

Some additional changes were made to some of the UI components. The Radio Group now has a new mode were each radio looks like a button. Some magic CSS numbers and calculation were used to keep compatibility with the existing style. Also for field it was introduced a new 'no-border' style, which allows to still wrap items but without the border logic.

## Tested scenarios

Sessions and Advance flow payments were made, using both options.

**Fixed issue**:  
Ticket: COWEB-1336


packages/lib/src/components/Ach/components/AchInput/AchInput.tsx
```diff
@@ -150,6 +144,21 @@ function AchInput(props: ACHInputProps) {
...
 
+                                <Field classNameModifiers={['bankAccountType', 'no-borders']} name={'bankAccountType'} useLabelElement={false}>
+                                    <RadioGroup
+                                        name={'bankAccountType'}
+                                        value={data.bankAccountType}
+                                        items={[
+                                            { id: 'checking', name: i18n.get('ach.checking') },
+                                            { id: 'savings', name: i18n.get('ach.savings') }
+                                        ]}
+                                        classNameModifiers={['gender']}
```
ribeiroguilherme, line 155: gender?  🤔 
Looks like the `RadioGroup` component does not use `classNameModifiers` .

ribeiroguilherme (approved):

packages/lib/src/components/Ach/components/AchInput/AchInput.tsx
m1aw: Oh yeah, I need to make that change. Also it's missing the clean up on some billing address types that were removed. 
The change has been made for V5, I'm just waiting to submit translations together with it.

PR #2901, opened 2024-10-14
Calling actions.reject() in beforeSubmit should not lead to onPaymentFailed

sponglord:
## Summary
Calling `actions.reject()` in the `beforeSubmit` callback should leave the UI in the current state.
It should not lead to a call to `handleFailedResult` which ultimately leads to a call to the `onPaymentFailed` callback and sets the UI to an error state.
This PR keeps v6 behaviour consistent with v5

## Tested scenarios
Can call `actions.reject()` from `beforeSubmit` and the UI stays in its current state
Added unit tests for `beforeSubmit` resolving and rejecting

**Fixed issue**:  COWEB-1448

### DIFF
--- a/packages/lib/src/components/internal/UIElement/UIElement.tsx
+++ b/packages/lib/src/components/internal/UIElement/UIElement.tsx
@@ -373,7 +373,7 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
     };
 
     /**
-     * Handles a session /payments or /payments/details response.
+     * Handles a /payments or /payments/details response.
      * The component will handle automatically actions, orders, and final results.
      *
      * @param rawResponse -
--- a/packages/lib/src/components/internal/UIElement/UIElement.tsx
+++ b/packages/lib/src/components/internal/UIElement/UIElement.tsx
@@ -23,6 +23,7 @@ import type {
 } from '../../../types/global-types';
 import type { IDropin } from '../../Dropin/types';
 import type { NewableComponent } from '../../../core/core.registry';
+import CancelError from '../../../core/Errors/CancelError';
 
 import './UIElement.scss';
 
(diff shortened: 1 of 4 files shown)


packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -190,12 +199,14 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
-                          reject
+                          reject: () => reject('beforeSubmitRejected')
                       })
                   )
                 : Promise.resolve(this.data);
 
-            return beforeSubmitEvent.then(this.submitUsingSessionsFlow);
+            return beforeSubmitEvent.then(this.submitUsingSessionsFlow).catch(e => {
+                throw new Error(e);
+            });
```
ribeiroguilherme, line 209: We think it can be simplified a bit by handling the error only at the `this.makePaymentsCall` catch block.
One way would be to reject by passing an custom `CancelError` object in `beforeSubmit` promise:

```js
const beforeSubmitEvent: Promise<PaymentData> = this.props.beforeSubmit
            ? new Promise((resolve, reject) =>
                  this.props.beforeSubmit(this.data, this.elementRef, {
                      resolve,
                      reject: () => reject(new CancelError()),
                  })
              )
            : Promise.resolve(this.data);

return beforeSubmitEvent.then(this.submitUsingSessionsFlow);
```

Then when handling the rejection, we could do this way:

```js
this.makePaymentsCall()
            .then(sanitizeResponse)
            .then(verifyPaymentDidNotFail)
            .then(this.handleResponse)
            .catch((error: PaymentResponseData | CancelError) => {
                if (error instanceof CancelError) {
                    this.setElementStatus('ready');
                    return;
                }
                
                this.handleFailedResult(error);
            });
```

This `CancelError` is just an internal error object that we can have on our codebase to cancel certain operations internally. In this case, we use for the `beforeSubmit` , but in the future we can use for other rejections too (Ex: if we have eventually a `beforeAdditionalDetails` , or any other callbacks). 

It can be a simple error object like this:

```js
class CancelError extends Error {}
```

So we can easily extend it in the future if needed. No need to pass any parameters to it for now.

ribeiroguilherme (approved):
Can you add the changeset pointing out the bugfix?

Also, would be nice to have an unit testsin place to cover this regression. It could be done in another PR

longyulongyu (approved):

PR #2910, opened 2024-10-21
PayMe and PayNow improvements

ribeiroguilherme:
## Summary
Parent branch containing the changes of #2889 and #2909


packages/lib/src/components/internal/Timeline/Timeline.scss
```diff
@@ -0,0 +1,73 @@
...
+.adyen-checkout-timeline {
+  scrollbar-color: token(color-background-quaternary) transparent;
+}
+
+.adyen-checkout-timeline__items {
+  all: unset;
+}
+
+.adyen-checkout-timeline-item {
+  scrollbar-color: var(color-background-quaternary) transparent;
```
longyulongyu, line 12: I think we need to use `token(color-background-quaternary)` here.

longyulongyu (approved):

PR #2921, opened 2024-10-25
V6 fix initial state upi intent

longyulongyu:
## Summary
cherry pick from fe369df73c5a0d11801ba68f511b96003de8b65e

### DIFF
--- a/packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
+++ b/packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
@@ -34,7 +34,7 @@ export default function UPIComponent({ defaultMode, onChange, onUpdateMode, payB
     const { i18n } = useCoreContext();
     const getImage = useImage();
     const [status, setStatus] = useState<UIElementStatus>('ready');
-    const [isValid, setIsValid] = useState<boolean>(defaultMode === UpiMode.QrCode);
+    const [isValid, setIsValid] = useState<boolean>(defaultMode === 'qrCode');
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -34,7 +34,7 @@ export default function UPIComponent({ defaultMode, onChange, onUpdateMode, payB
     const { i18n } = useCoreContext();
     const getImage = useImage();
     const [status, setStatus] = useState<UIElementStatus>('ready');
-    const [isValid, setIsValid] = useState<boolean>(true);
+    const [isValid, setIsValid] = useState<boolean>(defaultMode === 'qrCode');
```
sponglord, line 37: How come the v5 version has a constant (`defaultMode === UpiMode.QrCode`) but here we're using a string?

sponglord (approved):

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
longyulongyu: In v6, we use string literals for the `UpiMode`, but in v5 we define it as enum, that's why we need to code differently. The reason to change the type definition is because string literals are easier for JS users.

sponglord: Just an observation: we use the string `qrCode` in quite a few places - it just feels like something that could be defined as a constant somewhere.
This is also true in a lot of other places in our codebase

PR #2939, opened 2024-11-07
Feat: enable logos for SK and CZ online banking

m1aw:
## Summary

Set `showImage: true` for both CZ and SK online banking. Images should be already available.

Also remove `SKK` support from the playground in favour of `EUR`. 

## Tested scenarios

CZ component.
Screenshot 2024-11-07 at 18 23 15

SK component tested, but missing the full response in Test env, should be correct in Live according to LPM.

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/OnlineBankingCZ/index.ts
+++ b/packages/lib/src/components/OnlineBankingCZ/index.ts
@@ -15,7 +15,7 @@ class OnlineBankingCZElement extends IssuerListContainer {
     formatProps(props) {
         return {
             ...super.formatProps(props),
-            showImage: false,
+            showImage: true,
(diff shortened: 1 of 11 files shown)


packages/lib/src/components/OnlineBankingCZ/index.ts
```diff
@@ -15,7 +15,7 @@ class OnlineBankingCZElement extends IssuerListContainer {
     formatProps(props) {
         return {
             ...super.formatProps(props),
-            showImage: false,
+            showImage: true,
```
ribeiroguilherme, line 18: I guess it can be omitted since its default is true. Same for OnlineBankingSK

ribeiroguilherme (approved):

PR #2951, opened 2024-11-13
Refactored plcc related e2e tests to work in new framework

sponglord:
## Summary
Refactored `plcc.luhn.spec.ts` to work in the new framework

## Tested scenarios
e2e test passes


packages/e2e-playwright/tests/ui/card/binLookup/plcc/plcc.noluhn.nodate.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+        // Allow time for icon to load
+        await page.waitForTimeout(500);
+
+        // UI reset
+        brandingIconSrc = await card.brandingIcon.getAttribute('src');
+        expect(brandingIconSrc).toContain('nocard.svg');
+
+        // PM is not valid
+        cardValid = await page.evaluate('window.component.isValid');
+        expect(cardValid).toEqual(false);
```
ribeiroguilherme, line 44: Is there a way to test this without checking the 'internals' of the component state? Maybe by attempting to make the payment and then verifying that there is an error in the UI?

sponglord: I don't know, probably. 
But since I'm still testing internal state on line 30, to check that the component _is_ valid (and I don't want to press the Pay button then 'cos I want todo more in the UI) - I'll leave it, as it is for now,

PR #2955, opened 2024-11-14
Add fix to reshow card icon if previously it failed to load

sponglord:
## Summary
If an icon in the Credit card PAN field failed to load - an error handler set inline an style `display:none`.
And this inline style was never removed in the event of a subsequent, successful, icon load.
This PR fixes that

## Tested scenarios
New e2e test that was previously flaky or failing due to this - now passes consistently
Manually tested.


ribeiroguilherme (approved):
Do we need changeset for it?

PR #2956, opened 2024-11-14
Chore/bin lookup pan length tests

sponglord:
## Summary
Refactored `panLength.focus.regular.spec.ts` to work in the new framework

## Tested scenarios
e2e tests pass

### DIFF
--- a/packages/e2e-playwright/models/card.ts
+++ b/packages/e2e-playwright/models/card.ts
@@ -58,6 +58,9 @@ class Card extends Base {
     readonly addressField: Locator;
     readonly addressLabelWithFocus: Locator;
 
+    readonly kcpTaxNumberField: Locator;
+    readonly kcpTaxNumberLabelWithFocus: Locator;
+
     constructor(
         public readonly page: Page,
         rootElementSelector = '.adyen-checkout__card-input'
@@ -140,6 +143,12 @@ class Card extends Base {
          */
         this.addressField = this.rootElement.locator('.adyen-checkout__field--street'); // Holder
         this.addressLabelWithFocus = this.addressField.locator('.adyen-checkout__label--focused');
+
+        /**
+         * KCP related elements
+         */
+        this.kcpTaxNumberField = this.rootElement.locator('.adyen-checkout__field--kcp-taxNumber'); // Holder
+        this.kcpTaxNumberLabelWithFocus = this.kcpTaxNumberField.locator('.adyen-checkout__label--focused');
     }
 
     get availableBrands() {
--- a/packages/e2e-playwright/models/card.ts
+++ b/packages/e2e-playwright/models/card.ts
@@ -21,7 +21,6 @@ class Card extends Base {
 
     readonly cardNumberField: Locator;
     readonly cardNumberLabelElement: Locator;
-    readonly cardNumberLabelWithFocus: Locator;
     readonly cardNumberErrorElement: Locator;
     readonly cardNumberInput: Locator;
     readonly brandingIcon: Locator;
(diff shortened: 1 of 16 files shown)


packages/e2e-playwright/tests/ui/card/binLookup/panLength/panLength.focus.kcp.spec.ts
```diff
@@ -1,58 +1,54 @@
...
-        //
-        // // Expect focus to be place on tax number field
-        // await t.expect(cardPage.kcpTaxNumberLabelWithFocus.exists).ok();
+        await card.isComponentVisible();
+
+        await card.typeCardNumber(REGULAR_TEST_CARD);
+
+        // Expect UI change - tax number field has focus
+        await expect(card.cardNumberLabelWithFocus).not.toBeVisible();
+        await expect(card.kcpTaxNumberLabelWithFocus).toBeVisible();
```
ribeiroguilherme, line 29: It would be better to leverage the testing API rather than checking for locators with classnames.  There are different API's to check specific input/UI states (focus, readonly, disabled, etc..) If I am not mistaken, that also ensures consistent a11y

In this case something like `await expect(card.kcpTaxNumberInput).toBeFocused();` ideally would work

In the Card model, we ideally should keep the locator of the Input field, and to test its different tests we leverage the testing API. We should not need to keep locators for `withFocus` for example

packages/e2e-playwright/models/card.ts
```diff
@@ -96,13 +118,37 @@ class Card extends Base {
...
+        /**
+         * Social Security Number related elements
+         */
+        this.ssnField = this.rootElement.locator('.adyen-checkout__field--socialSecurityNumber'); // Holder
+        this.ssnLabelWithFocus = this.ssnField.locator('.adyen-checkout__label--focused');
+
+        /**
+         * AVS related elements
+         */
+        this.addressField = this.rootElement.locator('.adyen-checkout__field--street'); // Holder
```
ribeiroguilherme, line 144: We already have the model `card-avs.ts`, which contains the address field on it.

My question here is if we are going to add everything to this model (avs, kcp, address lookup, etc..) or if we are going to split in different models which each will contain their own custom logic/locators. 

I don't have a strong preference tbh, but we need to decide. We already started splitting things (as we have with `card-avs`) but we can always go back if we think that it is not needed.

ribeiroguilherme (approved):

packages/e2e-playwright/tests/ui/card/binLookup/panLength/panLength.focus.avs.spec.ts
```diff
@@ -1,44 +1,24 @@
...
+import { URL_MAP } from '../../../../../fixtures/URL_MAP';
+import { binLookupMock } from '../../../../../mocks/binLookup/binLookup.mock';
+import { optionalDateAndCvcWithPanLengthMock } from '../../../../../mocks/binLookup/binLookup.data';
+import { REGULAR_TEST_CARD } from '../../../../utils/constants';
 
-/**
- * NOTE - we are mocking the response until such time as we have a genuine card that returns the properties we want to test
- */
+test.describe('Test Card, binLookup w. panLength property & address fields', () => {
+    test('#1 Fill out PAN & see that focus moves to an address field since expiryDate & cvc are optional', async ({ card, page }) => {
```
longyulongyu, line 9: I think we can reuse the `cardWithAvs` from the `card.fixture.ts`? The `CardWithAvs` class extends from `Card`, so it contains all selectors from the card. If there are certain selectors missing, we can add them in the `CardWithAvs` class.

packages/e2e-playwright/models/card.ts
longyulongyu: I believe the `CardWithAvs` class has this selector - `await cardAvsPage.billingAddress.streetInput`?
Regarding the `focus`, wouldn't this code work? 
```ts
await expect(cardAvsPage.billingAddress.streetInput).toBeFocused();
```

sponglord: @ribeiroguilherme I'll make use of the other models / split into other models

PR #2973, opened 2024-11-20
Send error events for redirect errors

longyulongyu:
## Summary
- Send the redirect error event to the analytics when `doRedirect` throws an error
- Refine some types
- Fix some lint warnings

## Tested scenarios
- Added unit test
- Tested when manually simulate an error from `doRedirect`

**Fixed issue**:  COWEB-1463

### DIFF
--- a/packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
+++ b/packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
@@ -10,12 +10,14 @@ interface RedirectShopperProps {
     redirectFromTopWhenInIframe?: boolean;
     paymentMethodType?: string;
     onActionHandled?: (rtnObj: ActionHandledReturnObject) => void;
+    onRedirectError?: () => void;
 }
 
 class RedirectShopper extends Component<RedirectShopperProps> {
     private postForm;
     public static defaultProps = {
         beforeRedirect: resolve => resolve(),
+        onRedirectError: () => {},
         method: 'GET'
     };
 
@@ -49,7 +51,9 @@ class RedirectShopper extends Component<RedirectShopperProps> {
                 })
         );
 
-        dispatchEvent.then(doRedirect).catch(() => {});
+        dispatchEvent.then(doRedirect).catch(() => {
+            this.props.onRedirectError();
+        });
     }
 
     render({ url, method, data = {} }) {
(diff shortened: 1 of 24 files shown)


packages/lib/src/core/Analytics/EventsQueue.ts
```diff
@@ -24,7 +24,7 @@ const EventsQueue = ({ analyticsContext, clientKey, analyticsPath }: EventQueueP
         logs: []
     };
 
-    const runQueue = (checkoutAttemptId: string): Promise<any> => {
+    const runQueue = async (checkoutAttemptId: string): Promise<any> => {
```
ribeiroguilherme, line 27: `async` is not needed since there is no await

packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
```diff
@@ -49,14 +51,17 @@ class RedirectShopper extends Component<RedirectShopperProps> {
                 })
         );
 
-        dispatchEvent.then(doRedirect).catch(() => {});
+        dispatchEvent.then(doRedirect).catch(() => {
+            this.props.onRedirectError();
```
ribeiroguilherme, line 55: Can you recall me in which scenarios does the Redirect may fail?

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -161,17 +161,25 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+        }
+        if (this.constructor['analyticsType']) {
+            return this.constructor['analyticsType'];
+        }
+        if (this.constructor['type'] === 'scheme' || this.constructor['type'] === 'bcmc') {
+            return this.constructor['type'];
+        }
+        return this.props.type;
+    }
+
```
ribeiroguilherme, line 182: Feels a bit off this function to be here.. Looking at it, gives impression that would return a Component, but then it returns a string (tx variant or the `analyticsType`). Besides that, it is fully related to analytics as far as I understand.

But, I am not sure what to suggest here to be honest 😅  Couldn't that logic be handled internally by the `sendAnalytics` function?

ribeiroguilherme (approved):

packages/lib/src/components/Redirect/components/RedirectShopper/RedirectShopper.tsx
longyulongyu: To be honest, I cannot think of that many situations except for `beforeRedirect` callback gets rejected. `doRedirect` looks reliable to me so i cannot think of an error situation at this moment.

PR #3005, opened 2024-12-04
Add e2e tests for kcpAuthentication, stored-card-amex, stored-card-mc and stored-card-visa

longyulongyu:
## Summary
- Add e2e tests for kcpAuthentication, stored-card-amex, stored-card-mc and stored-card-visa

### DIFF
--- a/packages/e2e-playwright/models/base.ts
+++ b/packages/e2e-playwright/models/base.ts
@@ -14,6 +14,10 @@ export abstract class Base {
         await this.isComponentVisible();
     }
 
+    async isPayable() {
+        await this.page.waitForFunction(() => globalThis.component.isValid === true);
+    }
+
     async pay(options: { name?: RegExp | string } = { name: /pay/i }): Promise<void> {
         if (this.payButton) {
             await this.payButton.click();
--- a/packages/e2e-playwright/tests/e2e/card/kcpAuthentication.spec.ts
+++ b/packages/e2e-playwright/tests/e2e/card/kcpAuthentication.spec.ts
@@ -1,27 +1,46 @@
-import { test } from '../../../fixtures/card.fixture';
+import { expect, test } from '../../../fixtures/card.fixture';
+import { URL_MAP } from '../../../fixtures/URL_MAP';
+import {
+    KOREAN_TEST_CARD,
+    PAYMENT_RESULT,
+    REGULAR_TEST_CARD,
+    TEST_CVC_VALUE,
+    TEST_DATE_VALUE,
+    TEST_PWD_VALUE,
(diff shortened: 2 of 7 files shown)


packages/e2e-playwright/tests/e2e/card/kcpAuthentication.spec.ts
```diff
@@ -1,27 +1,46 @@
...
-        // click pay btn
-        // expect to see success msg
+    test('should submit the regular non-korean card payment', async ({ cardWithKCP, page }) => {
+        await cardWithKCP.goto(URL_MAP.cardWithKcp);
+        await cardWithKCP.fillCardNumber(REGULAR_TEST_CARD);
+        await cardWithKCP.fillCvc(TEST_CVC_VALUE);
+        await cardWithKCP.fillExpiryDate(TEST_DATE_VALUE);
+        await cardWithKCP.isPayable();
+        await cardWithKCP.pay();
+        await expect(cardWithKCP.paymentResult).toContainText(PAYMENT_RESULT.authorised);
```
sponglord, line 44: Does this  whole file not overlap with `e2e/card/kcp/card.kcp.spec.ts`?

longyulongyu: I am not 100% sure because there is a `paymentSuccessfulMock` in the `e2e/card/kcp/card.kcp.spec.ts` file i suppose it's not really making a payment call.
I think if we remove the `paymentSuccessfulMock` those 2 files are overlapping.

sponglord (approved):

packages/e2e-playwright/tests/e2e/card/kcpAuthentication.spec.ts
sponglord: Yeah, I think, at the time, we weren't sure if we had an actual, working korean local card number - so I guess that's why @ribeiroguilherme used a mock?

packages/e2e-playwright/models/base.ts
```diff
@@ -14,6 +14,10 @@ export abstract class Base {
         await this.isComponentVisible();
     }
 
+    async isPayable() {
```
sponglord, line 17: We have several places where we either say:
`await page.waitForFunction(() => globalThis.component.isValid === true);`
or
`await page.waitForFunction(() => globalThis.component.isValid === false);`

This function could take a parameter & then be used in all these places

longyulongyu: Yes it's a nice find. I'll refactor this in another PR 😄.

PR #3012, opened 2024-12-09
Add missing Trustly CZ translations

strublos:
## Summary
Add missing CZ translation for Trustly component

## Tested scenarios
Switch language to CZ - show correct labels intead of trustly.descriptor key.

**Fixed issue**:


ribeiroguilherme:
Hi @strublos -  thanks for creating the PR. We realized that the issue does not impact only CZ but also other translations, so we opened another PR to fix this. Therefore I will close this one. Thanks for flagging it.

PR #3015, opened 2024-12-10
Stop calling sanitizeResponse twice during the payment response handl…

sponglord:
…ing process

## Summary
As discussed previously - it is not necessary for the `UIElement`s `handleResponse` function to call `sanitizeResponse` on the passed object, since this function has already been run earlier in the handling-the-payments-response, or, handling-the-details-response, chains.

This also applies to it's use in the `ApplePay` and `GooglePay` components 

## Tested scenarios
unit tests pass


ribeiroguilherme (approved):
Might be worth to mention this in the changeset as sort of internal improvement, so at least we can easily keep track of the changes introduced in each release.

Issue #3009, opened 2024-12-09
Click2Pay: Missing keyboard activation for buttons


FilipSvozil:
**Describe the bug**
The _"Not you?"_ and _"Resend code"_ buttons cannot be activated using keyboard.

**To Reproduce**
1. Open Click2Pay component
2. Enter valid email
3. Tab to the _"Not you?"_ or _"Resend code"_ button (→ focus indicator shows correctly)
4. Press _Space_ or _Enter_ key to activate the button → doesn't work, and the screen scrolls down instead

**Expected behavior**
The button's onClick action should execute, and the screen should not scroll down.

**Screenshots**
<img width="446" alt="Screenshot 2024-12-09 at 9 57 33" src="https://github.com/user-attachments/assets/e0d14baf-97f4-4dbd-b7cb-21e14660fb63">

**Desktop (please complete the following information):**
 - MacOS 15
 - Chrome 131 and Safari 18

**Additional context**
We are using npm package: @adyen/adyen-web@6.5.0
* The _"Not you?"_ button is implemented as `<span>`, and doesn't handle keyboard activation
  * Should probably be `<button>` - then `role` and `tabIndex` would no longer be needed
  * https://github.com/Adyen/adyen-web/blob/v6.5.0/packages/lib/src/components/internal/ClickToPay/components/CtPSection/CtPLogoutLink.tsx#L32
* The _"Resend code"_ button is implemented as `<div>`, and doesn't handle keyboard activation
  * Should probably be `<button>` - then `role` and `tabIndex` would no longer be needed
    * The current `role="link"` also seems incorrect - it doesn't lead anywhere, it performs an action
  * https://github.com/Adyen/adyen-web/blob/v6.5.0/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPResendOtpLink.tsx#L91

-----

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#best_practices
> Where possible, it is recommended to use native HTML buttons rather than the button role

m1aw:
Hello @FilipSvozil thank you for raising the issue.

This has indeed escaped our review process, we will fix in the coming releases.

camil-adyen:
Hi @FilipSvozil, we released a fix for this issue in [v6.7.0](https://github.com/Adyen/adyen-web/releases/tag/v6.7.0). I'll close the issue for now, but feel free to reopen if you have any other comments.

PR #3029, opened 2024-12-17
Fixes Click to Pay link looking buttons into proper buttons

m1aw:
## Summary
We had some elements that should have been buttons but weren't. This prevented behind activated with the keyboard. 

This PR makes them proper buttons, but also adds the `variant` link to the button types. 

To keep things consistent with the rest of the SDK the style of this buttons has also been changed from primary color (black) with underline to link color (blue) with no underline.

We also changed some markup so this might be a breaking change (UI wise), since both CSS classes and the overall markup.

## Tested scenarios

All tests pass and it was manually tested with my CTP both buttons work and also the states of the OTP request message.

Screenshot 2024-12-17 at 17 10 02

**Fixed issue**:  
#3009

### DIFF
--- a/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
+++ b/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
@@ -100,31 +100,31 @@ const CtPOneTimePasswordInput = (props: CtPOneTimePasswordInputProps): h.JSX.Ele
     }, [data, valid, errors]);
 
     return (
-        <Field
-            name="oneTimePassword"
-            label={i18n.get('ctp.otp.fieldLabel')}
-            labelEndAdornment={
-                !props.hideResendOtpButton && (
-                    <CtPResendOtpLink disabled={props.isValidatingOtp} onError={handleOnResendOtpError} onResendCode={handleOnResendOtp} />
-                )
-            }
-            errorMessage={isOtpFielDirty ? resendOtpError || props.errorMessage || !!errors.otp : null}
-            classNameModifiers={['otp']}
-        >
-            <InputText
-                name={'otp'}
-                autocorrect={'off'}
-                spellcheck={false}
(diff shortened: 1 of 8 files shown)


packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
```diff
@@ -100,31 +100,31 @@ const CtPOneTimePasswordInput = (props: CtPOneTimePasswordInputProps): h.JSX.Ele
     }, [data, valid, errors]);
 
     return (
-        <Field
-            name="oneTimePassword"
-            label={i18n.get('ctp.otp.fieldLabel')}
-            labelEndAdornment={
```
m1aw, line 106: @ribeiroguilherme this was the reason the test was failing.

Basically we were creating the button inside the label, which made that the label invalid since it can only contain text. So all the selectors where not picking it up.

So I have refactored it a bit and moved the button to the outside of the label and used so CSS trickery to place the button where it was previously. 

My main concern with the change is the change in classnames, but let me know what you think.

ribeiroguilherme: The change is fine - it was not following the correct approach and now it is, so there is no discussion about it. I am just concern about merchant's customization in this case. 

Some classnames changed, and merchants might have target the 'links' CSS and this impact their changes, but seems like there is not much to do in this case as we need to apply this fix.

ribeiroguilherme (approved):

packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
m1aw: Yeah, that's what I was concerned too. I tried to minimize the impact of those changes, but I wanted to know your opinion on it.

In that case let's proceed with this.

PR #3031, opened 2024-12-18
Toggle component

ribeiroguilherme:
## Summary
- Creating Toggle component based on the Bento component
- Imported`toggle` css variables from Bento to properly provide the correct colors/values to the Toggle SCSS file

(This component will be used in the fastlane branch)

## Tested scenarios
- unit tests; storybook manual test

### DIFF
--- a/packages/lib/src/components/internal/Toggle/Toggle.scss
+++ b/packages/lib/src/components/internal/Toggle/Toggle.scss
@@ -8,7 +8,6 @@
   color: inherit;
   cursor: pointer;
   display: flex;
-  user-select: none;
   width: auto;
 
   @include mixins.box-sizing-setter(true);
--- a/packages/lib/src/components/internal/Toggle/Toggle.scss
+++ b/packages/lib/src/components/internal/Toggle/Toggle.scss
@@ -1,0 +1,64 @@
+.adyen-checkout-toggle {
+  color: inherit;
+  cursor: pointer;
+  display: flex;
+  user-select: none;
+  width: auto;
+}
+
+.adyen-checkout-toggle--label-first {
+  align-items: flex-start;
+  flex-direction: row-reverse;
+  justify-content: flex-end;
+}
+
+.adyen-checkout-toggle__input {
+  cursor: inherit;
+  opacity: 0;
+  position: absolute;
+}
+
+.adyen-checkout-toggle__input:checked+.adyen-checkout-toggle__track {
+  background-color: #00112c;
+  border: 0;
+  padding: 2px;
+}
+
+.adyen-checkout-toggle__track {
+  align-items: center;
+  background-color: #fff;
+  border: 1px solid #8d95a3;
+  border-radius: 12px;
+  display: flex;
+  height: 20px;
+  min-width: 36px;
+  padding: 2px 4px;
+  position: relative;
+}
+
+.adyen-checkout-toggle__handle {
+  align-content: center;
+  background-color: #00112c;
+  border-radius: 12px;
+  color: #00112c;
+  display: inline-flex;
+  height: 12px;
+  justify-content: center;
+  transition: transform .15s cubic-bezier(.2, 0, .4, .9);
+  width: 12px;
+}
+
+.adyen-checkout-toggle__input:checked+* .adyen-checkout-toggle__handle {
+  background-color: #fff;
+  height: 16px;
+  transform: translate(100%);
+  width: 16px;
+}
+
+.adyen-checkout-toggle__label-text {
+  font-size: 14px;
+  font-weight: 400;
+  letter-spacing: 0;
+  line-height: 20px;
+
+}
--- a/packages/lib/src/components/internal/Toggle/Toggle.scss
+++ b/packages/lib/src/components/internal/Toggle/Toggle.scss
@@ -1,64 +1,159 @@
+@use 'styles/mixins';
+@import 'styles/variable-generator';
+
 .adyen-checkout-toggle {
+  $component-root: &;
+  $label-padding: token(toggle-label-padding);
+
   color: inherit;
   cursor: pointer;
   display: flex;
   user-select: none;
   width: auto;
-}
 
-.adyen-checkout-toggle--label-first {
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/internal/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,78 @@
...
+        [onChange]
+    );
+
+    return (
+        <label className={`adyen-checkout-toggle ${conditionalClasses}`}>
+            <input
+                disabled={disabled}
+                checked={checked}
+                onChange={onInputChange}
+                aria-label={computedAriaLabel}
```
m1aw, line 41: We don't need `aria-label` here. The value from the label already comes from the wrapping label.

packages/lib/src/components/internal/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,78 @@
+import { h } from 'preact';
+import { useCallback, useMemo } from 'preact/hooks';
+import cx from 'classnames';
+import uuid from '../../../utils/uuid';
+import './Toggle.scss';
+
+interface ToggleProps {
+    label?: string;
```
m1aw, line 8: Label should be mandatory?

packages/lib/src/components/internal/Toggle/Toggle.scss
```diff
@@ -0,0 +1,159 @@
...
+@import 'styles/variable-generator';
+
+.adyen-checkout-toggle {
+  $component-root: &;
+  $label-padding: token(toggle-label-padding);
+
+  color: inherit;
+  cursor: pointer;
+  display: flex;
+  user-select: none;
```
m1aw, line 11: Was this just part of the copy paste?

I wonder if it's worth doing this.

packages/lib/src/components/internal/Toggle/Toggle.scss
```diff
@@ -0,0 +1,159 @@
...
+    border: token(toggle-track-border);
+    border-radius: token(toggle-track-border-radius);
+    display: flex;
+    height: token(toggle-track-height);
+    min-width: token(toggle-track-width);
+    padding: token(toggle-track-padding);
+    position: relative;
+
+    #{$component-root}__input:focus-visible + & {
+      @include mixins.b-focus-ring;
```
m1aw, line 49: We probably want to rename this focus ring, to something else. 

There's some Focus state changes coming at some point so it will be easier to refactor.

m1aw (approved):

packages/lib/src/components/internal/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,78 @@
...
+                aria-describedby={descriptionId}
+                role="switch"
+                type="checkbox"
+                className="adyen-checkout-toggle__input"
+            />
+
+            <span aria-hidden={true} className="adyen-checkout-toggle__track">
+                <span className="adyen-checkout-toggle__handle">
+                    {checked && (
+                        <svg role="img" xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="none">
```
longyulongyu, line 52: I think this part could potentially be refactored into a checkmark component, optionally though.

longyulongyu (approved):

packages/lib/src/components/internal/Toggle/Toggle.scss
ribeiroguilherme: Doing what exactly? 🤔  

The CSS was copy-paste from Bento's Toggle component (with minor adjustments), since we use the same DOM structure.

packages/lib/src/components/internal/Toggle/Toggle.scss
ribeiroguilherme: I understand your concern, but I am not aware of the focus state changes that will come up nor have no idea what to rename that to 😅  We are following the bento mixins here.

What is your proposal here?

packages/lib/src/components/internal/Toggle/Toggle.tsx
ribeiroguilherme: Yeah I was thinking the same .. To be honest I wasn't sure what to do, since:
- we don't have SVG as components in our codebase (although I don't see a problem to have few bento SVG's as components, for example this checkmark icon)
- we have this checkmark on the CDN  - although I didn't want to add complexity to this component to use the `getImage` hook to download the checkmark. Feels like too much to a simple UI component. Doing this would add more complexity, external dependency, and complicate more unit tests

packages/lib/src/components/internal/Toggle/Toggle.scss
m1aw: Use-select removes the possibility of copy and pasting code. It can be useful but I think it should be only used in extreme cases.

We don't do this for labels, and we don't do this for buttons so I don't think we should use it here too.

packages/lib/src/components/internal/Toggle/Toggle.scss
m1aw: I was just proposal to rename it something `checkout-` prefixed, instead of `b-` from the bento mixins.

Anyway, I see we already use it as the default, my bad.

I will change this, at some point. The changes are the ones we noticed about the button `focus` state not having a good enough outline around it. There's a few components where focus state is like the buttons and it probably needs some love.

packages/lib/src/components/internal/Toggle/Toggle.tsx
ribeiroguilherme: To be honest I am not sure - The bento definition for Toggle says that 'toggle only' is one of the possible ways of using it, although I don't see any use-case (at least for us) right now.

I added that anyway since I wanted to make the component comply with the Bento one. 
But if you think that we don't need this, then I can remove this feature

packages/lib/src/components/internal/Toggle/Toggle.scss
ribeiroguilherme: Makes sense

PR #3033, opened 2024-12-19
Migrating bcmc e2e tests

sponglord:
## Summary
Migrating bcmc e2e tests, ensuring all test cases are covered from the original testcafe tests


packages/e2e-playwright/tests/e2e/card/bcmc/dualBranding.spec.ts
```diff
@@ -13,7 +13,9 @@ import { URL_MAP } from '../../../../fixtures/URL_MAP';
 test.describe('Bcmc payments with dual branding', () => {
     test.describe('Bancontact (BCMC) / Maestro brands', () => {
         test.describe('Selecting the Bancontact brand', () => {
-            test('should submit the bcmc payment', async ({ bcmc }) => {
+            test('#1a should submit the bcmc payment', async ({ bcmc, page }) => {
+                const paymentsRequestPromise = page.waitForRequest(request => request.url().includes('/payments') && request.method() === 'POST');
```
longyulongyu, line 17: I think this can be moved to unit test though, inspecting the correct brand sent in the payload data.

packages/e2e-playwright/tests/ui/dropin/bcmc/dropinWithBcmc.fixture.ts
```diff
@@ -0,0 +1,28 @@
+import { test as base, mergeTests, expect } from '@playwright/test';
+import { DropinWithSession } from '../../../../models/dropinWithSession';
+import { test as card } from '../../../../fixtures/card.fixture';
+
+class DropinWithBcmc extends DropinWithSession {
```
longyulongyu, line 5: Isn't it not possible to use `mergeTest`, so that we don't need to create the class `DropinWithBcmc`? Like we do in `stored-card-ach.spec.ts` or `stored-card-bancontact.spec.ts`?

packages/e2e-playwright/tests/e2e/card/bcmc/dualBranding.spec.ts
sponglord: Why not just do it here - since we're submitting the payload anyway?

packages/e2e-playwright/tests/ui/dropin/bcmc/dropinWithBcmc.fixture.ts
sponglord: Then why do we have something like the class `DropinWithCard`?

packages/e2e-playwright/tests/e2e/card/bcmc/dualBranding.spec.ts
longyulongyu: We can do it here, but i think it's better to keep the e2e test suite simple and easy (not testing too much implementation details). Plus it doesn't calculate the coverage if we put too much things into the e2e tests.

packages/e2e-playwright/tests/ui/dropin/bcmc/dropinWithBcmc.fixture.ts
longyulongyu: i see, i think this this `DropinWithCard` is checking the brands before selecting the card(see the screenshot), the logic still belongs to dropin, but not generic enough to put it to `dropinWithSession` or `dropin`. Hence i feel the need to create an extra class.
Screenshot 2024-12-20 at 16 19 53

sponglord: I'm also checking the brands before selecting the BCMC

longyulongyu: So i guess the same reason for the `dropinWithBcmc` then..

longyulongyu (approved):

packages/e2e-playwright/tests/e2e/card/bcmc/dualBranding.spec.ts
sponglord: I'll look into it in another PR

PR #3043, opened 2024-12-31
Fastlane Card Component signup

ribeiroguilherme:
## Summary
This PR introduces the Fastlane UI inside the Card component. 

How it works:
- The fastlane UI is rendered **only** if the Card component gets a valid `fastlaneConfiguration` object
- The UI is displayed only when the card number entered is Visa or MC 
- There is no validation for the mobile phone number . That is a deliberate decision

Some highlights:
- Created `Link` component which we can use for hyperlinks
- Introduced `staticValue` to the`Field` component, so we can display static values (similar to how Bento does)
- Card component state.data might contain the `fastlaneData` property, which is an encoded blob containing the data for the payment

This flow can be tested through Storybook on the Fastlane stories.

## Tests
- Added unit tests covering callbacks and interactions

## Next PR's:
- translations


sponglord:
I know you say we don't validate the mobile number - but it feels wrong that you can submit this PM whilst the field is visible - but clearly not filled

ribeiroguilherme:
> I know you say we don't validate the mobile number - but it feels wrong that you can submit this PM whilst the field is visible, but clearly not filled

Yea I agree. It is weird.. the main point for now is to not ruin the conversion if the shopper didn't enter the phone or entered the wrong phone..

sponglord (approved):

packages/lib/src/components/Card/components/Fastlane/InfoModal.scss
```diff
@@ -0,0 +1,56 @@
...
+
+        &-image {
+            grid-area: image;
+            place-self: center;
+        }
+
+        &-header {
+            grid-area: header;
+            margin: 0;
+            line-height: token('text-title-m-line-height');
```
longyulongyu, line 27: I think no need to add `'` when using the token for the consistency reason, I believe it's still working.

packages/lib/src/components/internal/Link/Link.scss
```diff
@@ -1,5 +1,7 @@
-@use 'mixins';
+@use 'styles/mixins';
+@import 'styles/variable-generator';
```
longyulongyu, line 2: I think no need to import `styles/variable-generator`? No token is used in this file.

PR #3053, opened 2025-01-03
TS : argument `options` in ICore:createFromAction is optional

dimitriBouteille:
## Summary

This PR fix typescript for `ICore::createFromAction()` : 

- The argument `options` is optional.

Here is an example where the argument `options` is (probably) considered optional

https://github.com/Adyen/adyen-web/blob/46d6f27b81628d3785e10c58f7491eb77b9dc0df/packages/lib/src/core/core.ts#L226-L228

https://github.com/Adyen/adyen-web/blob/46d6f27b81628d3785e10c58f7491eb77b9dc0df/packages/lib/src/core/core.test.ts#L103-L109


sponglord (approved):

sponglord:
Thanks @dimitriBouteille - we have merged this PR

PR #3067, opened 2025-01-07
fix: use `rem` for the font size unit

longyulongyu:
## Summary
- Added the `packages/lib/src/styles/overrides.scss` to convert the font size from `px` to `rem`. Calculation is based on a root font-size of 16 pixel
- Fixed some places that font tokens are not used
- Change the Storybook global css so that it reacts to the font scaling - Font Size from the browser setting: `chrome://settings/appearance`
- Fix the validation icon incorrect position issue when using the customized font size

## Tested scenarios
Change the font size from `chrome://settings/appearance`, the sdk font size should adapt and the UI should still look decent.

**Fixed issue**:  COWEB-1472


packages/lib/src/components/internal/FormFields/FormFields.scss
```diff
@@ -153,6 +153,10 @@ span.adyen-checkout__input {
     [dir='rtl'] & {
         margin-left: token(spacer-060);
     }
+
+    > .adyen-checkout__icon {
+        display: block;
+    }
```
sponglord, line 159: just curious... was this selector missing then?

sponglord (approved):

packages/lib/src/styles/overrides.scss
```diff
@@ -0,0 +1,15 @@
...
+    // 14px
+    'text-body-font-size': 0.875rem,
+    // 16px
+    'text-subtitle-font-size': 1rem,
+    // 16px
+    'text-title-font-size': 1rem,
+    // 24px
+    'text-title-l-font-size': 1.5rem,
+    // 20px
+    'text-title-l-mobile-font-size': 1.25rem
```
ribeiroguilherme, line 14: Wouldn't this be `text-title-m-font-size` instead of `text-title-l-mobile-font-size`?

ribeiroguilherme (approved):
Click to Pay UI seems fine

packages/lib/src/components/internal/FormFields/FormFields.scss
longyulongyu: Yeah the position of the icon is been pushed down, not sitting in the center of the outer `<span>`. I think it's because the font scaling impacts the line height of the span which pushes the inline elements like `` down. When adds the `block` to the img, it doesn't align based on the line height, and takes up the whole space of `<span>`.

packages/lib/src/styles/overrides.scss
longyulongyu: Ah you're right. Will make the change.

Issue #2968, opened 2024-11-19
Adyen Custom Card Secured fields


kxbui-accesso:
**Is your feature request related to a problem? Please describe.**
I'm following [this](https://docs.adyen.com/online-payments/upgrade-your-integration/migrate-to-web-v6/from-v4/) to migrate Adyen Custom Card Secured fields from v4.9 to v6.5. Notice a few cosmetics issues with the Secured fields and missing callback types from all the events.

**Describe the solution you'd like**
1. The internal iframe shows gray border and display unused spaces.  Currently there's no way to add custom styles to the iframe element. 
![image](https://github.com/user-attachments/assets/9a65075c-5b3c-4168-9c03-ab5091d14950)

2. The library does not export all callback types for Custom Card events
![image](https://github.com/user-attachments/assets/2338671b-3d26-47ed-bbe4-eb6571bc451f)

3. Since I installed the latest version of @adyen/adyen-web, I started seeing some issues with ApplePayJS type definition from @types/applepayjs not being found. 
![image](https://github.com/user-attachments/assets/6458cc8c-7c17-4c92-94a1-f04d00787bd3)

Let me know if you need more information about the issue.

UPDATE: Number 1 was fixed by including the stylesheet `import '@adyen/adyen-web/styles/adyen.css'`

kxbui-accesso:
Thanks @ribeiroguilherme for the responses
#2 Our project uses strict types so it requires the types of these callbacks. It would be great if you can expose these types so it doesn't complain about missing types.
#3 Yes we're installing `@types/applepayjs` in our project. The version is `14.0.8`

ribeiroguilherme:
Hi @kxbui-accesso ,

The fix for the #2 is on the way.

About the #3 - can you provide a minimal setup on https://stackblitz.com/ with your Typescript setup , so I can test the issue myself? It will help a lot to troubleshoot it

sponglord:
@kxbui-accesso - your second problem (callback types for Custom Card events) has been addressed in our latest release.

I'm now going to close this issue

For you third problem (ApplePayJS type definition), if it still persists, can you please create a new issue.
We prefer to have each issue only relating to one, specific, problem

thanks

PR #3068, opened 2025-01-08
Improve types for CustomCard

sponglord:
## Summary
More explicitly declare the types for the configuration options on the CustomCard

**Fixed issue**:  #2968

### DIFF
--- a/packages/lib/src/components/CustomCard/types.ts
+++ b/packages/lib/src/components/CustomCard/types.ts
@@ -1,28 +1,174 @@
-import { CardConfiguration } from '../Card/types';
-import { SFError } from '../Card/components/CardInput/types';
+import { CardBrandsConfiguration } from '../Card/types';
+import { Placeholders, SFError } from '../Card/components/CardInput/types';
 import UIElement from '../internal/UIElement';
+import {
+    CbObjOnAllValid,
+    CbObjOnBinLookup,
+    CbObjOnBinValue,
+    CbObjOnBrand,
+    CbObjOnConfigSuccess,
+    CbObjOnFieldValid,
+    CbObjOnFocus,
+    CbObjOnLoad,
+    StylesObject
+} from '../internal/SecuredFields/lib/types';
+import Language from '../../language';
 
-export type CustomCardConfiguration = Omit<
-    CardConfiguration,
-    | 'clickToPayConfiguration'
-    | '_disableClickToPay'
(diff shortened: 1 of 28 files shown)


packages/lib/src/components/CustomCard/types.ts
```diff
@@ -1,28 +1,174 @@
...
+import {
+    CbObjOnAllValid,
+    CbObjOnBinLookup,
+    CbObjOnBinValue,
+    CbObjOnBrand,
+    CbObjOnConfigSuccess,
+    CbObjOnFieldValid,
+    CbObjOnFocus,
+    CbObjOnLoad,
+    StylesObject
```
ribeiroguilherme, line 13: I think we could also fix part of the #2968 issue by exposing these Custom Card types.

For that, we would need to add to the [main exposed type file](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/types.ts) something like:

```js
export { CbObjOnAllValid, etc..  } from './components/internal/SecuredFields/lib/types';
```



I'd propose also to rename these types to match how we have typed for the main callbacks at the [core/types.ts](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/types.ts#L222) (onSubmit, onAdditionalDetails, etc.. ) . Meaning that `CbObjOnBinLookup` would be renamed to `BinLookupData` , `CbObjOnAllValid` to `AllValidData` and so on..  That matches with our current exposed types

ribeiroguilherme (approved):
Maybe worth adding a changeset to this PR, mentioning that the types were improved?

packages/lib/src/components/CustomCard/types.ts
sponglord: > I think we could also fix part of the https://github.com/Adyen/adyen-web/issues/2968 issue by exposing these Custom Card types.

??
This is exactly the issue that this PR is fixing.
If you run `yarn build` and look at `es/index.d.ts` file you will see that ,with this change, all the types (`CbObjOnAllValid`, `CbObjOnBinLookup` etc)  are now visible

ribeiroguilherme: It is there but it is not being exported. You need to do expose it as I mentioned above

sponglord: ok, got it!

PR #3109, opened 2025-01-27
Added e2e tests for binLookup as it affects cvcPolicy

sponglord:
## Summary
Added e2e tests for `/binLookup` related to "optional" and "hidden" `cvcPolicy`

## Tested scenarios
New tests pass


packages/e2e-playwright/tests/ui/card/binLookup/cvc/card.cvc.policies.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+        // Hidden cvc field
+        await expect(card.cvcField).not.toBeVisible();
+
+        // Reset UI by deleting number
+        await card.deleteCardNumber();
+
+        // Visible, required cvc field
+        await expect(card.cvcField).toBeVisible();
+        await expect(card.cvcField).toHaveClass(/adyen-checkout__field__cvc/);
+        await expect(card.cvcField).not.toHaveClass(/adyen-checkout__field__cvc--optional/);
```
ribeiroguilherme, line 36: Do we need to query by classname? Can't we just check if the input label is visible?

It is usually better to use accessible locators rather than inspecting the code using classnames/ids/etc..

ribeiroguilherme (approved):

PR #3117, opened 2025-02-06
Chore/migrate branding reset tests

sponglord:
## Summary
Migrated tests around resetting state and UI after (dual)branding has occurred

## Tested scenarios
new tests all pass


packages/e2e-playwright/tests/ui/card/binLookup/branding/branding.reset.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        // Check brand has been set paymentMethod data
+        expect(cardData.paymentMethod.brand).toBe('mc');
+
+        const responsePromise = page.waitForResponse(response => response.url().includes('/binLookup') && response.request().method() === 'POST');
+
+        // Paste in card unrecognised by /binLookuo but which our regEx recognises as Visa
+        await card.fillCardNumber(UNKNOWN_BIN_CARD_REGEX_VISA);
+
+        await responsePromise;
+        await page.waitForTimeout(100); // Sadly needed to avoid flakiness on the CI
```
longyulongyu, line 24: I think we don't need a fixed timeout, we can just wait for the brand to be undefined.
```js
await page.waitForFunction(() => window.component.data.paymentMethod.brand === undefined);
```
Also I think we don't need to wait for the `responsePromise`, it should be covered by it👆

longyulongyu (approved):

packages/e2e-playwright/tests/ui/card/binLookup/branding/branding.reset.spec.ts
sponglord: Fixed

PR #3125, opened 2025-02-13
Fastlane - minor UI improvements

ribeiroguilherme:
## Summary
- Tooltip icon moved to be next to the text
- Modal does not take full screen anymore when presented in mobile view
- Merchant can now inject telephone number by passing it as configuration


packages/lib/src/components/Card/components/Fastlane/FastlaneSignup.scss
```diff
@@ -35,6 +35,8 @@
 }
 
 .adyen-checkout__button.adyen-checkout__button--fastlane-info-modal {
+    vertical-align: bottom;
+    margin-left: 4px;
```
longyulongyu, line 39: Is there a relevant bento token for it?

longyulongyu (approved):

PR #3130, opened 2025-02-17
Apple Pay - Third-party browser support

ribeiroguilherme:
## Summary
Adding support for third-party browsers

The main changes are located in the `ApplePay.jsx` and `ApplePaySdkLoader.ts` files
All the remaining files are tests, or the functions from the `utils.ts` file that I moved to separated files.

The changes highlights are:
- Now we have to load Apple Pay SDK in order to use it, regardless of the browser.
- The Apple Pay SDK must be loaded as soon as possible, since it is required for the  `isAvailable()` check
- The Apple Pay button now uses custom web element instead of just CSS
- If the checkout is hosted within an iframe, the Apple Pay Code is displayed as a pop-up instead of overlay modal. To force it to be rendered as overlay modal, the property `renderApplePayCodeAs` needs to be configured

The Apple Pay button and the Apple Pay Code can be seen on Storybook if HTTPS is configured.

## Tested scenarios
- Added tests, and tested on custom playground

**Fixed issue**:


packages/lib/src/components/ApplePay/services/ApplePaySdkLoader.ts
```diff
@@ -0,0 +1,29 @@
...
+
+class ApplePaySdkLoader {
+    private sdkLoadingPromise: Promise<void>;
+
+    public async load(): Promise<ApplePaySession> {
+        try {
+            const scriptElement = new Script(APPLE_PAY_SDK_URL, 'body', { crossOrigin: 'anonymous' });
+            this.sdkLoadingPromise = scriptElement.load();
+            await this.sdkLoadingPromise;
+            return window?.ApplePaySession;
```
longyulongyu, line 14: Maybe check if `window?.ApplePaySession` exists before returning it? If it's not, throws an error.

packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -91,8 +108,78 @@ class ApplePayElement extends UIElement<ApplePayConfiguration> {
...
+     * and upstreaming should still be considered.
+     *
+     * {@link https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/4440085-applepaycapabilities}
+     * @param merchantIdentifier
+     */
+    public async applePayCapabilities(merchantIdentifier?: string): Promise<ApplePayJS.PaymentCredentialStatusResponse> {
+        const identifier = merchantIdentifier || this.props.configuration.merchantId;
+
+        try {
+            await this.sdkLoader.isSdkLoaded();
```
longyulongyu, line 129: I am not very familiar with the Apple Pay flow, but we can reuse `await this.isAvailable();` here, no?

longyulongyu (approved):

packages/lib/src/components/ApplePay/ApplePay.tsx
ribeiroguilherme: We can't unfortunately.

The `isAvailable()` and this `applePayCapabilities()` are meant for the same thing (checking if shopper can use ApplePay), although the `applePayCapabilities()`  has more granular response ([you can check it here](https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/4440085-applepaycapabilities)) .

If we use `isAvailable()` and it returns false, the merchant won't be able to use the use the `applePayCapabilities()`  API directly.

longyulongyu: Alright, let's keep it as is.

packages/lib/src/components/ApplePay/services/ApplePaySdkLoader.ts
ribeiroguilherme: I think throwing an error here is a bit scary and can lead to inconsistencies 😅  

It might be that there is a delay to have the SDK instance in the window, and then we would end up throwing an error although it might get available at some point. 

I tested on slow network and it seems to always be added directly once the script is loaded, but we never know the changes done to their SDK and how that can impact our flow .

Perhaps in the future we could use some sort of pooling mechanism check , so after the Script is loaded, we check for certain time if the instance is available on the window or not before resolving this promise.

PR #3151, opened 2025-02-20
GooglePay - Fixing 'onClick' reject

### LINKED ISSUE #3140: [GooglePay] OnClick event doesn't work as expected and docs are incorrect
**Describe the bug**

This is for the GooglePay web component, also applies to the drop in method.

The `onClick` event is incorrectly documented where it [says a simple `resolve, reject`](https://docs.adyen.com/payment-methods/google-pay/web-drop-in/#events).
The `resolve` in fact doesn't expect anything just a call. However the `reject` method throw error if you `reject` something because of a incorrect implementation.

[This is the line in question](https://github.com/Adyen/adyen-web/blob/7991ada23aafe590b4d98f6affb0bf8fba06a8d2/packages/lib/src/components/GooglePay/GooglePay.tsx#L93)

**To Reproduce**
1) Create a instant payment button for Google Pay
2) adds a `onClick` with a `reject` call, without any params:
```js
// Pseudo-code
onClick(resolve, reject) => {
   ...
    reject()
}
```

First the type documented is wrong because the code expect an `error: google.payments.api.PaymentsError`, and then try to read `statusCode` of `undefined`. As in the docs doesn't expect any param.

Also the logic inside of the catch has a flaw. You can see that it expect a `CANCELED` which isn't a valid `statusCode` for the type in the `catch` signature.

**Expected behavior**
Properly handle the `reject` without throw any exception.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: IOS, but probably happening in any OS
 - Browser: Chrome, but probably happening in any browser
 - Version: 133.0.6943.98, but probably happening in any newest versions where Adyen components are expected to work

**Smartphone (please complete the following information):**
 Didn't tested, but definitely would have a similar behaviour as It's a Javascript basic error.

ribeiroguilherme:
## Summary
Calling reject() inside 'onClick' throwns an unhandled exception. This PR fixes the issue.

Fixes #3140

### DIFF
--- a/.changeset/afraid-clocks-drop.md
+++ b/.changeset/afraid-clocks-drop.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+GooglePay - Fixed issue where calling reject on onClick event was triggering an unhandled exception
--- a/packages/lib/src/components/GooglePay/GooglePay.test.ts
+++ b/packages/lib/src/components/GooglePay/GooglePay.test.ts
@@ -55,6 +55,38 @@ beforeEach(() => {
 });
 
 describe('GooglePay', () => {
+    describe('onClick()', () => {
+        test('should not call "initiatePayment" if the onClick reject() is called', async () => {
+            const googlepay = new GooglePay(global.core, {
+                onClick(resolve, reject) {
+                    reject();
+                }
+            });
+            googlepay.submit();
+
+            await new Promise(process.nextTick);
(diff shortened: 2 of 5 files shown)


sponglord (approved):
Is this anything that could be e2e tested?

ribeiroguilherme:
> Is this anything that could be e2e tested?

I added unit tests for the onClick callback.

PR #3156, opened 2025-02-21
Add stored payment logic to PayTo

m1aw:
## Summary
This PR does some clean up for the PayTo component. Also introduces stored component logic.

* Changes segmented controller to manage the aria expanded state
* Adds PayTo to the list of supportedStoredPaymentMethods
* Changes the PayTo stories to send mandate via backend calls (`/sessions` & `/payments`)
  * These mechanism can be used for any other future payment methods that need backend call data
* Adds test to cover some more of PayTo code
* Adds a missing translation

## Tested scenarios

**Fixed issue**:


packages/lib/src/components/internal/SegmentedControl/SegmentedControl.tsx
```diff
@@ -53,14 +70,16 @@ function SegmentedControl<T>({ classNameModifiers = [], selectedValue, disabled
...
-            {options.map(({ label, value, htmlProps }) => (
+            {options.map(({ label, value, controls, htmlProps }: NewSegmentedControlOption<T>) => (
                 <button
                     disabled={disabled}
                     key={value}
                     onClick={(event: MouseEvent) => onChange(value, event)}
                     className={cx('adyen-checkout__segmented-control-segment', {
                         'adyen-checkout__segmented-control-segment--selected': selectedValue === value
                     })}
+                    aria-controls={controls}
```
longyulongyu, line 81: Looks good, like the refactoring!

packages/lib/src/components/PayTo/components/PayToComponent.tsx
```diff
@@ -50,6 +29,24 @@ export default function PayToComponent(props: PayToComponentProps) {
 
     const [status, setStatus] = useState<UIElementStatus>('ready');
 
+    const inputOptions: SegmentedControlOptions<PayToInputOption> = useMemo(
```
longyulongyu, line 32: Do we need to use `useMemo`? It doesn't feel like an expensive calculation to me.

longyulongyu (approved):

packages/server/translations/ru-RU.json
```diff
@@ -102,6 +102,7 @@
     "boleto.sendCopyToEmail": "Отправить мне копию на эл. почту",
     "button.copy": "Копия",
     "button.download": "Загрузить",
+    "boleto.socialSecurityNumber": "CPF/CNPJ",
```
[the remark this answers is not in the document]
m1aw: It wasn't intentional, I think there's a job on Smartling now, but I was harder for me to exclude the changes. It should probably go all together in a new fix, since some languages like pt-BR are still missing.

packages/lib/src/components/PayTo/components/PayToComponent.tsx
m1aw: No, but I thought it was better then just recreating the object every render. This way we get the same reference no?

PR #3157, opened 2025-02-24
[phase 1] Pix biometrics enrollment

longyulongyu:
## Summary
- This PR is a draft for the enrollment flow
- The overall flow for enrollment is: 
Shopper clicks the enroll button on the merchant's website
-> Shopper redirects to the hosted checkout page
-> Shopper selects an issuer
-> Shopper redirects to the issuer's app or website to confirm the enrollment mandate
-> Shopper is redirected back to the hosted page 
-> Shopper creates passkeys and is challenged with biometrics 
-> Enrollment is done, shopper is redirected back to merchant's page 
Test the flow with mocks on https://localhost:3020/iframe.html?args=&id=components-paybybankpix--merchant-page
- All the todos will be fixed in the next PRs
- Added extra callback `afterAdditionalDetails` in the core. This is necessary because when the user is redirected back to the hosted checkout page from the issuer, the details call returns an action. We need to expose this action as a UIElement so it can be mounted. This action does not represent the final state of the payment, so we must interrupt the flow to prevent it from reaching the `onPaymentCompleted` or `onPaymentFailed` stages.
- Load the passkey sdk via cdn


packages/lib/src/components/PayByBankPix/PayByBankPix.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+        const issuer = this.state.data?.issuer ? { issuer: this.state.data?.issuer } : {};
+        const riskSignals = this.state.data?.riskSignals ? { riskSignals: this.state.data.riskSignals } : {};
+        return {
+            paymentMethod: { type: TxVariants.paybybank_pix, ...issuer, ...riskSignals }
+        };
+    }
+
+    async createEnrollment(enrollment: Enrollment) {
+        const { action = {} } = await postEnrollment({ enrollment, clientKey: this.props.clientKey, loadingContext: this.props.loadingContext });
```
m1aw, line 62: This is not a request for change just a discussion.

I'm wondering if this enrolment backend call could have not ben a regular `/payment` call that resolves to this endpoint on the backend. I say this because this flow seems somewhat similar to PayTo, where we are basically registering a shopper to make `CardOnFile` stored transactions correct?

I'm just wondering if this should have been kept consistent on the LPM side.

packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
```diff
@@ -0,0 +1,81 @@
...
+        if (registrationOptions) return;
+
+        const { enrollmentId, clientKey } = props as AwaitProps;
+        const response = await getEnrollmentStatus({ enrollmentId, clientKey, loadingContext });
+        if (response.registrationOptions) {
+            setRegistrationOptions(response.registrationOptions);
+        }
+        return response;
+    };
+    const isAwait = (props: EnrollmentProps): props is AwaitProps => props.type === 'await';
```
m1aw, line 34: I think maybe it's worth making it more clear why "isWait".

I got a bit confused why this logic was down here. As typically we switch between payment method "modes"/"behaviour" in the UIElement.

If I understand correctly this is just the state after the IssuerList ".submit()".

packages/lib/src/components/PayByBankPix/components/Enrollment/components/AwaitLogoContainer.tsx
```diff
@@ -0,0 +1,21 @@
+import { h } from 'preact';
+import './AwaitLogoContainer.scss';
+import useImage from '../../../../../core/Context/useImage';
+
+// todo: add logos after uploaded to cdn
+const LOGO_MAPPING = {
+    openFinance: { name: 'pix', alt: 'Open Finance' },
```
m1aw, line 7: Don't forget to move the alt to i18n.

packages/lib/src/components/PayByBankPix/components/Enrollment/useIssuerWithLogo.ts
```diff
@@ -0,0 +1,29 @@
...
+import { IssuerItem } from '../../../internal/IssuerList/types';
+import getIssuerImageUrl from '../../../../utils/get-issuer-image';
+import { useCoreContext } from '../../../../core/Context/CoreProvider';
+import useImage from '../../../../core/Context/useImage';
+
+interface UseIssuerWithLogoProps {
+    txVariant: string;
+    issuers: IssuerItem[];
+}
+export const useIssuerWithLogo = ({ issuers, txVariant }: UseIssuerWithLogoProps): IssuerItem[] => {
```
m1aw, line 11: Is this being re-implemented because we cannot use the logic for images that already exists for IssuerListContainer?

packages/lib/src/components/PayByBankPix/components/PayByBankPix/PayByBankPix.tsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }, [passKeyInitError]);
+
+    // todo: uncomment it!
+    /*    if (!passkeyService)
+        return null;
+    }*/
+
+    return shouldEnroll ? (
+        <Enrollment
```
m1aw, line 76: Why not move this logic up to UIElement level?

packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
```diff
@@ -0,0 +1,81 @@
...
+import IssuerListIntroduction from './components/IssuerListIntroduction';
+import AwaitLogoContainer from './components/AwaitLogoContainer';
+import './Enrollment.scss';
+
+function Enrollment(props: EnrollmentProps) {
+    const { i18n, loadingContext } = useCoreContext();
+    const issuerListRef = useRef<IIssuerList>(null);
+    const [registrationOptions, setRegistrationOptions] = useState<string>(null);
+
+    this.showValidation = () => {
```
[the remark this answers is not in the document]
longyulongyu: Let me take a look, it should be possible.

packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
longyulongyu: It causes typescript not able to narrow the proper type `AwaitProps `, one way is to use the `isAwait` assist typescript to narrow it down. I couldn't find a better way but i'll put a comment here.

The reason i didn't put too much logic inside the UIElement is i want to group a few preact components which belong to the same business flow together. So the issuer list and await belong to the enrollment flow so they sit inside the enrollment preact component.

packages/lib/src/components/PayByBankPix/components/PayByBankPix/PayByBankPix.tsx
longyulongyu: I feel UIElement should just deal with very simple props & data transformation - so `formatData` and `formatProps`, and the preact components should know the business logic (in my case interact with the passkey) and set proper state for the `formatData` to pick them up. So I abstract them into a preact component.

packages/lib/src/components/PayByBankPix/components/Enrollment/useIssuerWithLogo.ts
longyulongyu: I wanted to, but IssuerListContainer is an UIElement..

m1aw: Yeah, I understood. 
If we have time I would say it's worth extracting the code, otherwise let's keep it like this.

packages/lib/src/components/PayByBankPix/components/PayByBankPix/PayByBankPix.tsx
m1aw: I disagree with the approach a little bit.
I don't think preact components should have much payment specific logic. That's place of the UI Element. 
Do we have any other components? That a similar UI setup where different views of the payment method are handled inside preact components?

m1aw (approved):

packages/lib/src/components/PayByBankPix/PayByBankPix.tsx
longyulongyu: I asked LPM, let's wait for his reply. In case of any fixes, I'll put them in the next PR.

PR #3174, opened 2025-03-10
[phase 2]: refactor the code & add payment draft flow

longyulongyu:
## Summary
- Moved the business logic to the UIElement
- Added the payment draft flow
- Added some comments
- Added the related code for PasskeyService and PasskeyLoader

## Tested scenarios
- enrollment flow with mock https://localhost:3020/iframe.html?globals=&args=&id=components-paybybankpix--merchant-page&viewMode=story
- change the pollStatus of the redirectUrl on the Simulated issuer's page: **pollStatus=complete** to trigger passkey creation


packages/lib/src/components/PayByBankPix/PayByBankPix.tsx
```diff
@@ -20,28 +23,49 @@ const hasRedirectResult = (): boolean => {
 class PayByBankPixElement extends UIElement<PayByBankPixConfiguration> {
     public static type = TxVariants.paybybank_pix;
     private static TIMEOUT_MINUTES = 1;
+    private readonly passkeyService: PasskeyService;
 
     public static defaultProps: PayByBankPixConfiguration = {
         showPayButton: true,
-        _isAdyenHosted: window.location.hostname.endsWith('adyen.link') || hasRedirectResult(), // todo: remove hasRedirectResult
+        _isAdyenHosted: window.location.hostname.endsWith('adyen.com') || hasRedirectResult(), // todo: remove hasRedirectResult
```
ribeiroguilherme, line 30: This might break SSR apps.

We had this issue with [Apple Pay release few weeks ago](https://github.com/Adyen/adyen-web/pull/3204), where we were using the `window` instance as part of the static defaultProps, and doing that causes the SSR frameworks to complain/not compile properly.

longyulongyu: Nice catch, I'll add a safe check 
```ts
typeof window !== 'undefined' && window.location.hostname.endsWith('adyen.com')
```

PR #3175, opened 2025-03-10
[WIP] Fastlane analytics

ribeiroguilherme:
## Summary
- Adding analytic events to Fastlane. For more context of the events being added here, check COWEB-1487

### DIFF
--- a/packages/lib/src/components/Card/components/Fastlane/FastlaneSignup.tsx
+++ b/packages/lib/src/components/Card/components/Fastlane/FastlaneSignup.tsx
@@ -1,5 +1,5 @@
 import { Fragment, h } from 'preact';
-import { useEffect, useMemo, useState } from 'preact/hooks';
+import { useCallback, useEffect, useMemo, useState } from 'preact/hooks';
 import cx from 'classnames';
 import Toggle from '../../../internal/Toggle';
 import Img from '../../../internal/Img';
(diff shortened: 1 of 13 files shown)


packages/lib/src/core/Analytics/utils.ts
```diff
@@ -192,7 +195,15 @@ export const getCardConfigData = (cardProps: CardConfiguration): CardConfigData
         hasOnLoad: onLoad !== CardInputDefaultProps.onLoad,
         // Card level props
         hasOnBinLookup: !!onBinLookup,
-        hasOnEnterKeyPressed: !!onEnterKeyPressed
+        hasOnEnterKeyPressed: !!onEnterKeyPressed,
+        /**
+         * Fastlane
+         */
+        // CHECK: Do we always append the prop even if it is not used? e.g. merchants not using fastlane
```
ribeiroguilherme, line 202: Not sure what is the best here

sponglord: For other, _universal_, props we won't to know what config the merchant has ended up with, be it something they've set or a default that they've picked up.
But for something opt-in like Fastlane, we should only send it if it's relevant, imo

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -126,6 +126,12 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
         this.setState({
             showDefaultPaymentMethodList: true
         });
+
+        this.props.modules?.analytics.sendAnalytics('dropin', {
+            type: ANALYTICS_EVENT.info,
+            infoType: InfoEventTypes.clicked,
+            target: 'otherpaymentmethod_button'
```
sponglord, line 133: Similar to my comment above - can the backend handle this target value, "otherpaymentmethod_button" without errors?

packages/lib/src/components/Card/components/Fastlane/FastlaneSignup.tsx
```diff
@@ -50,6 +53,20 @@ const FastlaneSignup = ({
...
+    const handleToggleChange = useCallback(() => {
+        const newValue = !isChecked;
+        setIsChecked(newValue);
+
+        onSubmitAnalytics({
+            type: ANALYTICS_EVENT.info,
+            infoType: InfoEventTypes.clicked,
+            target: 'fastlane_signup_consent_toggle',
+            configData: {
+                isToggleOn: newValue
```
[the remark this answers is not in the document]
ribeiroguilherme: The requests are working fine

PR #3188, opened 2025-03-14
Fixes small a11y problems

m1aw:
## Summary
Small fixes long description to document this changes for the future.

### Constrast changes

These are documented in COWEB-1484. However to summarise the issue is that the `color-label-tertiary` never has enough contrast, even against a white background, so here I just override it via our scss token loader. The same happens for the red color used for the error messages. 

### Aria expands attribute on radio buttons

Basically this change keeps being requested by merchants who have automated tools. Currently `aria-expands` on radio elements is still not part of the official aria spec. See the proposal here: https://github.com/w3c/aria/issues/1404

So for now we are going to disable this feature. The objective is to enable it soon as the proposal get's approved.

### SVG alt description

This is a 2 in 1, we remove `aria-describedby` from the SVG and also add additional information to both the CVC hint image and the expiry date hint image.

To start with, I could not find anywhere any documentation supporting the use of `aria-describedby` for this use case. The alternatives found in documentation for alt text on SVGs are `aria-label` / `aria-labelledby` or svg `<title>`.  We are going with `title`. 

Further more, there's been a discussion between us and merchants where some auditors are marking these images alt text as not meaningful (enough). So we are also adding the context text to the image, which will hopefully be enough information for the user. This information might turn out to be to much but we will iterate on this.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/.changeset/quiet-berries-wonder.md
+++ b/.changeset/quiet-berries-wonder.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+A11y improvements: contrast, aria-expands and svg alt text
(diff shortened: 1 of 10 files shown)


sponglord (approved):
Looks good to me. Just asking Yu to check the `variable-generator.scss` change

longyulongyu (approved):

PR #3192, opened 2025-03-17
fix: storybook loads translations locally on ci for e2e tests

longyulongyu:
## Summary
This PR fixed the issue that Storybook tries to load translations from CDN on CI running e2e tests. This should not be a problem except when we perform a release on the `changeset-release/main` branch. As we are changing the package.json version and ofc the translation files are not available on the CDN for the version that needs to be released.

The issue occurs when we try to stop msw from initiating on the storybook "production" mode.

This PR specified the translation url for storybook. When storybook is running on CI for e2e tests, we load translations locally. However this will introduce another issue in the future when we publish Storybook on Netlify, hence we need to know if the Storybook is on Netlify by introducing a new process environment variable `NETLIFY`. When it is on netlify, we load translations from remote.

## Tested scenarios
All e2e tests pass.

### DIFF
--- a/packages/lib/.storybook/preview.tsx
+++ b/packages/lib/.storybook/preview.tsx
@@ -1,6 +1,6 @@
 import './main.css';
 import { Preview } from '@storybook/preact';
-import { DEFAULT_COUNTRY_CODE, DEFAULT_SHOPPER_LOCALE, DEFAULT_AMOUNT_VALUE, SHOPPER_LOCALES, protocol } from '../storybook/config/commonConfig';
+import { DEFAULT_COUNTRY_CODE, DEFAULT_SHOPPER_LOCALE, DEFAULT_AMOUNT_VALUE, SHOPPER_LOCALES } from '../storybook/config/commonConfig';
 import { initialize, mswLoader } from 'msw-storybook-addon';
 
 const isDev = process.env.NODE_ENV === 'development';


packages/lib/.storybook/preview.tsx
```diff
@@ -1,13 +1,16 @@
...
 import { initialize, mswLoader } from 'msw-storybook-addon';
 
 const isDev = process.env.NODE_ENV === 'development';
 if (isDev) {
     initialize({ onUnhandledRequest: 'bypass' });
 }
 const loaders = isDev ? { loaders: [mswLoader] } : {};
+// When storybook is running locally or on CI for e2e tests, we load translations locally.
+// When storybook is published to Netlify, we should set the Netlify process env and Storybook will load translations remotely from the cdn server.
+const translationUrlConfig = process.env.NETLIFY !== 'true' ? { _environmentUrls: { cdn: { translations: `${protocol}//localhost:3030/` } } } : {};
```
ribeiroguilherme, line 13: Do we need to pass the full URL here? I think we can just pass `/`

ribeiroguilherme (approved):

packages/lib/.storybook/preview.tsx
longyulongyu: > Do we need to pass the full URL here? Can't we just pass `/`? I expected the proxy settings to allow us to do that. Then we make sure we don't depend on the port

It seems working, i'll just change it and let it run on ci.

PR #3200, opened 2025-03-20
allow-custom-billing-address-specifications

M-Izadmehr:
## Summary
Adyen-web drop-in can get `specifications` which can provide address specifications. This can be used to specifically determine for each country, which bill address fields are needed, and what are specific details needed (label/placeholder/optional/etc).

However, in the code, the `specifications` is not actively used, and finally, when renderingthe  Credit Card Address component, the main criteria depend on `billingAddressMode` only. And the specifications is ignored.

Here is a summary of data flow:

1. In the Credit Card components in `CardInput.tsx`  ([github](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/Card/components/CardInput/CardInput.tsx#L469)). While the specifications is used it is not passed to `FieldsToRendered`

```
<FieldsToRender
   {...restOfProps}
    partialAddressSchema={partialAddressSchema}
/>
```

2. in case `FieldsToRender` uses `CardFieldsWrapper`
3. `CardFieldsWrapper` will use the `partialAddressSchema` as `specifications` and completely ignores specifications you pass in ([github](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/Card/components/CardInput/components/CardFieldsWrapper.tsx#L171))

```
            {billingAddressRequired && (
                <Address
                    ...rest
                    specifications={partialAddressSchema}
                />
            )}
```

so it either uses the hardcoded partial form (meaning only postal code) or `null`.

This MR will make sure that if specifications are passed, it is preferred over default specifications.

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.tsx
@@ -92,7 +92,7 @@ const CardInput: FunctionalComponent<CardInputProps> = props => {
 
     const showBillingAddress = props.billingAddressMode !== AddressModeOptions.none && props.billingAddressRequired;
 
-    const partialAddressSchema = handlePartialAddressMode(props.billingAddressMode);
+    const partialAddressSchema = handlePartialAddressMode(props.billingAddressMode, specifications.getSpecifications());
     // Keeps the value of the country set initially by the merchant, before the Address Component mutates it
     const partialAddressCountry = useRef<string>(partialAddressSchema && props.data?.billingAddress?.country);
 
--- a/packages/lib/src/components/Card/components/CardInput/types.ts
+++ b/packages/lib/src/components/Card/components/CardInput/types.ts
@@ -4,8 +4,7 @@ import { AddressData, PaymentAmount } from '../../../../types';
 import { InstallmentOptions } from './components/types';
 import { ValidationResult } from '../../../internal/PersonalDetails/types';
 import { CVCPolicyType, DatePolicyType, CbObjOnError, StylesObject } from '../../../internal/SecuredFields/lib/types';
-import Specifications from '../../../internal/Address/Specifications';
-import { AddressSchema } from '../../../internal/Address/types';
+import { AddressSchema, AddressSpecifications } from '../../../internal/Address/types';
 import { Resources } from '../../../../core/Context/Resources';
 import { SRPanel } from '../../../../core/Errors/SRPanel';
 import RiskElement from '../../../../core/RiskModule';
@@ -122,7 +121,7 @@ export interface CardInputProps {
     showPayButton?: boolean;
     showStoreDetailsCheckbox?: boolean;
     showWarnings?: boolean;
-    specifications?: Specifications;
+    specifications?: AddressSpecifications;
     storedPaymentMethodId?: string;
     styles?: StylesObject;
     trimTrailingSeparator?: boolean;
--- a/packages/lib/src/components/Card/components/CardInput/utils.ts
+++ b/packages/lib/src/components/Card/components/CardInput/utils.ts
@@ -166,7 +166,10 @@ export const extractPropsForSFP = (props: CardInputProps) => {
     } as SFPProps; // Can't set as return type on fn or it will complain about missing, mandatory, props
 };
 
-export const handlePartialAddressMode = (addressMode: AddressModeOptions): AddressSpecifications | null => {
+export const handlePartialAddressMode = (addressMode: AddressModeOptions, specifications: AddressSpecifications): AddressSpecifications | null => {
+    if (specifications) {
+        return specifications;
+    }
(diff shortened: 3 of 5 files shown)


packages/lib/src/components/Card/components/CardInput/CardInput.tsx
```diff
@@ -92,7 +92,7 @@ const CardInput: FunctionalComponent<CardInputProps> = props => {
 
     const showBillingAddress = props.billingAddressMode !== AddressModeOptions.none && props.billingAddressRequired;
 
-    const partialAddressSchema = handlePartialAddressMode(props.billingAddressMode);
+    const partialAddressSchema = handlePartialAddressMode(props.billingAddressMode, specifications.getSpecifications());
```
M-Izadmehr, line 95: this will allow us to use the custom `specifications` properly

packages/lib/src/components/Card/components/CardInput/types.ts
```diff
@@ -122,7 +121,7 @@ export interface CardInputProps {
     showPayButton?: boolean;
     showStoreDetailsCheckbox?: boolean;
     showWarnings?: boolean;
-    specifications?: Specifications;
+    specifications?: AddressSpecifications;
```
M-Izadmehr, line 124: fixed the type here, the value of `specifications` coming from props is not the class rather its the `AddressSpecifications` object

packages/lib/src/components/Card/components/CardInput/utils.ts
```diff
@@ -166,7 +166,10 @@ export const extractPropsForSFP = (props: CardInputProps) => {
     } as SFPProps; // Can't set as return type on fn or it will complain about missing, mandatory, props
 };
 
-export const handlePartialAddressMode = (addressMode: AddressModeOptions): AddressSpecifications | null => {
+export const handlePartialAddressMode = (addressMode: AddressModeOptions, specifications: AddressSpecifications): AddressSpecifications | null => {
+    if (specifications) {
```
M-Izadmehr, line 170: this is the main addition to this MR, in case we get specifications, we can prefer it over `addressMode`

packages/lib/src/components/internal/Address/Specifications.ts
```diff
@@ -5,10 +5,14 @@ import { AddressField } from '../../../types';
 const SCHEMA_MAX_DEPTH = 2;
 
 class Specifications {
-    private specifications: AddressSpecifications;
+    private readonly specifications: AddressSpecifications;
```
M-Izadmehr, line 8: added for safety

packages/lib/src/components/internal/Address/Specifications.ts
```diff
@@ -5,10 +5,14 @@ import { AddressField } from '../../../types';
...
 
-    constructor(specifications?) {
-        this.specifications = { ...ADDRESS_SPECIFICATIONS, ...specifications };
+    constructor(specifications?: AddressSpecifications) {
+        if (specifications && typeof specifications === 'object') {
```
M-Izadmehr, line 11: add a small check here to make sure the input type is correct

packages/lib/src/components/internal/Address/Specifications.ts
```diff
@@ -86,6 +90,14 @@ class Specifications {
             .flat(SCHEMA_MAX_DEPTH)
             .filter((element): element is AddressField => typeof element === 'string');
     }
+
+    /**
+     * Returns the specifications object.
+     * @returns specifications - returns combines specifications
+     */
+    getSpecifications(): AddressSpecifications {
```
M-Izadmehr, line 98: added this new getter to give the merged specification

M-Izadmehr:
Can you please guide me to find what is the best approach here to add this change to both V5 and V6?

m1aw:
Hey @M-Izadmehr 

If I understand correctly the objective of the PR is to allow passing `specification` property to the card component so you can customise the billing address form?

M-Izadmehr:
> If I understand correctly the objective of the PR is to allow passing `specification` property to the card component so you can customise the billing address form?

Hey @m1aw, exactly. 

In our case, we noticed our users from some countries are not used to having some of the fields. In this case for example for `NL` we want to have houseNumber as required, however for `TW` to keep it as optional and change its label

`specification` seemed to support this use case very well

m1aw:
@M-Izadmehr thank you for the PR. 

We agree that this is a feature we want to expose, we aren't certain if this is something we want to expose in it's current form. We had some changes to address component in the pipeline, we want to ensure this feature is compatible with this.

So for now we are parking this change. We will try to revisit ASAP.

M-Izadmehr:
Hi @m1aw,
do we have any sample or docs on how it's planned to expose this? I'd be more than happy to adjust it to be consistent with it.
Currently lack of customization on the form is proving to be a blocker for us, so I'd like to help finding a solution

PR #3208, opened 2025-03-24
Playwright - Drop-in tests adjustments

ribeiroguilherme:
## Summary
When working with Drop-in tests, we are currently merging Card fixture with Drop-in fixture (using playwright `mergeTests`) and then we use both objects within our tests.

The problem with the approach above is that combining fixtures can lead to its locators to find multiple elements (e.g. multiple cvc input fields). The Card fixture is meant to be used in a page where there is a single Card component and not when there are multiple.

This PR changes the approach how we handle Drop-in tests:
- We stop combining fixtures for such tests. When selecting a payment method within the Drop-in, the Drop-in will return the locator of that specific payment method that was selected. It returns the locator of the payment method content, meaning that when clicking on the `Cards` label, we return a locator pointing to its payment details content
- We can then use that specific locator to instantiate the payment method model (e.g. instantiate the `Card` model using the correct root locator). This approach narrows down the locators search and do not lead to finding multiple locators of the same payment method 

## Tests
- Tested that having a stored card as first payment method does not break the whole e2e test setup


m1aw (approved):
Good change!

PR #3216, opened 2025-03-25
a11y - Fix read all the expiry date errors

m1aw:
## Summary
The objective of this PR is to make the SF errors for Expiry date to be read. This wasn't happen as they are not triggered onBlur but onChange.

## Tested scenarios

- Expiry date error is read once is typed
  - [x] Tested on MacOS VoiceOver Chrome 
  - [ ] Test on other screen readers
- Expiry date error is read a second time it occurs
  - [x] Tested on MacOS VoiceOver Chrome
  - [ ] Test on other screen readers
- Expiry date error is *NOT* read when removed
  - [x] Tested on MacOS VoiceOver Chrome
  - [ ] Test on other screen readers

Test regressions:
- [ ] Triggered each of the non onBlur errors above. Check if they get read out twice.
- [ ] Test screen reader on QRCode to check there's no regression

**Fixed issue**:

### DIFF
--- a/packages/lib/src/core/Errors/SRPanel.tsx
+++ b/packages/lib/src/core/Errors/SRPanel.tsx
@@ -19,7 +19,7 @@ export class SRPanel extends BaseElement<SRPanelProps> {
         showPanel: false,
         id: 'ariaLiveSRPanel',
         ariaAttributes: {
-            'aria-relevant': 'all',
+            'aria-relevant': 'additions',
             'aria-live': 'polite',
             'aria-atomic': 'true'
         }
@@ -104,11 +104,9 @@ export class SRPanel extends BaseElement<SRPanelProps> {
         return (
             <div
                 className={this.showPanel ? 'adyen-checkout-sr-panel' : 'adyen-checkout-sr-panel--sr-only'}
-                role={'log'}
-                {...this.props.ariaAttributes}
                 {...(process.env.NODE_ENV !== 'production' && { 'data-testid': this.id })}
             >
-                <SRMessages setComponentRef={this.setComponentRef} />
+                <SRMessages setComponentRef={this.setComponentRef} customAria={this.props.ariaAttributes} />
             </div>
         );
     }
--- a/packages/lib/src/components/Card/components/CardInput/useSRPanelForCardInputErrors.ts
+++ b/packages/lib/src/components/Card/components/CardInput/useSRPanelForCardInputErrors.ts
@@ -1,7 +1,7 @@
 import { useEffect, useState } from 'preact/hooks';
 import useSRPanelContext from '../../../../core/Errors/useSRPanelContext';
 import { SetSRMessagesReturnFn } from '../../../../core/Errors/SRPanelProvider';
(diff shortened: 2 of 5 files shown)


sponglord:
> We now remove the SRPanel when there's no error messages

What exactly do you mean? Where is this "removal" happening?

m1aw:
@sponglord the removal is happening in `SRMessages.ts` because we have ` return messages ?`.

sponglord:
> @sponglord the removal is happening in `SRMessages.ts` because we have ` return messages ?`.

But that was always there, wasn't it? Your description makes it sound like this is new functionality.

For context: the reason I'm talking about this at all is because your description says
> We now remove the SRPanel when there's no error messages.

packages/lib/src/core/Errors/SRPanel.tsx
```diff
@@ -19,7 +19,7 @@ export class SRPanel extends BaseElement<SRPanelProps> {
         showPanel: false,
         id: 'ariaLiveSRPanel',
         ariaAttributes: {
-            'aria-relevant': 'all',
+            'aria-relevant': 'additions',
```
sponglord, line 22: Can you confirm that all the errors still get read out when the (Card) PM is completely empty and the "Pay" button is pressed?
(I thought that was why I'd used "all" instead of "additions" originally - but it was a long time ago...!)

m1aw:
> But that was always there, wasn't it? Your description makes it sound like this is new functionality.

I guess I didn't make it clear. It was always there but it was only an inner div of the div where aria-live was. Now I moved the aria-live attribute to be after the `message` condition. 

So when message is empty the aria-live gets gets removed.

packages/lib/src/core/Errors/SRPanel.tsx
m1aw: I have just tested the Pay button scenario. It does not change with `aria-relevant`.
However It's not always reading from aria-live. I will double check exactly if is a new behaviour.

sponglord:
> > But that was always there, wasn't it? Your description makes it sound like this is new functionality.
> 
> I guess I didn't make it clear. It was always there but it was only an inner div of the div where aria-live was. Now I moved the aria-live attribute to be after the `message` condition.
> 
> So when message is empty the aria-live gets gets removed.

Does that not break the a11y requirement, for such an "error" panel, that there is a permanent element in the DOM with an `aria-live` attribute?

m1aw:
> Does that not break the a11y requirement, for such an "error" panel, that there is a permanent element in the DOM with an `aria-live` attribute?

It was never the case that it needed to be permanently in the DOM. Was it?

sponglord:
> It was never the case that it needed to be permanently in the DOM. Was it?

It _definitely_ has to be in the DOM at startup, so the screenreader can find and identify it as an area to watch.
That was clear from levelAccess
> [element] with `aria-live=”polite[or whatever]"` should pre-exist in the DOM, in the markup, at page load

Whether after that it has to be permanently present I don't know

Issue #3189, opened 2025-03-17
Incorrect Finnish (fi-FI) Translations (in Credit Card Component)


Kennyoro:
**Describe the bug**
There are incorrect Finnish translations in the Adyen credit card component. Most notably, the "Name on Card" field has a clearly incorrect translation that Finnish customers would immediately recognize as a mistake, along with several other grammar and word choice issues throughout the component.

**To Reproduce**

- Load the Adyen credit card component with Finnish (fi-FI) localization
- Observe the field labels and messages in the component
- Note incorrect translations, particularly the "Name on Card" field which contains a clearly visible translation error

**Expected behavior**
All text elements in the credit card component should display proper Finnish translations that are grammatically correct and use appropriate terminology for payment processing.

**Screenshots**

![Image](https://github.com/user-attachments/assets/301cfa03-f475-4fd3-af8e-11f76cdbb125)

**Desktop (please complete the following information):**

- OS: All operating systems
- Browser: All browsers
- Version: All versions

**Smartphone (please complete the following information):**

- Device: All devices
- OS: All operating systems
- Browser: All browsers
- Version: All versions

**Additional context**
We reviewed and corrected the translations in the JSON file. Our Finnish translator has gone through the entire file and made corrections to known mistakes and other obvious grammar/word choice issues. Without specific context for all lines, they focused on correcting what's relevant to our implementation and what we already know to be incorrect.

The most significant issue is with the "Name on Card" field, which has a visibly incorrect translation. While customers might understand what it means despite the error, it's a clear translation mistake that every Finnish customer would recognize, potentially affecting their trust in our payment process.

We would like to submit these corrected translations to be included in the Adyen library to improve the customer experience for Finnish users.

[fi-FI.json](https://github.com/user-attachments/files/19281612/fi-FI.json)

sponglord:
@Kennyoro - thanks for raising this issue. 
We're now reviewing these translations with our translation service provider and will keep you posted

sponglord:
@Kennyoro - below is a summary of findings from our translation provider, based on your feedback:

- There are 2-3 actual errors [in the original translation file] while most other changes are preferential.
- The [original] translation quality could have been improved, but it was not critically flawed.
- The original translation was largely correct, with only a few areas that could have been optimized.
- Finnish is highly context-dependent, and some [of the suggested changes] do not necessarily improve accuracy.
[...]

PR #3239, opened 2025-04-03
Fix fi-FI translation file

m1aw:
## Summary
Fix errors in Finnish translations. 

## Tested scenarios

**Fixed issue**:  
#3189


sponglord:
What's concerning here is that the problematic translation from issue #3189 for `creditCard.holderName` has not been re-translated

sponglord:
I do notice however that one change Smartling agreed was correct for `ctp.errors.FRAUD`
`Käyttäjätili **oli** lukittu tai poistettu käytöstä` --> `Käyttäjätili **on** lukittu tai poistettu käytöstä`
hasn't made it into the new translation file

PR #3252, opened 2025-04-09
Pay by bank Pix

longyulongyu:
## Summary
- Due to a feature toggle enabled on test, we cannot fully enroll / pay with passkeys. Use my playground site instead:
[Enrollment flow](https://longyulongyu--checkout-playground.netlify.app/?countryCode=BR&amount=0), [Payment flow](https://longyulongyu--checkout-playground.netlify.app/?countryCode=BR&amount=25900)
- ~First enroll your device on https://localhost:3020/iframe.html?args=&globals=&id=dropin-paybybankpix--create-enrollment&viewMode=story~
- ~Then you can pay with the enrolled device https://localhost:3020/iframe.html?args=&globals=&id=dropin-paybybankpix--pay-with-enrolled-device&viewMode=story~
- ~Logos are not available yet and will be uploaded~

## Tested scenarios
Added unit tests.


packages/lib/src/core/core.ts
```diff
@@ -206,16 +207,38 @@ class Core implements ICore {
...
+            .catch((e: PaymentResponseData | Error) => {
+                if (e instanceof CancelError) {
+                    return;
+                }
+                if (e instanceof Error) {
+                    this.options.onError?.(e as AdyenCheckoutError);
+                } else {
+                    cleanupFinalResult(e);
+                    this.options.onPaymentFailed?.(e);
+                }
```
ribeiroguilherme, line 224: The `CancelError` logic seems correct, but I am not sure that we should trigger `onError` here.  Is that a desired behavior?

Seems like before we used to only call `onPaymentFailed`. I am afraid we might start triggering `onError` for certain flows that we are supposed to just trigger the ` onPaymentFailed`.

packages/lib/src/core/core.ts
```diff
@@ -206,16 +207,38 @@ class Core implements ICore {
...
+         * Typically, the response will not include an action object, except for the case of `paybybank_pix` payment method.
+         * In terms of `paybybank_pix`, the action UIElement will be created and passed to the `afterAdditionalDetails` callback, allowing it to be mounted on the page.
+         */
+        if (this.options.afterAdditionalDetails && response?.action) {
+            const actionEle = this.createFromAction(response.action);
+            this.options.afterAdditionalDetails(actionEle);
+            return Promise.reject(new CancelError('Handled by afterAdditionalDetails'));
+        }
+        return Promise.resolve(response);
+    };
```
ribeiroguilherme, line 240: Does it mean that every integration offering this payment method will have to implement this `afterAdditionalDetails` callback?

In what part of the flow is it needed?

packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -76,15 +76,19 @@ function Await(props: AwaitComponentProps) {
...
     };
 
-    const checkStatus = (): void => {
-        const { paymentData, clientKey, throttleInterval } = props;
+    const checkStatus = async (): Promise<void> => {
+        const { paymentData, clientKey, throttleInterval, pollStatus } = props;
+
+        // If there's a custom pollStatus function, call it.
+        // Otherwise, call the default one.
+        const pollStatusFunction = pollStatus ?? (() => checkPaymentStatus(paymentData, clientKey, loadingContext, throttleInterval));
```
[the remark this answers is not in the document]
longyulongyu: Yes, I tested other payment methods using Await, it still works.

packages/lib/src/core/core.ts
longyulongyu: Yes but it's only on the hosted checkout page. Merchants using this payment method are offered to go the hosted checkout page to finalize the flows.
The scenario is when shopper is redirected back from the issuer, the hosted checkout page will make details call, the response is an action, and in the `afterAdditionalDetails ` function we are creating the UIElement from the action object and expose it to the hosted checkout page. So they can mount it somewhere.

packages/lib/src/components/PayByBankPix/PayByBankPix.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+            if (this.props.storedPaymentMethodId) {
+                // If the provided deviceId from the server does not match the localstorage deviceId, do not render the stored payment component.
+                const shouldShowStoredPaymentMethod = await this.passkeyService.canUseStoredCredential();
+                return shouldShowStoredPaymentMethod ? Promise.resolve() : Promise.reject();
+            }
+
+            return Promise.resolve();
+        } catch (error) {
+            this.handleError(
```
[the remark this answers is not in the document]
longyulongyu: Yes that's a good point, will remove this line and just reject the promise.

packages/lib/src/core/core.ts
longyulongyu: Ok I'll remove it to be on the safer side.

packages/lib/src/core/core.ts
ribeiroguilherme: Hunn I see.. Do you think we could try flagging it as an internal callback? 

E.g. marking it as `@internal` in the 'types.ts` and perhaps also prefixing with `_` ( `_afterAdditionalDetails`) . 
I just don't want people to start thinking that they can use it

ribeiroguilherme (approved):

PR #3266, opened 2025-04-17
[WIP] UPI - Callback to validate the VPA

ribeiroguilherme:
## Tested scenarios

**Fixed issue**:


packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -57,6 +60,52 @@ class UPI extends UIElement<UPIConfiguration> {
...
+        const shouldValidateVpa = this.paymentType === TxVariants.upi_collect && this.props.onVpaValidation;
+
+        if (shouldValidateVpa) {
+            this.setElementStatus('loading');
+
+            return new Promise<void>((resolve, reject) => this.props.onVpaValidation(this.state.data.virtualPaymentAddress, { resolve, reject }))
+                .then(() => {
+                    this.submitAnalytics({
+                        type: ANALYTICS_EVENT.info,
+                        infoType: InfoEventTypes.fieldValid
```
longyulongyu, line 92: Maybe add the `target: 'virtualPaymentAddress'` as well for the consistency reason. I think for the card analytics, we send the field name in the target.

ribeiroguilherme: I added it

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -20,39 +22,55 @@ interface UPIComponentProps {
...
-        vpaInputHandlers?.validateInput();
-    };
+    const upiRef = useRef<UpiComponentRef>({
+        setStatus: (status: UIElementStatus) => setStatus(status),
+        showInvalidVpaError: () => vpaInputHandlersRef.current.showInvalidVpaError(),
+        showValidation: () => vpaInputHandlersRef.current.validateInput()
+    });
+
+    useEffect(() => {
+        setComponentRef(upiRef.current);
```
longyulongyu, line 69: Maybe this can be simplified by merging functions from 2 refs:
Inside UPIComponent.tsx:
```js
    const upiRef = useRef<UpiComponentRef>({
        setStatus: (status: UIElementStatus) => setStatus(status)
    });

    useEffect(() => {
        setComponentRef({ ...upiRef.current, ...vpaInputHandlersRef.current });
    }, [setComponentRef, vpaInputHandlersRef]);
```
and inside the VpaInput.tsx, we have already defined 2 functions, we just need to rename the `validateInput` to the "public" function `showValidation`:
```js
    const vpaInputHandlersRef = useRef<VpaInputHandlers>({
        showValidation: () => triggerValidation(),
        showInvalidVpaError: () => {
            const error = { isValid: false, errorMessage: 'upi.collect.field.invalid-id-error', shouldValidate: true };
            setErrors('virtualPaymentAddress', error);
        }
    });
```

The current one is also fine but feel a bit verbose.

longyulongyu (approved):

PR #3290, opened 2025-04-30
Automate a11y tests in playwright

m1aw:
# Automated Accessibility Testing Framework

### 1. Changed the Base Model
- Modified `/packages/e2e-playwright/models/base.ts` to expose `getA11yErrors()`. This should be used if we want to call the a11y test inside specific playwright tests. 

#### New Files:
- **Automated Test Configuration**:
  - Added `/packages/e2e-playwright/automated-a11y.config.ts` used to run the tests in separate pipeline
  
- **Automated Test Suite**:
  - Added `/packages/e2e-playwright/tests/automated-a11y/automated.spec.ts`
  - Dynamically generates tests for all storybook components
  - Excludes problematic components like Apple Pay that can't run in e2e tests
  - Configures known violations by component ID

### 3. CI Integration
- Added GitHub workflow `.github/workflows/automated-a11y.yml`:
  - Runs accessibility tests on pull requests and can be manually triggered
  - Stores test results as artifacts

### 4. Added Scripts
- In root `package.json`:
  - Added `"test:automated-a11y": "yarn workspace @adyen/adyen-web-playwright test:automated-a11y:headless"`

- In `/packages/e2e-playwright/package.json`:
  - Added scripts:
    ```json
    "test:automated-a11y:headless": "npx playwright test --config=automated-a11y.config.ts",
    "test:automated-a11y:headed": "npx playwright test --config=automated-a11y.config.ts --headed",
    "test:automated-a11y:ui-mode": "npx playwright test --config=automated-a11y.config.ts --ui"
    ```
  - Added dependency: `"@axe-core/playwright": "^4.10.1"`

## Benefits

- **Comprehensive Testing**: Automatically tests all components against WCAG 2.0 and 2.1 standards (A & AA levels)
- **Early Detection**: Catches accessibility issues during development/CI rather than post-deployment
- **Maintainable**: Allows for documenting known issues while working toward resolution
- **Scalable**: As new components are added to Storybook, they're automatically tested

## Usage

Run automated a11y tests with:

```bash
# Headless mode (CI)
yarn test:automated-a11y

# With browser UI
yarn workspace @adyen/adyen-web-playwright test:automated-a11y:headed

# With Playwright UI mode for debugging
yarn workspace @adyen/adyen-web-playwright test:automated-a11y:ui-mode
```


.github/workflows/automated-a11y.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Run Automated Accessibility Tests
+      run: yarn test:automated-a11y -- --project=${{ matrix.runner-browser }}
+    
+    - name: Archive test result artifacts
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
+        name: a11y-report-${{matrix.api-version}}-${{matrix.runner-browser}}-${{ github.sha }}
+        path: packages/e2e-playwright/playwright-report
+        retention-days: 5
```
ribeiroguilherme, line 53: Does it make sense to always generate the artifacts? 
Can we see highlighted a11y errors when the test fails?

package.json
```diff
@@ -41,5 +42,6 @@
         "tar-fs": "^3.0.9",
         "**/estimo/nanoid": "^5.1.5",
         "cheerio": "1.0.0-rc.12"
-    }
+    },
+    "packageManager": "yarn@1.22.21+sha1.1959a18351b811cdeedbd484a8f86c3cc3bbaf72"
```
ribeiroguilherme: This should be removed

packages/e2e-playwright/automated-a11y.config.ts
```diff
@@ -0,0 +1,83 @@
...
+            }
+        },
+
+        {
+            name: 'webkit',
+            use: {
+                ...devices['Desktop Safari']
+            }
+        }
+    ],
```
ribeiroguilherme, line 69: Does it have any effect? I don't see tests running for firefox nor webkit.

In the .yml file we are setting only "chromium" , and checking the pipeline that was executed I can see only "chromium" there.

packages/lib/.storybook/preview-head.html
```diff
@@ -0,0 +1,3 @@
+<!-- .storybook/preview-head.html -->
+<!-- Add meta tag for Permissions-Policy -->
+<meta http-equiv="Permissions-Policy" content="camera=(), geolocation=(), microphone=()" />
```
ribeiroguilherme: Out of curiosity - why do we need a `preview-head.html` file?

packages/e2e-playwright/tests/e2e/card/card.spec.ts
```diff
@@ -5,9 +5,13 @@ import { URL_MAP } from '../../../fixtures/URL_MAP';
 test.describe('Card - Standard flow', () => {
     test('#1 Should fill in card fields and complete the payment', async ({ card }) => {
         await card.goto(URL_MAP.card);
+
         await card.typeCardNumber(REGULAR_TEST_CARD);
         await card.typeCvc(TEST_CVC_VALUE);
         await card.typeExpiryDate(TEST_DATE_VALUE);
+
+        expect(await card.getA11yErrors()).toHaveLength(0);
```
ribeiroguilherme, line 13: Why do we need the a11y assert here and in the `dropin.sessions.spec.ts` file? Aren't these checks covered by the `automated.spec.ts` ?

packages/e2e-playwright/tests/automated-a11y/automated.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+// These are the list of violations that are know for each expose component
+// The key is the story id, this can be found in the test that is failing or in storybook
+// The values in the violation array are the ID of the printed violation
+const KNOWN_A11Y_VIOLATIONS = {
+    'components-payto--default': ['aria-valid-attr-value'], 
+    'dropin-default--style-customization': ['color-contrast'], // Demo page
+    'internals-toggle--toggle-only': ['label'],
+    'issuerlist-onlinebankingcz--default': ['link-in-text-block'], // TODO
+    'issuerlist-onlinebankingpl--default': ['link-in-text-block'], // TODO
+    'issuerlist-onlinebankingsk--default': ['link-in-text-block'], // TODO
```
ribeiroguilherme: What does this TODO mean? It is not clear what needs to be done (e.g. do we need to fix the component? do we need to fix the aria rules here? do we need to fix the story itself?)

packages/e2e-playwright/models/automated.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    constructor(page: Page) {
+        super(page);
+
+        this.payButton = this.page.locator('#component-root');
+    }
+
+    async goto(url: string) {
+        await this.page.goto(url, { waitUntil: 'domcontentloaded', timeout: 15000 });
+        await this.isComponentVisible();
```
ribeiroguilherme: This method is already available in the parent class. I believe you can add enhance it there with the change that you have here:
```
await this.page.goto(url, { waitUntil: 'domcontentloaded', timeout: 15000 });
```
Seems logical to me. And then this method in the `automated.ts` can go way

packages/e2e-playwright/models/automated.ts
```diff
@@ -0,0 +1,39 @@
...
+        const storyUrl = this.page.url();
+        // Check if is 'internal' component story
+        return storyUrl && storyUrl.includes('internals-');
+    }
+
+    protected a11yComponentSelector() {
+        if(this.isInternalStory()){
+            return ('#storybook-root');
+        }
+        return '#component-root';
```
ribeiroguilherme, line 23: Can't we always use `#storybook-root` ? It seems to be always available as the root

packages/e2e-playwright/models/automated.ts
```diff
@@ -0,0 +1,39 @@
+import type { Page, Locator } from '@playwright/test';
+import { Base } from './base';
+
+export class Automated extends Base {
+    readonly payButton: Locator;
+
+    constructor(page: Page) {
+        super(page);
+
+        this.payButton = this.page.locator('#component-root');
```
ribeiroguilherme: The 'payButton' is getting assigned to it the main Component div. This seems a bit off 🤔

packages/e2e-playwright/fixtures/base-fixture.ts
```diff
@@ -1,7 +1,12 @@
 import { test as base, expect } from '@playwright/test';
 import http from 'http';
+import AxeBuilder from '@axe-core/playwright';
 
-const test = base.extend({
+type AxeFixture = {
+    makeAxeBuilder: () => AxeBuilder;
+};
+
+const test = base.extend<AxeFixture>({
```
ribeiroguilherme: Does this actually do something? I don't see this `makeAxeBuilder` being used anywhere

packages/e2e-playwright/models/automated.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    async isComponentVisible() {
+        if(this.isInternalStory()){
+            // Becasue these don't have paybutton wait 3s - it was trial and error
+            // another way of doing this is welcome
+            return this.page.waitForTimeout(3000);
+        }
+        await this.payButton.waitFor({ state: 'visible' });
+    }
+}
```
ribeiroguilherme, line 34: I believe that, if my comments added to this file are all valid, we might not need to create a class that extends `base` . We perhaps could just use the `base` model directly

.github/workflows/automated-a11y.yml
m1aw: This is just the same logic we have for E2E playwright.
We can change to just generate on fail, but I guess we can just be consistent.

packages/e2e-playwright/fixtures/base-fixture.ts
m1aw: Indeed, this was the first attempt. It worked but I like the one in the model more. I will clean up.

packages/e2e-playwright/automated-a11y.config.ts
m1aw: Initially I made it run on all browsers, and so I just made it target a specific browser in the workflow file. 
This is easier in my view because if you need to add them in the future you can edit on one place, and also if you want to run locally on different projects you can, makes it easier for testing.
I can also remove, but I don't think is a big deal.

packages/e2e-playwright/tests/e2e/card/card.spec.ts
m1aw: Yeah, this is an example if we want to do manually.
For example, if you want to check a11y **after** the test runs, for example in an error state with all input fields filled.

This is just an example, then it's up to the developer writing the test to do so.

packages/e2e-playwright/tests/automated-a11y/automated.spec.ts
m1aw: Good point, I will add more information, in this case it's because we need to fix our link color. Which has already been flagged and we are working on with design.

packages/e2e-playwright/models/automated.ts
m1aw: Need to change variable name.

packages/e2e-playwright/models/automated.ts
m1aw: Yeah, I wasn't sure if I wanted to make the change or not on the parent.

packages/e2e-playwright/models/automated.ts
m1aw: There are other stories, for example Giftcard, where there's some UI elements made by us, at the same level of `#component-root`, so I felt like this is better.

packages/lib/.storybook/preview-head.html
m1aw: Can be removed now, was for some tests.

packages/e2e-playwright/models/automated.ts
m1aw: I can make this change, the objective was however to have the flexibility of having completely separate logic.

I prefer it like this, if we need to extend of change anything for the a11y tools we can do so here, and not impact, or have to test changes to the main tests.

Thoughts?

.github/workflows/automated-a11y.yml
ribeiroguilherme: Ah ok. I thought we were generating only if there are errors

packages/e2e-playwright/models/automated.ts
ribeiroguilherme: I understand. It is just that these classes do not seem to deviate much - it looks like the `base` can have everything you need in there and there is no need (at least for now) to extend it.

But I will leave it up to you to decide, it is not a big deal.

ribeiroguilherme (approved):

PR #3299, opened 2025-05-07
Riverty - Converted component to Redirect

ribeiroguilherme:
## Summary
- Adjusted `Riverty` to be a full redirect. (It needs to have its own Component since it must pass the `subtype: 'redirect'` in the state data)
- Adjusted Storybook code to accept passing the shopper details when creating the session OR when making the /payments call for the advanced flow. Passing shopper details is required for certain payment methods to work (e.g. Riverty)
- 

## Tested scenarios
- Tested Riverty component on Storybook, getting redirected to Riverty page and then redirect back to Storybook
- Adjusted unit tests


sponglord:
Don't you also need to update `components-name-map.ts`, `components-map.ts`, and `components/index.ts` to remove the Riverty entries?

sponglord:
Will you also move the entries for Riverty in `components-map.ts`, `components-name-map.ts`, `components/index.ts`, & `components/tx-variants.ts` to the "Redirect" sections?

packages/e2e-playwright/tests/a11y/openInvoices/riverty.spec.ts
sponglord, line 1: You've deleted this test rather than make one that points to another OpenInvoice component.
So we now have no OpenInvoice e2e tests?

packages/e2e-playwright/fixtures/openInvoices.fixture.ts
sponglord, line 1: See comment below about us now having no OpenInvoice e2e tests

packages/e2e-playwright/tests/a11y/openInvoices/riverty.spec.ts
ribeiroguilherme: We have integration tests for `AfterPay` ( `AfterPay.test.ts`) which cover the OpenInvoice component as a whole (it covers input fields, error messages, onSubmit being called). 

I believe we don't need to add a OpenInvoice E2E for now, as long as we are covering on unit test/integration test level.

PR #3300, opened 2025-05-07
Feature/make analytics more declarative

sponglord:
## Summary
Creating analytics events is more declarative: 
- you can see what kind of event ('`info', 'log', 'error'`) you are creating at the time you make it
- all analytics events are created through creating a `new` instance of a specific class (`AnalyticsInfoEvent`,  `AnalyticsLogEvent`, `AnalyticsErrorEvent`), making it easy to see _all_ the places in the codebase that analytics are being added
- the `createAnalyticsObject` util has been removed: this util was meant to select which properties were added to an analytics event - but had the side effect of obscuring how an analytics event was formed. Now you fully form your analytics event when you create it
- `analyticsPreProcessor` has been removed, it is no longer necessary; and also tended to obscure how an analytics event was formed
- the new classes are typed so warn you, in the IDE, if you are adding properties that, for example, an error event doesn't support (and which would lead to the api call failing)
- `configData` and other PM related properties e.g. `isExpress` are added by the relevant PM (i.e the class that extends `UIElement`) rather than being done in `analyticsPreProcessor`

## Tested scenarios
Unit tests have been updated
All unit tests pass
Manually checked that all analytics events are correctly created and successfully passed to the api

### Notes for reviewers:  
This looks like a big PR, but that's mainly because the signature to create a new analytics event has changed, so it touches everywhere we currently do that - but these are minor changes.

The files that really reflect the change are:
- `core/Analytics/AnalyticsEvent.ts`
- `core/Analytics/AnalyticsInfoEvent.ts`
- `core/Analytics/AnalyticsLogEvent.ts`
- `core/Analytics/AnalyticsErrorEvent.ts`
- `core/Analytics/Analytics.tsx`
- `components/internal/UIElement/UIElement.tsx` (primarily its `submitAnalytics` function, line 175)
- `components/Card/Card.tsx` (its `submitAnalytics` function, line 200)
- `Card.tsx` also contains several examples of the use of `new AnalyticsInfoEvent()`


packages/lib/src/core/Analytics/types.ts
```diff
@@ -87,41 +107,50 @@ export interface AnalyticsObject {
...
+    target?: string; // is this ever used?
 };
 
-export type AnalyticsConfig = {
-    analyticsContext?: string;
-    clientKey?: string;
-    locale?: string;
-    amount?: PaymentAmount;
-    loadingContext?: string;
+export type NewAnalyticsEventObjectError = {
```
ribeiroguilherme, line 135: I see there are couple of types prefixed by `New` -  what is the intention of using the `New`?  Are they just normal types and you are using `New` because you are refactoring/renaming things?

In general, `New` is a special keyword to create objects and it is not used to define types. It seems a bit off to me

packages/lib/src/core/Analytics/utils.ts
```diff
@@ -1,62 +1,94 @@
...
+    'errorType',
+    'expressPage',
+    'isExpress',
+    'isStoredPaymentMethod',
+    'issuer',
+    'validationErrorCode',
+    'validationErrorMessage'
+];
+
+export const createNewAnalyticsEvent = (aObj: CreateNewAnalyticsEventObject): EnhancedAnalyticsObject => {
```
ribeiroguilherme, line 44: It is a great start decoupling all the pre-processing from the analytics module and to make the creation of events more declarative.

Given that, I believe there is more room from improvement when creating analytic events. We can make it more declarative and more explicit. I don't believe that creating a factory function to create all sort of events is the best way to go. Its API will just increase over time; it is also hard to track which kind of events are used in the codebase; it is also not strongly typed, and we have to keep adding custom logic per event category to perform certain checks..

I believe that a more Object Oriented approach would be better here. When I say 'object oriented', that would mean that we could have:
```js
abstract class AnalyticEvent {
  public string type;
  public string target;
// etc.. whatever is common to all analytics here
```

And then we would have more granular events:

```js
class AnalyticInfoEvent extends AnalyticEvent {
 // Info related properties here
```

And then from within Components or any other module, we could just instantiate an object related to that event and dispatch it:

```js
const event = new AnalyticInfoEvent({ target: 'pay_button' });
this.submitAnalytics(event);
```

There are great benefits in doing it like this:
- stronger type check and strict interfaces when creating analytic events
- easy to trace which events are being used where
- each analytic event class can encapsulate its own logic to validate certain things related to its category
- no need to do type casting (e.g. usage of `as EnhancedAnalyticsObject`) - The Analytics module just needs to care that we are passing it an object which is an instance of `AnalyticEvent` . It does not need to care which category that is 

Following this approach will improve maintainability and will help the team later on to work on the Analytics improvements/refactoring.

packages/lib/src/core/Analytics/types.ts
sponglord: It was just due to the refactoring - to easily distinguish the "new" stuff I was creating from what was already there.
I'll replace it all once the final shape of the code is decided

packages/lib/src/core/Analytics/AnalyticsEvent.ts
```diff
@@ -0,0 +1,13 @@
+import uuid from '../../utils/uuid';
+import { getUTCTimestamp } from './utils';
+
+export abstract class AnalyticsEvent {
+    public timestamp: string;
+    public id: string;
+    public component: string;
+
```
ribeiroguilherme, line 8: I believe `target` can also be added here in the abstract object. It is available for all event types

packages/lib/src/core/Analytics/AnalyticsInfoEvent.ts
```diff
@@ -0,0 +1,43 @@
+import { AnalyticsEvent } from './AnalyticsEvent';
+import { ANALYTICS_VALIDATION_ERROR_STR } from './constants';
+import { mapErrorCodesForAnalytics } from './utils';
+import { AnalyticsInfoEventObject } from './types';
+
+export class AnalyticsInfoEvent extends AnalyticsEvent {
+    public type: string;
+    public target: string;
+    public component: string;
```
ribeiroguilherme, line 9: Fields that are defined on the parent object do not need to be re-defined on the extending classes. You can get rid of `component` field here and in the other events

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -84,28 +88,34 @@ const Analytics = ({ locale, clientKey, analytics, amount, analyticsContext, bun
...
-        sendAnalytics: null
-    };
+            if (analyticsObj instanceof AnalyticsErrorEvent) {
+                event = ANALYTICS_EVENT.error;
+            }
 
-    anlModule.sendAnalytics = props.enabled === true ? analyticsPreProcessor(anlModule) : () => {};
+            if (!event) {
+                throw new AdyenCheckoutError(SDK_ERROR, 'You are trying to create an analytics event without an event type');
+            }
```
ribeiroguilherme, line 112: I believe you should be able to retrieve the event category from the object, without using any  `if` statements.

You can define on the parent `AnalyticsEvent` an abstract method that the extending classes must implement to return the category:

```js
export abstract class AnalyticsEvent {
  // ...

  public abstract getEventCategory(): string;

}
```

Then:
```js
export class AnalyticsErrorEvent extends AnalyticsEvent {
   // ...

  public getEventCategory(): string {
    return AnalyticsEventCategory.Error;
  }

}
```

Then in this part of the code, you can:

```js
sendAnalytics: (event: AnalyticsEvent): boolean => {
  if (!props.enabled) return false;
  if (event instanceof AnalyticsEvent === false) throw new AdyenCheckoutError(SDK_ERROR, 'You are trying to create an invalid analytic event');

  const eventCategory = event.getCategory();

  addAnalyticsEvent(eventCategory, event);
}
```

packages/lib/src/core/Analytics/types.ts
```diff
@@ -87,42 +104,38 @@ export interface AnalyticsObject {
...
+    target?: string;
+    issuer?: string;
+    isExpress?: boolean;
+    expressPage?: string;
+    isStoredPaymentMethod?: boolean;
+    brand?: string;
+    validationErrorCode?: string;
+    validationErrorMessage?: string;
+    configData?: Record<string, string | boolean>;
+    component?: string;
```
ribeiroguilherme, line 118: I suggest to move the event types declaration into their own files. I looked up in the code and it seems that we do not need to expose it as we are doing it here. We also discussed in the past to put the types closer to where they are being used if possible (except for types that must be exposed to merchants e.g. `CardConfiguration`)

We can move each of these type declarations to their respective files . For example, for the log event category in the `AnalyticsLogEvent` file, we can:

```js
interface LogEventProps {
  type: string;
  // ...
}

export class AnalyticsLogEvent extends AnalyticsEvent {
  constructor(props: LogEventProps) { ... }
}
```

And ideally we should not mark all fields as optional. We must enforce at least the required fields to be passed in (e.g. `type` is a required field)

ribeiroguilherme (approved):

packages/lib/src/core/Analytics/AnalyticsEvent.ts
sponglord: No, it's not allowed on an error event

packages/lib/src/core/Analytics/Analytics.ts
sponglord: Good one!

Issue #2351, opened 2023-10-11
Allow Apple Pay payments from an iframe (Payment Not Completed)


jarekcieslak:
**Is your feature request related to a problem? Please describe.**

Currently, when the Adyen web drop-in component is placed on a page embedded in an iframe, the Apple Pay button is visible. However, upon tapping it, the Apple Pay modal window is immediately dismissed with a message stating "Payment not completed."

This issue arises because the Adyen web drop-in component internally creates an Apple payment session using the URL of the iframe, rather than that of the parent page. Refer to [this link](https://github.com/Adyen/adyen-web/blob/789b285b86e3eee43c69d6feb3da0f12960840a9/packages/lib/src/components/ApplePay/ApplePay.tsx#L98). The Apple Pay modal window opens in the context of the parent page, leading to a domain mismatch.

This poses a problem for me as my web app, which provides a checkout experience based on the web drop-in component, is rendered inside an iframe. Naturally, I'd like to utilize Apple Pay within this iframe.

For proof of concept, I forked the adyen-web library and modified the URL used to create the Apple Pay session to match that of the parent page, which resolved the issue. Consequently, I was able to process Apple Pay payments from within the iframe.

**Describe the solution you'd like**
I would like the ability to override the default behavior. When passing the Checkout configuration object, I want to specify the URL that should be used to create the Apple Pay session, enabling me to use Apple Pay within the iframe.

**Describe alternatives you've considered**
I've thought about creating an Apple Pay session using the onValidateMerchant hook. However, this approach necessitates the creation of a custom backend endpoint and complicates the entire integration process.

**Additional context**
- I'm using @adyen/adyen-web@5.51.0.
- I've been in contact with Sebastian Lerch at [sebastian.lerch@adyen.com](mailto:sebastian.lerch@adyen.com) regarding this issue, and he recommended submitting a feature request here.
- Screenshot of the issue: 
![payment not completed](https://github.com/Adyen/adyen-web/assets/3534325/82d77709-4e7f-4903-914f-8beaf31cff05)

ribeiroguilherme:
Hey @jarekcieslak ,

Thanks for reaching out and for your proposal. One of the considerations when working ApplePay is that it doesn't work well with iframes indeed, and that was by design. To overcome this problem the iframe origin must match with the parent origin, which is not always the case.

[...]

PR #3340, opened 2025-05-31
feat: allow domainName to be overrided for iframe apply pay session

idl3:
## Summary
- iframe ApplePay sessions do not work as the parent frame URL is not being sent down to the ApplePaySession instantiation. It uses `window.location.hostname` so it will always be the iframe URL instead of the parent URL.

Some basic principles on why this approach:
- Cross-origin domains is a choice, not a norm.
- window.location.hostname works for most cases if not cross-origin
- In the case of cross-origin domains, the parent domain will be passed to the child by whatever mechanism desired, it is not the responsibility of the adyen-web package. It only accepts the override in the conditions that the app has decided it to be for security reasons.
- It is still subject to the apple pay domains that are added.

## Tested scenarios
- Having a page load an iframe which contains adyen-web with ApplePay loaded. Payment method has both domains white listed but Apple Pay only works when loading the iframe URL standalone, but when loaded from the parent site, it shows an error and closes the payment sheet.

## Does it work?
- Yes, I have built the package and used it locally in my own application and tested that iframe apple pay works and payment was successful.

### DIFF
--- a/packages/lib/src/components/ApplePay/types.ts
+++ b/packages/lib/src/components/ApplePay/types.ts
@@ -231,7 +231,7 @@ export interface ApplePayConfiguration extends UIElementProps {
     /**
      * Used to override the domain name for the Apple Pay button.
      */
-    domainNameOverride?: string;
+    domainName?: string;
 }
 
 export interface ApplePayElementData {
--- a/packages/lib/src/components/ApplePay/types.ts
+++ b/packages/lib/src/components/ApplePay/types.ts
@@ -227,6 +227,11 @@ export interface ApplePayConfiguration extends UIElementProps {
      * Used to tweak the text of the button types that contain text ('Continue with', 'Book with', etc)
      */
     buttonLocale?: string;
+
+    /**
+     * Used to override the domain name for the Apple Pay button.
+     */
+    domainNameOverride?: string;
 }
 
 export interface ApplePayElementData {
--- a/.changeset/curvy-bats-worry.md
+++ b/.changeset/curvy-bats-worry.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+Adds ApplePay domain name override for iframe usage
--- a/packages/lib/src/components/ApplePay/ApplePay.tsx
+++ b/packages/lib/src/components/ApplePay/ApplePay.tsx
@@ -342,13 +342,13 @@ class ApplePayElement extends UIElement<ApplePayConfiguration> {
 
     private async validateMerchant(resolve, reject) {
         const { hostname: domainName } = window.location;
-        const { clientKey, configuration, loadingContext, initiative } = this.props;
+        const { clientKey, configuration, loadingContext, initiative, domainNameOverride } = this.props;
         const { merchantName, merchantId } = configuration;
         const path = `v1/applePay/sessions?clientKey=${clientKey}`;
         const options = { loadingContext, path };
         const request: ApplePaySessionRequest = {
             displayName: merchantName,
-            domainName,
+            domainName: domainNameOverride || domainName,
             initiative,
             merchantIdentifier: merchantId
         };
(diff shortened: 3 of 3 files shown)


ribeiroguilherme:
@idl3 thanks for reaching out. 

In your case, is the parent URL in a complete different domain than the Checkout URL? How did you overcome this issue so far? Is that an issue with the Safari browser or with any browser?

I believe this is related to https://github.com/Adyen/adyen-web/issues/2351

frebos88:
> @idl3 thanks for reaching out.
> 
> In your case, is the parent URL in a complete different domain than the Checkout URL? How did you overcome this issue so far? Is that an issue with the Safari browser or with any browser?
> 
> I believe this is related to #2351

We have such a case. The main limitation is that you do not allow to specify the initiating domain, which is what this PR aims to solve. Allowing iframes to initiate payments has already been resolved (see webkit PR as an example).
https://github.com/WebKit/WebKit/pull/11485

A simple example could be the following:

- Company runs their ecommerce on www.example.com
- Their inhouse payments run on pay.example.com through an iframe with <iframe allow="payment">
- The Apple allowed domain to process payments is www.example.com as it should be
- **Adyen** uses **pay**.example.com to initiate the payment by using window.location.hostname, **it fails**

The PR resolves this by allowing pay.example.com to send www.example.com as the payment domain in the Adyen configuration. The same exact thing would apply for completely different domains, where the company may be outsourcing the payments parts.

idl3:
> @idl3 thanks for reaching out.
> 
> In your case, is the parent URL in a complete different domain than the Checkout URL? How did you overcome this issue so far? Is that an issue with the Safari browser or with any browser?
> 
> I believe this is related to #2351

Yeah, as @frebos88 mentioned, we are embedding `pay.example.com` into `anotherexample.com` through an iframe, and as such we need to allow `anotherexample.com` to be the main domain that's initiating apple pay, not `pay.example.com`. 

By exposing an override will help us to circumvent this issue since the code will now only use `window.location.host` to determine the apple pay session domain.

packages/lib/src/components/ApplePay/types.ts
```diff
@@ -227,6 +227,11 @@ export interface ApplePayConfiguration extends UIElementProps {
      * Used to tweak the text of the button types that contain text ('Continue with', 'Book with', etc)
      */
     buttonLocale?: string;
+
+    /**
+     * Used to override the domain name for the Apple Pay button.
+     */
+    domainNameOverride?: string;
```
ribeiroguilherme: Can you rename this property to `domainName` ? We don't need to flag it as `Override`

ribeiroguilherme (commented):
Can you also push a changeset? 

You have to run `yarn changeset` on the root level and flag it as 'minor' change, and add the summary of this change. This will go to our release notes

packages/lib/src/components/ApplePay/types.ts
ribeiroguilherme: Perhaps enhance its description sharing that it is useful/required for the scenario where ApplePay is placed within an iframe

idl3:
Gotcha will make the changes 👌

frebos88:
> Gotcha will make the changes 👌

@idl3 Are you still looking in to adressing this?

idl3:
> > Gotcha will make the changes 👌
> 
> @idl3 Are you still looking in to adressing this?

Have just made the changes and pushed up the changeset as well, rebased to the latest main and force pushed just in case.

idl3:
Not sure why tests are failing tbh, might need someone to take a look at it 🙏🏼

ribeiroguilherme:
Hi @idl3  - we would like to proceed and merge the PR, but unfortunately we are unable to run the required pipelines due to the actions that are using secrets.

We don't have a fix for it for now, but what I can do from our side is cherry-pick the changes and create another PR which the pipelines will be executed. Is it ok to proceed like this?

idl3:
> Hi @idl3  - we would like to proceed and merge the PR, but unfortunately we are unable to run the required pipelines due to the actions that are using secrets.
> 
> 
> 
> We don't have a fix for it for now, but what I can do from our side is cherry-pick the changes and create another PR which the pipelines will be executed. Is it ok to proceed like this?

Sure thing, no problem for me 👌

ribeiroguilherme:
I created a PR with these changes, and I expect it to go live next week 👍  I will close this one then. Thanks!

PR #3342, opened 2025-06-02
feat: expose `core`

Ymasri93:
## Summary
**Context**

With redirect-based payment methods like iDEAL, shoppers may return back to the site using a different browser after completing the payment on an external site or application. This behavior makes it unreliable to depend on locally stored `sessionData`, as it may not be available upon return.

Currently, when initializing `AdyenCheckout` after redirection, the sdk always triggers a setup call. This call fails if `sessionData` is missing (e.g., due to localStorage missing the `sessionData` in the new browser context).

To address this, we’re now exposing the `Core` class. This provides more flexibility for PBL/merchants: in redirection to a different browser scenarios, instead of instantiating a new `AdyenCheckout`, they can directly instantiate `Core`. This approach bypasses the setup call, avoiding failure due to missing `sessionData`.

### DIFF
--- a/.changeset/happy-numbers-bow.md
+++ b/.changeset/happy-numbers-bow.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+Export `Core` class.
--- a/packages/lib/src/core/AdyenCheckout.ts
+++ b/packages/lib/src/core/AdyenCheckout.ts
@@ -19,4 +19,4 @@ AdyenCheckout.setBundleType = (type: string) => {
     Core.setBundleType(type);
 };
 
-export { AdyenCheckout };
+export { AdyenCheckout, Core };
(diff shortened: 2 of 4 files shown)


packages/lib/src/core/core.defaultProps.ts
```diff
@@ -2,7 +2,8 @@ import type { CoreConfiguration } from './types';
 
 const defaultProps: CoreConfiguration = {
     exposeLibraryMetadata: true,
-    showPayButton: true
+    showPayButton: true,
+    _skipSessionSetup: false
```
Ymasri93: Probably not needed

sponglord:
Can you expand the description to explain what issue you're fixing here, and how exposing the Core can be used to fix it?

longyulongyu:
> Can you expand the description to explain what issue you're fixing here, and how exposing the Core can be used to fix it?

Updated in the summary.

PR #3356, opened 2025-06-13
Feature: Add srpanel for gift card

m1aw:
## Summary

Adds error panel to Giftcard.

There's a few tricks however, the `useSRPanelForGiftcardErrors` hook is passed to render function because Giftcard component is a Class component. 

Also we had to use `mapErrorsToValidationRuleResult` to conver SF errors into `useForm` like errors. 

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
+++ b/packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
@@ -1,0 +1,129 @@
+import { useEffect, useState } from 'preact/hooks';
+import { SetSRMessagesReturnObject } from '../../../core/Errors/types';
+import useSRPanelContext from '../../../core/Errors/useSRPanelContext';
+import { usePrevious } from '../../../utils/hookUtils';
+import { ERROR_ACTION_BLUR_SCENARIO, ERROR_ACTION_FOCUS_FIELD } from '../../../core/Errors/constants';
+import { getArrayDifferences } from '../../../utils/arrayUtils';
+import { ValidationRuleResult } from '../../../utils/Validator/ValidationRuleResult';
+import { setFocusOnField } from '../../../utils/setFocus';
+import { SetSRMessagesReturnFn } from '../../../core/Errors/SRPanelProvider';
+
+/**
+ * Interface for gift card field errors, matching the encrypted field types
+ */
+export interface GiftcardStateErrors {
+    encryptedCardNumber: ValidationRuleResult;
+    encryptedSecurityCode: ValidationRuleResult;
+}
+
+/**
+ * Interface for transformed error objects returned from mapErrorsToValidationRuleResult
+ */
+export interface TransformedError {
+    isValid: boolean;
+    errorMessage: string;
+    errorI18n: string;
+    error: string;
+    rootNode: HTMLElement;
+}
+
+/**
+ * Interface for errors object with transformed error entries
+ */
+export interface TransformedErrorsObj {
+    [key: string]: TransformedError | null;
+}
+
+/**
+ * Props interface for the error handling hook
+ */
+interface UseSRPanelForGiftcardErrorsProps {
+    errors: TransformedErrorsObj;
+    isValidating: { current: boolean } | boolean;
+}
+
+/**
+ * Interface for sorted error objects used for comparison
+ */
+interface SortedErrorObject {
+    field: string;
+    errorMessage: string;
+}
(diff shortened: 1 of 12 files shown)


packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
```diff
@@ -0,0 +1,129 @@
...
+                        setSRMessagesFromStrings(latestSRError);
+                    } else {
+                        // Clear SR panel if there are no more errors
+                        clearSRPanel();
+                    }
+                    break;
+                }
+            }
+        } catch (_) {
+            // Fail silently - we don't want to break the component if SRPanel fails
```
m1aw, line 114: Need to clean this.

packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
```diff
@@ -0,0 +1,129 @@
...
+                layout
+            });
+
+            // Store the sorted list of errors for comparison
+            const currentErrorsSortedByLayout = srPanelResp?.currentErrorsSortedByLayout;
+            setSortedErrorList(currentErrorsSortedByLayout);
+
+            // Handle different error scenarios based on the action type
+            switch (srPanelResp?.action) {
+                case ERROR_ACTION_FOCUS_FIELD:
```
m1aw, line 83: I'm wondering if we ever get this types of errors on Giftcard.

sponglord: I'm not sure if I understand your comment...
This `case` is the primary piece of code that, in the event of the shopper pressing pay, moves focus to the relevant field (the first field in error)

And, if I test locally, I don't see the focus moving

packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
```diff
@@ -0,0 +1,129 @@
...
+import { SetSRMessagesReturnObject } from '../../../core/Errors/types';
+import useSRPanelContext from '../../../core/Errors/useSRPanelContext';
+import { usePrevious } from '../../../utils/hookUtils';
+import { ERROR_ACTION_BLUR_SCENARIO, ERROR_ACTION_FOCUS_FIELD } from '../../../core/Errors/constants';
+import { getArrayDifferences } from '../../../utils/arrayUtils';
+import { ValidationRuleResult } from '../../../utils/Validator/ValidationRuleResult';
+import { setFocusOnField } from '../../../utils/setFocus';
+import { SetSRMessagesReturnFn } from '../../../core/Errors/SRPanelProvider';
+
+/**
```
sponglord: Is this interface used anywhere? (GiftcardStateErrors)
It's greyed out in my IDE and I can find it mentioned outside of this component

packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
```diff
@@ -0,0 +1,129 @@
...
+    // Track sorted list of errors for comparison with previous state
+    const [sortedErrorList, setSortedErrorList] = useState<SortedErrorObject[]>(null);
+    // Track previous error list for detecting changes
+    const previousSortedErrors = usePrevious(sortedErrorList);
+
+    // Get SRPanel context for managing screen reader messages
+    const { setSRMessagesFromObjects, setSRMessagesFromStrings, clearSRPanel, shouldMoveFocusSR } = useSRPanelContext();
+
+    // Fixed layout array defining the order of fields for error sorting
+    const layout = ['encryptedCardNumber', 'encryptedSecurityCode'];
```
sponglord: Remainder: missing `encryptedExpiryDate`

packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
m1aw: Yeah, I think I didn't understand the feature. I just realised what I was doing wrong and it's fixed.

packages/lib/src/components/Giftcard/components/useSRPanelForGiftcardErrors.ts
m1aw: Removed.

sponglord (approved):

PR #3359, opened 2025-06-16
[wip] Cosdk 457 bank transfer UI improvements

longyulongyu:
## Tested scenarios

**Fixed issue**:


packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -199,7 +199,13 @@ class DropinElement extends UIElement<DropinConfiguration> implements IDropin {
             if (isPMHeader) {
                 return;
             }
-            super.handleKeyPress(e);
+            // Call handleKeyPress on activePaymentMethod or action if available
+            const paymentMethodHandleKeyPress = (this.componentFromAction as any)?.handleKeyPress || this.activePaymentMethod?.handleKeyPress;
```
longyulongyu, line 203: Let me know if there is a better way to do it. Ideally we just need to call `this.activePaymentMethod?.handleKeyPress;`, but the props from `this.activePaymentMethod` are not merged with the action, so when calling the `this.activePaymentMethod?.handleKeyPress;` we couldn't get the latest props. Maybe we can do something in the dropin `handleAction`.

packages/lib/src/components/internal/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,102 @@
...
+                'adyen-checkout-tooltip--hidden': !props.visible || !isAnchorVisible,
+                [`adyen-checkout-tooltip--${position}`]: true
+            })}
+            ref={tooltipRef}
+            style={style}
+        >
+            {props.text}
+            <div className={`adyen-checkout-tooltip-arrow adyen-checkout-tooltip-arrow--${position}`} />
+        </div>,
+        document.body
```
m1aw: I'm concern with this. It might need to get checked by LevelAccess.

It has 2 minor quirks: we don't announce the "Copied!", probably easy to fix by giving `aria-live` to the tooltip; because the tooltip is the description of the button, it reads label > description > role, which is "Copy, Copy, button" so it duplicates information. 

I think both points are minor and we can iterate on them.

packages/lib/storybook/stories/components/BankTransfer.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { BankTransferConfiguration } from '../../../src/components/BankTransfer/types';
+import BankTransfer from '../../../src/components/BankTransfer';
+
+type BankTransferStory = StoryObj<PaymentMethodStoryProps<BankTransferConfiguration>>;
+
+const meta: Meta<PaymentMethodStoryProps<BankTransferConfiguration>> = {
+    title: 'Components/Bank Transfer'
+};
+
+export const Default: BankTransferStory = {
```
m1aw, line 14: Might be worth creating a story with just the result page so we can run the automated a11y testing on it.

m1aw: Also this might be very relevant because merchants might want to save the action and then display this after the first time is rendered.

packages/lib/src/components/internal/Tooltip/SingletonTooltipProvider.tsx
```diff
@@ -0,0 +1,75 @@
...
+    showTooltip: (state?: TooltipProps) => void;
+    hideTooltip: () => void;
+};
+
+const TooltipContext = createContext<TooltipContextValue | null>(null);
+
+const SingletonTooltipProvider = ({ children }: { children?: ComponentChildren }) => {
+    const [tooltipProps, setTooltipProps] = useState<TooltipProps | null>(null);
+    const tooltipId = useRef(TooltipController.tooltipId);
+    const isPrimaryInstanceRef = useRef(false);
```
m1aw, line 26: I don't understand why this is not just state but a ref. Because I see we have a clever mechanism to keep "updated" but maybe add a comment on why this is needed?

m1aw (approved):

packages/lib/src/components/internal/Tooltip/Tooltip.tsx
longyulongyu: I added `aria-live` to the tooltip div, but I couldn't reliably let the voice over announce the text, shouldn't I use the sr panel hook `useA11yReporter` for it?:
```html
<div
  id={props.id}
  role="tooltip"
  aria-live={'polite'}
</div>
```

m1aw: Yeah, I have tested and also it doesn't seem to be consistent for me. I think we can iterate on this.

longyulongyu: Ok, I did some search, chatgpt suggests me to do following:
- Remove `aria-describedby` from the button, aria-describedby causes delayed or stale announcements when focus changes. This is also the case happened for me, the old status message was announced when i focus to the next copy button
- Add `aria-live` as you suggested
- Change the tooltip css to be visually hidden not `display:none`. Because even if you later change `display: block` and update the content, some screen readers won't announce the change reliably because they missed the content being added while hidden. I think this was the reason why adding `aria-live` was not working for me.
```css
    &--hidden {
        width: 1px;
        height: 1px;
        padding: 0;
        margin: -1px;
        overflow: hidden;
        clip: rect(0 0 0 0); /* fallback for older browsers */
        clip-path: inset(50%);
        white-space: nowrap;
        border: 0;
    }
```

It sounds valid to me, what do you think?

m1aw: I think is worth giving it a try, but I'm unsure I understand the last part.

longyulongyu: When using `display:none` to toggle visibility, I have the feeling that the screen reader doesn't see or is too late to add the content to the accessibility tree (there is also a delay of 100ms to show/hide the tooltip), hence the content of the tooltip div is not always announced.
Using this approach can make sure the element is always visible to the screen reader.

PR #3366, opened 2025-06-24
feature: Extend onError and onComplete details in Await component

sethkrasnianski:
## Summary

This extends the `state.data.details` object for each of the mentioned callbacks to include additional context about the operation's outcome. Doing so enables developers to implement more nuanced response handling given the specific details of the failure or completion.

For my particular case, I needed a signal for when Blik's `onAdditionalDetails` callback, which stemmed from `Await`'s `onError` callback, had a `resultCode` value of `refused`. This was to facilitate a UI side effect, which I could otherwise not produce with the current `state.data.details` object data.

## Tested scenarios

I've updated the two existing test cases for the component to assert the presence of the two new properties.

**Fixed issue**:


ribeiroguilherme:
@sethkrasnianski is there any problem with getting the final result from the `/payment/details` request that is used in the `onAdditionalDetails` callback?

Adding new parameters to the callback response will break the API contract. There are merchants that simply forward the full object to their backend, and by adding a new parameter there it will break their API integration.

PR #3390, opened 2025-07-04
ACH - Fixing pay button label

ribeiroguilherme:
## Summary
Component pay button now displays "Pay {value}" instead of "Continue purchase"

### DIFF
--- a/packages/lib/src/components/Ach/components/AchComponent.tsx
+++ b/packages/lib/src/components/Ach/components/AchComponent.tsx
@@ -11,6 +11,7 @@ import { achValidationRules, achFormatters } from './validate';
 import StoreDetails from '../../internal/StoreDetails';
 import useSRPanelForAchErrors from './useSRPanelForACHErrors';
 import useImage from '../../../core/Context/useImage';
+import { PREFIX } from '../../internal/Icon/constants';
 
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/Ach/components/AchComponent.tsx
```diff
@@ -188,7 +190,7 @@ function AchComponent({ onChange, payButton, showPayButton, placeholders, hasHol
 
             {enableStoreDetails && <StoreDetails disabled={isFormDisabled} onChange={setStorePaymentMethod} />}
 
-            {showPayButton && payButton({ status, label: i18n.get('confirmPurchase') })}
+            {showPayButton && payButton({ status, icon: getImage({ imageFolder: 'components/' })('bento_lock') })}
```
m1aw: This should be more consistent.
For CardInput we use: 
```js
getImage({ imageFolder: 'components/' })(`${PREFIX}lock`)
```
Personally I don't like that we are defining the image that we load in every single component, and sometimes differently.

ribeiroguilherme: pushed fix

m1aw (approved):

PR #3394, opened 2025-07-09
Fix/e2e tests failing for dual brand UI

sponglord:
## Summary
Several e2e tests began to fail because they were dependent on a particular brand order, and sorting rules, for dual branded cards. Recent api changes have removed these rules, and brand order is now arbitrary

This PR fixes those failing tests

## Tested scenarios
Failing tests pass

**Fixed issue**:

### DIFF
--- a/.changeset/weak-cats-invent.md
+++ b/.changeset/weak-cats-invent.md
@@ -1,5 +1,0 @@
----
-'@adyen/adyen-web': patch
----
-
-Fixing e2e tests that are failing because they were depending on a fixed brand order. Recent api changes mean order is now arbitrary


.changeset/weak-cats-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
```
longyulongyu: I think we don't need changeset for non source code, it doesn't need to be in the [CHANGELOG](https://github.com/Adyen/adyen-web/blob/main/packages/lib/CHANGELOG.md)

longyulongyu (approved):

PR #3401, opened 2025-07-18
Fix: remove Klarna type from global Window object

Vayren:
## Summary
There is a conflict between Adyen Klarna type and custom Klarna type that any merchant can add for TS Klarna support.

Adyen Klarna types:
`
declare global {
    interface Window {
        Klarna: any;
        klarnaAsyncCallback: any;
    }
}
`

A merchant may also add Klarna types to benefit from TS support:
`
declare global {
    interface Window {
        Klarna: IKlarna;
        klarnaAsyncCallback: IKlarnaAsyncCallback;
    }
}
`
In this case, a type conflict will prevent pushing the changes. So, to fix the issue, the suggested solution is to remove Adyen Klarna types, as it doesn't provide a useful type for Klarna.

### DIFF
--- a/packages/lib/src/components/Klarna/types.ts
+++ b/packages/lib/src/components/Klarna/types.ts
@@ -4,7 +4,6 @@ import { AdditionalDetailsData } from '../../core/types';
 
 declare global {
     interface Window {
-        Klarna: any;
         klarnaAsyncCallback: any;
     }
 }


ribeiroguilherme:
Hi @Vayren - we can't accept your proposal since this will break our code.

The correct approach here is to properly define the Typescript types for it.

Can you push the type definition for the `IKlarna`  and `IKlarnaAsyncCallback` ? These types can be added to `src/components/Klarna/types.ts`

Vayren:
@ribeiroguilherme thanks for the suggestion! I've updated the Klarna type in the global Window object. The only thing to keep in mind is that it's not a complete type for Klarna and may be updated in the future, since I haven't had an experience with full Klarna functionality on UI.

ribeiroguilherme:
@Vayren I took a look at the changes, but it does break our Klarna integration due to certain types that are defined/missing.

I was wondering - Can't you redeclare the same interface but with your specific types? It should override the "loose" typing provided by us.

I haven't tried myself, but something like:

```ts
// types/global-overrides.d.ts
declare global {
  interface Window {
    Klarna: IKlarna;
    klarnaAsyncCallback: IKlarnaAsyncCallback;
  }
}
```

Then add this file to your `tsconfig.json` and hopefully it should work.

PR #3403, opened 2025-07-23
fix(getActionHandler): `onAdditionalDetails` callback not exposing the `actions`

longyulongyu:
## Summary
When using `createFromAction` from the core, the `onAdditionalDetails` callback doesn't receive the `actions`. So merchants do not have a consistent way to call `actions.resolve` or `actions.reject`. This PR fixed it for actions created by `getActionHandler` i.e. , `await` & `qrCode` actions.

Those actions internally call `onComplete`, externally merchants' `onAdditionalDetails` should be triggered , and we should follow the same logic inside `UIElement.handleAdditionalDetails` or `submitDetails` from the core when UIElement is not presented.

## Tested scenarios
Added unit test

Manually tested: 
`MBWay` (`await` action) & both `PromptPay` & `BCMC_mobile` (`qrCode` actions) 
with the scenarios where the action is handled by `createFromAction`; and also by `handleAction` 
\- to see that the `actions` object is always present in the `onAdditionalDetails` callback and that `onPaymentCompleted` is always subsequently called

Tested the `redirect` action to see that this situation didn't arise (`checkout.submitDetails`, which is called after a redirect, creates its own `Promise`, with its own `resolve` & `reject` functions that can then be use to finalise the payment)

Tested the `sdk `action with the `Klarna` widget to see that the onComplete for the KlarnaPayments component directly calls it's `super.handleAdditionalDetails` function. So this issue doesn't arise.

Tested that this doesn't apply to `bankTransfer` actions. (They are concluded in a 3rd party app and not via `onAdditionalDetails`)

Tested that this doesn't apply to `voucher` actions, since they don't conclude in a call to `onAdditionalDetails`

### DIFF
--- a/packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
+++ b/packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
@@ -20,7 +20,9 @@ const getActionHandler = statusType => {
         const config = {
             ...props,
             ...action,
-            // onComplete should trigger `onAdditionalDetails` callback or `submitDetails` call for the sessions flow
+            // `onComplete` should trigger `onAdditionalDetails` callback for the advanced flow,
+            // or call `this.session.submitDetails` for the sessions flow.
+            // The above logic is handled in `UIElement.handleAdditionalDetails` or `core.submitDetails`
             onComplete: (state: AdditionalDetailsData, component?: UIElement) => {
                 if (component) {
                     // We use a type assertion to call the protected 'handleAdditionalDetails' method from the UIElement.


packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
```diff
@@ -20,7 +20,16 @@ const getActionHandler = statusType => {
         const config = {
             ...props,
             ...action,
-            onComplete: props.onAdditionalDetails,
+            // onComplete should trigger `onAdditionalDetails` callback or `submitDetails` call for the sessions flow
```
sponglord: Can you expand this comment a bit to explain why the presence, or not, of `component` indicates that it's a /sessions flow

sponglord (approved):

packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
longyulongyu: I wasn't clear about my comments. Not sure why I wrote it like that. Will update.

The determination of whether a session flow does not depend on the presence or absence of the UI component. Instead, this logic is encapsulated within `UIElement.handleAdditionalDetails` or `core.submitDetails`.

packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
```diff
@@ -20,7 +20,18 @@ const getActionHandler = statusType => {
...
+            // `onComplete` should trigger `onAdditionalDetails` callback for the advanced flow,
+            // or call `this.session.submitDetails` for the sessions flow.
+            // The above logic is handled in `UIElement.handleAdditionalDetails` or `core.submitDetails`
+            onComplete: (state: AdditionalDetailsData, component?: UIElement) => {
+                if (component) {
+                    // We use a type assertion to call the protected 'handleAdditionalDetails' method from the UIElement.
+                    // This is safe because this is internal framework code.
+                    (component as unknown as { handleAdditionalDetails: (state: AdditionalDetailsData) => void }).handleAdditionalDetails(state);
+                } else {
+                    core.submitDetails(state.data);
```
sponglord, line 32: How can I test this part of the flow, i.e. that `core.submitDetails` gets called?
Surely `component` will always be present since, by the time this function is called, a component (UIElement) will have been created to handle the action?

longyulongyu: For a quick test, I've temporarily moved `core.submitDetails(state.data);` into the if statement, as I couldn't find any existing code actions that satisfy the condition. Despite this, I've still implemented the logic to check for the UIElement's presence, just in case.

PR #3404, opened 2025-07-25
Fixing issue with actions in onAdditionalDetails, in the core

sponglord:
### Description:
Some payment actions that lead to a call to `onAdditionalDetails`, when initiated via `checkout.createFromAction`, were seeing `onAdditionalDetails` being called without the `actions` argument containing the `resolve` & `reject` functions. These are needed to complete the flow, as detailed in our documentation.

This applies to `threeDS`, `await` & `qrCode` actions. 

Calling `checkout.createFromAction` means the component that is instantiated just receives the raw, merchant defined, `onAdditionalDetails` callback, and not the "wrapped" version of this callback that is generated by `UIElement.handleAction`. 

It is this latter, wrapped, version, called within ` UIElement.handleAdditionalDetails`, that creates a `Promise` that can then be finalised with the `resolve` & `reject` functions that are passed, as the third argument, to the merchant's own `onAdditionalDetails` callback.
Our documentation states that the merchant callback should expect to receive these 2 functions and that they should be called to properly finalise the flow & lead to `onPaymentCompleted` or `onPaymentFailed` being called.

### Fix:
This PR fixes this problem, within `core.ts > createFromAction`, by ensuring that the call to `onAdditionalDetails` (referenced internally as the "onComplete" fn) is always pushed towards the Component (UIElement) `handleAdditionalDetails` function

## Tested scenarios
- Manually tested `threeDS`, `await` (via `MBWay`) & `qrCode` (via `PromptPay`) actions, both with `handleAction` & `createFromAction`, to see them complete correctly in both scenarios.

- Also tested `threeDS` in the "MDFlow", to see that the `onComplete` function passed by _this_ flow is called _instead_ of any others.

- Tested the `redirect` action to see that this situation didn't arise (`checkout.submitDetails`, which is called after a redirect, creates its own `Promise`, with its own `resolve` & `reject` functions that can then be use to finalise the payment)

- Tested the `sdk` action with the Klarna widget to see that the `onComplete` for the `KlarnaPayments` component directly calls it's `super.handleAdditionalDetails` function. So this issue doesn't arise.

- Tested that this doesn't apply to `bankTransfer` actions. (They are concluded in a 3rd party app and not via `onAdditionalDetails`)

- Tested that this doesn't apply to `voucher` actions, since they don't conclude in a call to `onAdditionalDetails`

- Also tested how a card and storedCard within the Dropin performed, to ensure the `elementRef` (which points to the `DropinElement`) is persisted through the 3DS2 flow and reaches the `ThreeDS2Challenge` component.

- Tested that both `await` & `qrCode` components also performed correctly within the Dropin

- All e2e tests pass
- Additional unit tests added

### DIFF
--- a/packages/playground/src/pages/ThreeDS/ThreeDS.js
+++ b/packages/playground/src/pages/ThreeDS/ThreeDS.js
@@ -45,7 +45,6 @@ import { handleOnPaymentCompleted, handleOnPaymentFailed } from '../../handlers'
                     //         isMDFlow: true
                     //     })
                     //     .mount('.threeds-field');
-
                 } else {
                     actions.resolve({
                         resultCode,


longyulongyu:
Can you rebase from `main` and solve the conflicts? There are couple of commits ahead of this PR and it would be nice to keep it update to date and pass all the e2e tests first.

packages/lib/src/components/PayByBankPix/stories/SimulatedMerchantPage.tsx
```diff
@@ -1,4 +1,4 @@
-import { h } from "preact";
+import { h } from 'preact';
```
m1aw, line 1: Extremely minor but does anyone know why this is happening? Ideally we don't have styling changes in such a big commit.

packages/lib/.storybook/theme.js
```diff
@@ -1,21 +1,21 @@
 import { create } from '@storybook/theming';
 
 export default create({
-  base: 'dark',
-  brandTitle: 'Adyen Web Components',
-  brandTarget: '_self',
+    base: 'dark',
```
m1aw, line 4: Same here, styling changes just make it harder to review the PR and focus on the changes.

packages/lib/src/components/PayByBankPix/stories/SimulatedMerchantPage.tsx
sponglord: I've no idea why this happened - somewhere in the merge there was some kind of linting misfire

packages/lib/.storybook/theme.js
sponglord: Is it a 2 space vs 4 space misalignment in the linter?

PR #3417, opened 2025-07-30
deploy server on netlify

m1aw:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/server/.netlify/functions/hello.mjs
+++ b/packages/server/.netlify/functions/hello.mjs
@@ -1,0 +1,3 @@
+export default async (req, context) => {
+    return new Response("Hello, world!");
+  };


packages/server/index.js
```diff
@@ -62,6 +62,10 @@ module.exports = (app = express(), options = {}) => {
 
     app.all('/api/donations', (req, res) => createDonation(res, req.body));
 
+    app.all('/api/health', (req, res) => {
```
longyulongyu, line 65: Why do we need the /api/health endpoint?

longyulongyu:
Could you provide a brief description of this pull request and the core idea behind it?

packages/server/index.js
m1aw: Yeah, it's nice for debugging the serverless functions. I think we should keep it. Otherwise to test we need to make proper `/paymentMethods` or `/payments` call.

PR #3420, opened 2025-08-01
feat: UPI UI improvements

longyulongyu:
## Summary
Improved the user experience for the UPI component:

- **On mobile devices,** the QR code flow has been removed. Shoppers will be guided to pay with their installed UPI apps (Intent flow) or by entering their UPI ID.
- **On desktop devices,** the QR code flow is now the default, with the option to pay via UPI ID also available.
- Added a validation alert to guide shoppers to select a UPI app before proceeding with the payment.
- Updated instructional labels for better clarity across all payment modes.

## Tested scenarios
Added unit tests

**Fixed issue**:  COSDK-439


packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -115,6 +125,7 @@ class UPI extends UIElement<UPIConfiguration> {
                         type={paymentMethodType}
                         showCountdownTimer
                         shouldRedirectAutomatically
+                        showAmount
```
ribeiroguilherme, line 128: I expected the `showAmount` to default to true and not be needed here, since we usually always display the amount

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -69,23 +77,25 @@ export default function UPIComponent({ defaultMode, onChange, onUpdateMode, payB
 
             setSelectedApp(app);
             setIsValid(true);
+            setStatus('success');
```
ribeiroguilherme: Do we need to setStatus to `success` here? Does it do something?

`success` in general means that the payment was completed, but in this particular case it means that one app from the intent list was selected. It seems a bit weird to me

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -69,23 +77,25 @@ export default function UPIComponent({ defaultMode, onChange, onUpdateMode, payB
...
+        }
+    }, [selectedApp]);
 
     useEffect(() => {
         if (mode !== 'qrCode') {
             onChange({
-                data: { ...(vpa && { virtualPaymentAddress: vpa }), ...(selectedApp && { app: selectedApp }) },
-                errors,
-                valid,
+                data: { ...(selectedApp && { app: selectedApp }) },
```
ribeiroguilherme, line 98: Don't we need to pass the `virtualPaymentAddress`  anymore in case it is available?

packages/lib/src/components/UPI/types.ts
```diff
@@ -1,9 +1,10 @@
 import { UIElementProps } from '../internal/UIElement/types';
 import { TxVariants } from '../tx-variants';
+import { SegmentedControlOption } from '../internal/SegmentedControl/SegmentedControl';
 
 export type UpiType = TxVariants.upi_qr | TxVariants.upi_intent | TxVariants.upi_collect;
 
-export type UpiMode = 'vpa' | 'qrCode' | 'intent';
+export type UpiMode = 'vpa' | 'qrCode' | 'intent' | '';
```
ribeiroguilherme: Do we need to support an empty value for it? That seems a bit off

packages/lib/storybook/stories/internals/Alert.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+                loadingContext={'test'}
+                i18n={
+                    new Language({
+                        locale: 'en-US',
+                        translations: {}
+                    })
+                }
+                resources={new Resources('https://checkoutshopper-test.cdn.adyen.com/checkoutshopper/')}
+            >
+                <Alert {...args}>dummy alert message</Alert>
```
ribeiroguilherme, line 35: Interesting that a simple UI component has a direct dependency on having the CoreProvider + Resources + Language modules in order for it to work. Perhaps this needs to be more decoupled at some point.

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -31,29 +34,36 @@ class UPI extends UIElement<UPIConfiguration> {
...
+    formatProps(props: UPIConfiguration): UPIConfiguration {
         if (!isMobile()) {
             return {
                 ...super.formatProps(props),
-                defaultMode: props?.defaultMode ?? 'qrCode',
-                // For large screen, ignore the apps
-                apps: []
+                defaultMode: ['qrCode', 'vpa'].includes(props?.defaultMode) ? props.defaultMode : 'qrCode',
+                apps: [], // For desktop, ignore the apps
+                segmentedControlOptions: [getQrOption(props.i18n), getVpaOption(props.i18n)]
```
ribeiroguilherme, line 43: Not a big deal, but in this case we are mixing the `segmentedControlOptions` with the `props` which are usually passed by the merchant.

Perhaps that can be achieved differently by just creating a private field to this class and assign to it the `segmentedControlOptions` (same as we do with the `selectedMode`). Then we don't have to expose internal props as part of the `UPIConfiguration` type definition

ribeiroguilherme (approved):

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
longyulongyu: I don't really need a `success` status as long as it's not `error` or `loading`. `ready` also sounds good to me.

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
longyulongyu: `vpa` calls `onChange` directly at line 136. This `useEffect` is basically effective for QR and Intent flows.

packages/lib/src/components/UPI/UPI.tsx
longyulongyu: Actually there are more components that don't show the amount, only `PayTo` and the new `UPI` show it.😄

packages/lib/src/components/UPI/types.ts
longyulongyu: Nice catch, will remove it.

packages/lib/src/components/UPI/UPI.tsx
longyulongyu: I thought about it but didn't go with it because then i need to repeat some of the business logic from `formatProps`.

packages/lib/src/components/UPI/UPI.tsx
ribeiroguilherme: Oh really? I thought that was the default behavior 😄

PR #3425, opened 2025-08-08
fix(installments): correct missing border and background styles

longyulongyu:
## Summary
The installments dropdown was missing its border and background color, making it appear transparent against the page background. This change applies the correct styling to fix the visual defect.

## Tested scenarios
Both mc card and visa card should have correct styles for the installments dropdown.
https://localhost:3020/?path=/story/cards-card--with-installments

**Fixed issue**:  COSDK-573

### DIFF
--- a/.changeset/pretty-adults-peel.md
+++ b/.changeset/pretty-adults-peel.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed a styling issue where the installments dropdown was missing its border and background color.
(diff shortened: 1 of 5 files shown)


m1aw (approved):
Looks good the change. But what caused the problem? The installments style got changed or overriden somewhere?

longyulongyu:
> Looks good. But what caused the problem? The installments style got changed or overriden somewhere?

The installment dropdown border was removed for all types of installments, but it should only be removed for the "revolving" type.

Screenshot 2025-08-11 at 11 39 17

m1aw:
I see the last changed to this was a year ago? Was that the case?

longyulongyu:
> I see the last changed to this was a year ago? Was that the case?

I suppose, checked the commit message - `* Removing border and repositioning radioGroup's installments dropdown` which i suppose it's for the revolving plan, and overlooked the styles for the regular installments.

Issue #3419, opened 2025-07-31
Missing `billingAddress` field in `PaymentData` interface


obber:
**Describe the bug**

I am trying to extract `state.data.billingAddress` to send to our server for the initial `/payments` API call in an advanced flow.

The TS definition for the `PaymentData` interface is missing.

Reference source: https://github.com/Adyen/adyen-web/blob/6b9003ebd2d4a6825bb8ce99cf673ae0aba36e44/packages/lib/src/types/global-types.ts#L290-L301

**To Reproduce**

1. Open up a Typescript Project that uses the `@adyen/adyen-web` package.
2. Import the `PaymentData` type to observe the missing field. 

**Expected behavior**

Expected `PaymentData.billingAddress` to be defined.

**Screenshots**

Auto-complete in my editor:

<img width="903" height="622" alt="Image" src="https://github.com/user-attachments/assets/cbde0707-7731-4392-afef-2352a839499b" />

However, I do see the field is present in the actual JS console:

<img width="1160" height="558" alt="Image" src="https://github.com/user-attachments/assets/90ad766c-86da-4472-bde1-bba383d52027" />

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser: Chrome
 - Version: 138.0.7204.169

m1aw:
Hey @obber thank you for the feedback, we will make a fix for this soon.

obber:
> Hey [@obber](https://github.com/obber) thank you for the feedback, we will make a fix for this soon.

Thanks @m1aw! Appreciate it.

ribeiroguilherme:
@obber We released 6.20.0 which contains the fix for this.  I am closing this thread for now, but feel free to reach out if the problem persists. Cheers

PR #3431, opened 2025-08-08
fix missing props in PaymentData

m1aw:
## Summary

Add missing properties to the `PaymentData` interface to align with the data structure expected in payment submission flows.
This aligns with what is right now on `BeforeSubmitActions`.

## Tested scenarios

**Fixed issue**:  
#3419

### DIFF
--- a/packages/lib/src/types/global-types.ts
+++ b/packages/lib/src/types/global-types.ts
@@ -300,8 +300,6 @@ export interface PaymentData extends PaymentMethodData {
     storePaymentMethod?: boolean;
     billingAddress?: AddressData;
     deliveryAddress?: AddressData;
-    shopperEmail?: string;
-    shopperName?: string;
 }
 
 export type ResultCode =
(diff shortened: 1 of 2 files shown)


packages/lib/src/types/global-types.ts
```diff
@@ -298,6 +298,10 @@ export interface PaymentData extends PaymentMethodData {
     clientStateDataIndicator: boolean;
     sessionData?: string;
     storePaymentMethod?: boolean;
+    billingAddress?: AddressData;
+    deliveryAddress?: AddressData;
+    shopperEmail?: string;
+    shopperName?: string;
```
ribeiroguilherme: Can we actually return `shopperEmail` and `shopperName` here? Which payment method does that?

m1aw: I think I have a new idea how to fix in a proper way. Let me try some typescript ideas.

m1aw: I tried a few things but it was a way bigger rework of how we type things. I have fixed for this use case.

PR #3447, opened 2025-08-14
Stop preselecting a card brand in dual branding scenarios outside of EU

sponglord:
## Summary
Stop preselecting a card brand in dual branding scenarios outside of Europe. This will allow us to continue doing low cost routing outside of Europe.

## Tested scenarios
Manually tested with dual brands that are present in the "inclusion" list, and those that aren't, to see that:
- the `brand` is only preselected when the brand is in the inclusion list.
- the dual branding buttons are only shown when the brand is in the inclusion list.

Added new e2e test to check the above

All existing e2e tests pass

**Relates to issue**: #3387

### DIFF
--- a/packages/e2e-playwright/tests/ui/card/binLookup/branding/dualBranding.spec.ts
+++ b/packages/e2e-playwright/tests/ui/card/binLookup/branding/dualBranding.spec.ts
@@ -21,6 +21,10 @@ test.describe('Card - Testing full UI (PAN icons & dual branding buttons) after 
             // Get a binLookup result
             await card.typeCardNumber(BCMC_DUAL_BRANDED_VISA);
 
+            // Check brand has been set, by default, in paymentMethod data
+            let cardData: any = await page.evaluate('window.component.data');
+            expect(cardData.paymentMethod.brand).not.toBe(undefined);
+
             /**
              * Dual brand icons
              */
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/Card/components/CardInput/utils.ts
```diff
@@ -341,3 +341,16 @@ export const mapDualBrandButtons = (dualBrandSelectElements: DualBrandSelectElem
...
+ *
+ * If the result from Array.filter contains anything - then we are in a EU dual branding regulation scenario, i.e.
+ * - Show the new dualBranding UI Buttons
+ * - Preselect a card brand
+ */
+export const mustHandleDualBrandingAccordingToEURegulations = (
+    EU_BrandArray: string[],
+    returnedDualBrandingObjects: DualBrandSelectElement[] | BrandObject[],
+    key: string
+) => !!returnedDualBrandingObjects.filter(item => EU_BrandArray.includes(item[key])).length;
```
longyulongyu, line 356: We can just use `some` which returns a boolean by default:
`returnedDualBrandingObjects.some(item => EU_BrandArray.includes(item[key]))`

sponglord: I'll refine this in another PR - I'm in a different branch now and can't easily switch out to make this change

PR #3449, opened 2025-08-14
chore(storybook): Enable a11y testing mocks

m1aw:
## Summary
The objective is to enable a few mocks that are going to be necessary for a11y testing. This also updates/fixes MSW "service worker" since it was not working properly under certain scenarios.

Added mock to the following payment methods:
- PayTo

## Tested scenarios

**Fixed issue**:


packages/e2e-playwright/automated-a11y.config.ts
```diff
@@ -71,7 +71,7 @@ const config: PlaywrightTestConfig = {
     /* Run your local dev server before starting the tests */
     webServer: [
         {
-            command: 'npm run build:storybook && npm run start:prod-storybook',
+            command: 'DISABLE_MSW=true npm run build:storybook:e2e && npm run start:prod-storybook',
```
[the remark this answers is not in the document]
longyulongyu: I had the same doubt as Gemini.

packages/e2e-playwright/playwright.config.ts
longyulongyu, line 1: I am getting runtime errors  when using the ui mode, and using https. I think it can be reproduced by any playwright test. Maybe because of recent node upgrade? I didn't have the issue before.

```
@ base-fixture.ts:18
TypeError: Protocol "https:" not supported. Expected "http:"
```

m1aw: Strange let me check.

m1aw: I couldn't reproduce, put something on your calendar we can go over this.

longyulongyu: Pushed a [commit](https://github.com/Adyen/adyen-web/pull/3449/commits/5c60ab92a2d90e7e78ef917199771b7cee23f95c) as we agreed.

PR #3451, opened 2025-08-15
fix(pay-button): alignment and font size for the secondary amount label

longyulongyu:
## Summary
Fixed alignment and font size for the Pay button's secondary amount label.

## Tested scenarios
Added story: https://localhost:3020/?path=/story/internals-button--payment-button

**Fixed issue**:  COSDK-150

### DIFF
--- a/packages/lib/src/components/internal/PayButton/PayButton.test.tsx
+++ b/packages/lib/src/components/internal/PayButton/PayButton.test.tsx
@@ -45,6 +45,18 @@ describe('PayButton', () => {
         expect(button).not.toHaveTextContent(PAY_BTN_DIVIDER);
     });
 
+    test('should not render a secondary amount if primary amount is not correct', () => {
+        renderPayButton({
+            amount: {
+                currency: '',
+                value: 1000
+            },
+            secondaryAmount: { currency: 'HRK', value: 7534 }
+        });
+        const button = screen.getByRole('button', { name: 'Pay' });
+        expect(button).not.toHaveTextContent(PAY_BTN_DIVIDER);
+    });
+
     test('should not render any amount if a specific label is provided', () => {
         renderPayButton({
             label: 'Redirect to',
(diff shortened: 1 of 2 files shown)


packages/lib/storybook/stories/internals/Button.stories.tsx
```diff
@@ -55,36 +62,44 @@ export const Default: StoryObj = {
...
     argTypes: {
-        disabled: { control: 'boolean' },
-        inline: { control: 'boolean' },
-        ariaLabel: { control: 'text' }
+        amount: { control: 'object' },
+        secondaryAmount: { control: 'object' }
     },
     args: {
-        disabled: false
+        amount: { value: 1000, currency: 'EUR' },
```
ribeiroguilherme, line 97: When the `amount` parameter is defined but it is empty, or the value/currency are not properly set, we still display the secondary amount.

Should we add a check to not let that happen? We should show the secondary amount only if the amount is properly defined

packages/lib/storybook/stories/internals/Button.stories.tsx
```diff
@@ -55,36 +62,44 @@ export const Default: StoryObj = {
...
+        secondaryAmount: { value: 1200, currency: 'USD' },
```
ribeiroguilherme, line 98: Did you verify if we are handling it correctly for the Bulgarian symbol? Is it normal to show `BGN` (as mentioned in the ticket) ?

ribeiroguilherme (approved):

packages/lib/storybook/stories/internals/Button.stories.tsx
longyulongyu: It's normal to show `BGN` when the `shopperLocale` is not `bg-BG`.

If the locale is `bg-BG`, we are displaying `лв` correctly:
Screenshot 2025-08-15 at 17 02 49

packages/lib/storybook/stories/internals/Button.stories.tsx
longyulongyu: I can check if the amount is empty or not `Object.keys(amount).length>0`, but do we need to further check if value is a number and currency is a valid string? It feels checking too much.
Or checking specifically that the amount value and currency should be defined?

ribeiroguilherme: Maybe we can have (or we might already have 🤔 ) an utility somewhere that does this?

```js
const isAmountValid = amount => !!amount?.value && !!amount?.currency
```

longyulongyu: Sounds good, will update.

PR #3461, opened 2025-08-21
Feature/use v6 secured fields

sponglord:
## Summary
Use `v6.0.0` SecuredFields, which: 
- bumped the version of the JWE package, `jose`, to `5.10.0` (from `4.15.9`)
- no longer creates an ACH bundle, since in AdyenWeb `v6.12.0`, the ACH  components stopped using securedFields in its inputs
- disallows the "compat" version on Live, 

Also:
- Refactored the util in `CardInput/utils.ts` that checks if we are in a Europe based dual branding scenario, to use `Array.some` rather than `Array.filter`
- Tiny refactor to make it a bit more obvious when we do, or don't, send analytics, after the initial setup call.

## Tested scenarios

**Fixed issue**:


ribeiroguilherme:
I see that there is a feature flag `useModern` - are there specific merchants waiting for this feature? are we planning to do any A/B tests with PBL?

Asking because I discussed with the team and we are wondering why not enable this to all merchants.

sponglord:
> I see that there is a feature flag `useModern` - are there specific merchants waiting for this feature? are we planning to do any A/B tests with PBL?
> 
> Asking because I discussed with the team and we are wondering why not enable this to all merchants.

There's no specific merchants waiting for this feature. More a general sense of file size still (and always) being an issue, coupled with Q&A issues talking about the latency around the securedFields loading time.

As for making it true by default... maybe.
The current bundle (i.e v5.x) is compiled to support ES11
This v6 bundle at it's lowest setting compiles to ES2017 - which is in line with our lowest "setting" for CheckoutWeb.
The `useModern` flag compiles it to something more in line with ES2020.

So the question is which level should we set, for securedFields, by default?

ribeiroguilherme:
What is the difference in file size when using the normal version vs the modern?

> So the question is which level should we set, for securedFields, by default?

Yes - the question is if we should enforce the modern bundle (ES2020) straight away or if we should not provide this option at all and just offer the ES2017. 

I don't think we should introduce 'feature flag' here. That is not something that we generally do. We need to make a decision on which ES version we want to support and we go for it. Besides that, this also introduce different bundle types for SF which is not nice.  We should keep it simple, merchant does not need to care about it.

This also adds communication overload - we need to explain to merchants that `useModern` uses ES2020 which is X% smaller, but it _might_ not work on all browsers and they will have to test by themselves, and that will bring more questions. This will also raise expectations that loading time will decrease, although I don't believe that this is the root issue for the latency problem.

I am in the opinion that Secured Fields is very critical and if we are afraid of ES2020, then we just go for ES2017 directly. If the size difference from both versions (ES2017 vs ES2020) is very big and if we think that is safe going ES2020, then we take the risk and we go for it.

sponglord:
> What is the difference in file size when using the normal version vs the modern?

Unzipped: 174Kb (normal) versus 102Kb (modern)
Zipped about: 43Kb versus 25Kb

We do already have multiple bundle types for securedFields - we have the "compat" bundle that we historically pointed merchants to for IE11 support and that we still point merchants to if they need to run a custom http domain for testing.

Similarly we produce for different bundles for CheckoutWeb - so I'm not of the opinion that having different bundles is a problem.

We also have many feature flags already in existence for the card component - so one more doesn't really make a difference.

The reality in the real world is that one size does not fit all.

However, I do think we could just go with the ES2020 bundle, making the `useModern` flag `true` by default and then, as a defensive measure,  having the option to point merchants back to the ES2017 build if they run into problems.

packages/lib/src/components/Card/types.ts
```diff
@@ -410,6 +410,13 @@ export interface CardConfiguration extends UIElementProps {
      */
     supportedShopperInteractions?: string[];
 
+    /**
+     * For the PAN field: If the card number is now at one of it's valid lengths and the formatting process has added a separator
+     * to the end of the string... delete the separator
+     * - merchant set config option. Defaults to true
+     */
+    trimTrailingSeparator?: boolean;
+
```
ribeiroguilherme, line 419: Where is this property being used? I see it being added but I don't see where it is being used. 

Also, why are we adding this? was this requested by a merchant? I didn't fully understand its use-case by the description

sponglord: It gets sent thorough to the securedFields
I'm adding this so we see it in the `configData` (It got missed first time round). 
It has _always_ been there but we don't see it logged in the `configData`. I want it there so we can see if merchants ever use it

PR #3464, opened 2025-08-22
[WIP ] Fastlane analytics

ribeiroguilherme:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/PayPalFastlane/FastlaneSDK.ts
+++ b/packages/lib/src/components/PayPalFastlane/FastlaneSDK.ts
@@ -1,23 +1,31 @@
 import { resolveEnvironments } from '../../core/Environment';
+import type { FastlaneTokenData } from './services/request-fastlane-token';
 import requestFastlaneToken from './services/request-fastlane-token';
 import { convertAdyenLocaleToFastlaneLocale } from './utils/convert-locale';
 import Script from '../../utils/Script';
 import AdyenCheckoutError from '../../core/Errors/AdyenCheckoutError';
-import type {
-    FastlanePaymentMethodConfiguration,
-    FastlaneWindowInstance,
+import {
     FastlaneAuthenticatedCustomerResult,
-    FastlaneShippingAddressSelectorResult,
+    FastlaneConsentRenderState,
+    FastlanePaymentMethodConfiguration,
     FastlaneSDKConfiguration,
-    FastlaneConsentRenderState
+    FastlaneShippingAddressSelectorResult,
+    FastlaneSignupConfiguration,
+    FastlaneWindowInstance
 } from './types';
-import type { FastlaneTokenData } from './services/request-fastlane-token';
+
+import Analytics from '../../core/Analytics';
+import type { AnalyticsModule } from '../../types/global-types';
+import { AnalyticsInfoEvent } from '../../core/Analytics/AnalyticsInfoEvent';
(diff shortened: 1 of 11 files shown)


packages/lib/src/components/PayPalFastlane/FastlaneSDK.ts
```diff
@@ -30,24 +37,43 @@ class FastlaneSDK {
...
             console.warn("Fastlane SDK: 'forceConsentDetails' should not be used on 'live' environment");
 
-        const { apiUrl } = resolveEnvironments(configuration.environment);
+        const { apiUrl, analyticsUrl } = resolveEnvironments(configuration.environment);
 
         this.checkoutShopperURL = apiUrl;
         this.clientKey = configuration.clientKey;
         this.forceConsentDetails = configuration.forceConsentDetails || false;
-        this.locale = convertAdyenLocaleToFastlaneLocale(configuration.locale || 'en-US');
+        this.fastlaneLocale = convertAdyenLocaleToFastlaneLocale(configuration.locale || 'en-US');
```
johnayeni, line 45: [Nice to have] Having an enum for these country codes would be great so that we can type them. We use the raw string in a number of places

johnayeni (dismissed):

packages/lib/src/components/PayPalFastlane/FastlaneSDK.ts
ribeiroguilherme: I see benefits in having them as enum, but i believe that there are two downsides of using enums for the `locale` property:

1. That means we would [type this](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/types.ts#L140) to be using the `enum`. In the past we learned that typing external configurations as `enum` is not as friendly as typing them as plain strings. (I don't remember the reason - perhaps the merchant is forced to import that `enum` type in order to set the Locale if they are using Typescript 🤔 we can revisit that)
2. Let's say we start supporting a new language today in the `locale` . The idea for now is that it is backwards compatible, so if the merchant passes the new `locale` on v6.0.0 it will just work fine. If we start adding specific types for `locale`, then that implies that we are versioning the translations support

packages/lib/src/utils/Script.ts
```diff
@@ -10,8 +10,7 @@ interface IScript {
 interface IScriptProps {
     src: string;
     component: string;
-    // TODO: The optional flag will be fixed on PR #3464
-    analytics?: AnalyticsModule;
+    analytics: AnalyticsModule;
```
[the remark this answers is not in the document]
ribeiroguilherme: You should have checked the whole codebase to see that what you are saying does not make sense

packages/lib/.size-limit.cjs
```diff
@@ -32,7 +32,7 @@ module.exports = [
     {
         name: 'UMD',
         path: 'dist/umd/adyen.js',
-        limit: '116 KB',
+        limit: '120 KB',
```
johnayeni, line 35: [Question] do we have a target size limit that we plan not to ever exceed for the bundle sizes?

johnayeni (approved):

packages/lib/.size-limit.cjs
ribeiroguilherme: Not really. This mechanism is in place in order to make sure that tree-shaking is working fine, but the codebase size  will always grow as long as we add more features to it 😬

PR #3465, opened 2025-08-26
[wip]feat(upi-autopay): added the support for upi autopay

longyulongyu:
## Tested scenarios

**Fixed issue**:  COSDK-601


packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.scss
```diff
@@ -17,6 +17,10 @@
     .adyen-checkout__button--pay {
         margin-top: token(spacer-070);
     }
+
+    .adyen-checkout__alert-message--info {
+        margin-top: token(spacer-070);
+    }
```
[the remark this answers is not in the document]
longyulongyu: Not true, will not fix.

packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.tsx
```diff
@@ -122,7 +128,7 @@ export default function UPIComponent({
                     <span className="adyen-checkout-upi-instruction-label">{i18n.get('upi.intent.instruction')}</span>
                     {status === 'error' && <Alert icon={'cross'}>{i18n.get('upi.error.noAppSelected')}</Alert>}
                     <UPIIntentAppList disabled={status === 'loading'} apps={apps} selectedAppId={selectedApp?.id} onAppSelect={handleAppSelect} />
-
+                    {mandate && <UPIMandate mandate={mandate} amount={amount} />}
```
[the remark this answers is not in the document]
longyulongyu: done

packages/lib/src/components/UPI/components/UPIMandate/UPIMandate.tsx
```diff
@@ -0,0 +1,55 @@
...
+        // rule === 'max' and no payment amount
+        return i18n.get('upi.mandate.max.withoutAmount', { values: { mandateAmount: formattedMandateAmount, frequency: frequencyText } });
+    }, [mandate, amount, i18n]);
+
+    if (!mandateText) return null;
+
+    return (
+        <Alert icon="info" type="info">
+            {i18n.get('upi.mandate.intro', { values: { mandateText } })}
+        </Alert>
```
ribeiroguilherme, line 65: I noticed few things that are different from the design. I am not sure if that is intentional or not, but I will highlight here:

- The Info icon has the wrong color (it should be black)
- The alert block is missing the gray border on it
- The amount labels are not in bold
- Can you double-check the padding and font-sizes? I can't tell by the screenshots if it is correct

packages/lib/src/components/UPI/components/UPIMandate/UPIMandate.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export interface UPIMandateProps {
+    mandate: Mandate;
+    amount?: PaymentAmount;
+}
+
+const UPIMandate = ({ mandate, amount }: UPIMandateProps): h.JSX.Element => {
+    const { i18n } = useCoreContext();
+
+    const mandateText = useMemo((): string => {
```
ribeiroguilherme: Maybe we can be more defensive in the sense that we apply validation to verify that the `mandate` is properly declared. If the required fields are missing, we throw a warning and do not render the UI. That could be done here or on the `UPI.tsx` level as well. What do you think?

Right now, if we pass empty values to the mandate fields or an empty mandate object, the component will render the `UPIMandate` with weird translation labels on it

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -113,6 +115,7 @@ class UPI extends UIElement<UPIConfiguration> {
                         introduction={this.props.i18n.get('upi.qrCodeWaitingMessage')}
                         countdownTime={5}
```
ribeiroguilherme: As part of the document we refined (`UPI - Q&A from Devs`), it was highlighted that we could add a configuration property to allow the countdown to be configured.  It seems that it was not included as part of the main ticket.

Do you think that this is something that can be added in this PR too?

ribeiroguilherme (dismissed):

packages/lib/src/components/UPI/components/UPIMandate/UPIMandate.tsx
longyulongyu: We don't really have the official figma design for upi autopay, it's copied from [this](https://www.figma.com/design/ShYUgz60fUmyqluaBke3lL8Z/01-%C2%B7-Drop-in-%C2%B7-Web?node-id=11982-9988&m=dev) component. 
However i think this Alert component is not aligned with Bento in the figma, we should just use the design from the Bento [one](https://adyen-bento.netlify.app/?path=/docs/components-alert--docs).

I can adjust our Alert component to be aligned with the Bento alert, but one obvious difference is that bento is using a solid blue svg as the icon, and it seems they are using inline svgs, but we are just loading the svg via `<image>`. The easiest way is to replace our info icon with a solid blue background though, what do you think?

> The amount labels are not in bold

I think i missed it, need to take a look how to adapt the changes.

packages/lib/src/components/UPI/UPI.tsx
longyulongyu: Sure we can add one.

packages/lib/src/components/UPI/components/UPIMandate/UPIMandate.tsx
longyulongyu: Sounds good, i'll adjust.

packages/lib/src/components/UPI/components/UPIMandate/UPIMandate.tsx
longyulongyu: It should be fixed now.

packages/lib/src/components/UPI/components/UPIMandate/UPIMandate.tsx
longyulongyu: Done

packages/lib/src/components/UPI/UPI.tsx
longyulongyu: Done

ribeiroguilherme (approved):

PR #3473, opened 2025-08-27
chore: Add script that marks old releases as deprecated

m1aw:
## Summary

Creates a script that can be used by running:

`DRY_RUN=false MAJOR_VERSION=X node .github/scripts/deprecate-releases.js`

The script fetchs all available version via the Github API. Loops thru them, and if they match the major version then it changes the release name/title to include `[DEPRECATED]`. 

To get your API token go to Your profile, and create a PAT with `Contents` R/W with the scope for this project.

## Tested scenarios

- Changes already made 🙃 

**Fixed issue**:

### DIFF
--- a/.github/scripts/deprecate-releases.js
+++ b/.github/scripts/deprecate-releases.js
@@ -1,5 +1,16 @@
 // A script to mark all GitHub releases in a repository as [DEPRECATED].
 // Uses the GitHub REST API and supports a dry run mode.
+// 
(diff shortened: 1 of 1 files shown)


ribeiroguilherme (approved):
Perhaps you can add the instructions (how to run the script; get the token and so on) in its header as code comments.  So we don't have to go back to the PR to find out more how to use it when deprecating the next major version.

PR #3475, opened 2025-08-28
fix(PayTo): SegmentedController aria-controls IDs

m1aw:
## Summary

There was an error caught by our a11y tool that PayTo SegmentedController wasn't using IDs that were pointing properly to the Fieldset. 
This PR makes that change. 
Fieldset now supports optional IDs.

## Tested scenarios

**Fixed issue**:


longyulongyu:
Don't forget to add a Changeset.

PR #3491, opened 2025-09-04
Analytics - Patching attemptId data 

ribeiroguilherme:
## Summary

Adjusted Analytics module to start performing the **setUp** call even if the attempt ID is saved in the Session storage.

This is useful in case shopper refreshes the page and another payment session is created. By doing the **setUp** another time, we can link the new payment session to the existing attempt ID.

Summary of changes:
- Moved the logic of checking the Session storage from `collect-id` to the Analytics **setUp** part. Collect ID now is responsible only for making the API request
- Removed the `loadingContext` and `amount` properties from the Analytics constructor, since they are not used 

## Tested scenarios
- Added unit tests
- Checked that the _setUp happens again_ with the same attempt ID if the the page is refreshed


packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -57,25 +76,36 @@
...
+                };
+
+                capturedCheckoutAttemptId = await collectId(collectIdPayload);
+
+                storage.set({
+                    id: capturedCheckoutAttemptId,
+                    timestamp: isSessionReusable ? checkoutAttemptIdSession.timestamp : Date.now()
+                });
+            } catch (error: unknown) {
+                console.warn(error);
```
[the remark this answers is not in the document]
ribeiroguilherme: The error message propagated here mentions the issue

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -6,20 +6,39 @@ import { debounce } from '../../utils/debounce';
...
+let capturedCheckoutAttemptId: string | null = null;
 let sendEventsTimerId = null;
 
-const Analytics = ({ locale, clientKey, analytics, amount, analyticsContext, bundleType }: AnalyticsProps): AnalyticsModule => {
+/**
+ * If the checkout attempt ID was stored more than fifteen minutes ago, then we should request a new ID.
+ * More here: COWEB-1099
+ */
+function isSessionCreatedUnderFifteenMinutes(session: CheckoutAttemptIdSession): boolean {
+    const FIFTEEN_MINUTES_IN_MS = 15 * 60 * 1000;
```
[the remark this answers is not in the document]
ribeiroguilherme: This variable is used only here. The performance/clarity improvement is very minimal. I won't do that.

packages/lib/src/core/Services/analytics/collect-id.ts
```diff
@@ -46,31 +36,9 @@ const collectId = ({ analyticsContext, clientKey, locale, analyticsPath, bundleT
...
-                }
-                return undefined;
-            })
-            .catch(() => {
-                return Promise.reject(FAILURE_MSG);
-            });
+        memoizedPromise = httpPost<{ checkoutAttemptId: string }>(options, telemetryEvent).then(conversion => conversion?.checkoutAttemptId);
 
-        return promise;
+        return memoizedPromise;
```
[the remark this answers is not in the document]
ribeiroguilherme: Done

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -57,25 +76,36 @@ const Analytics = ({ locale, clientKey, analytics, amount, analyticsContext, bun
...
-                } catch (e: any) {
-                    console.warn(`Fetching checkoutAttemptId failed.${e ? ` Error=${e}` : ''}`);
-                }
+
+            const availableCheckoutAttemptId: string | undefined = isSessionReusable ? checkoutAttemptIdSession.id : analyticsData?.checkoutAttemptId;
+
+            try {
+                const collectIdPayload = {
+                    ...finalSetupProps,
+                    ...payload,
```
sponglord, line 97: So now you're always sending an entry for `payload` and `analyticsData` even if they're not defined?
Just wondering if this affects what the data team expect to receive

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -86,6 +116,10 @@ const Analytics = ({ locale, clientKey, analytics, amount, analyticsContext, bun
 
         getEnabled: () => props.enabled,
 
+        flush: () => {
+            void sendAnalyticsEvents();
+        },
+
```
sponglord, line 124: What's this function for?

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -57,25 +76,36 @@ const Analytics = ({ locale, clientKey, analytics, amount, analyticsContext, bun
...
 
     return {
-        /**
-         * Make "setup" call, to pass containerWidth, buildType, channel etc, and receive a checkoutAttemptId in return
-         * @param initialEvent -
-         */
-        setUp: async (initialEvent: AnalyticsInitialEvent) => {
-            const { payload, enabled } = props; // TODO what is payload, is it ever used?
+        setUp: async (setupProps?: AnalyticsInitialEvent): Promise<void> => {
+            const defaultProps: Partial<AnalyticsInitialEvent> = {};
```
sponglord: Why are you defining an empty object?

packages/lib/src/core/Analytics/Analytics.ts
ribeiroguilherme: It is used to send all the queued events without the need to wait for the interval. It is going to be used in the Fastlane analytics in the next PR

packages/lib/src/core/Analytics/Analytics.ts
ribeiroguilherme: It is going to be used in the next PR - It will be used to define the default value of `checkoutStage` variable.

To give you more context: 
I made all the changes assuming that the API would be accepting the `checkoutStage` property, but apparently it is not live yet, so I removed the code using this new field in order to release this changes sooner, but I will add this back into the Fastlane analytics PR

packages/lib/src/core/Analytics/Analytics.ts
ribeiroguilherme: I am not adding a new entry there, I am just spreading the object. It is the same logic as it was there before

sponglord: Fair enough!

packages/lib/src/core/Analytics/Analytics.ts
sponglord: Maybe you could add a comment in the future explaining that this is for a future PR? Then we wouldn't need to have this conversation! :-P

sponglord (approved):

packages/lib/src/core/Analytics/Analytics.ts
longyulongyu: I think we can move everything from line 80-91 to the try block, because it should be non-blocking execution and we don't care if anything fails in the analytics.

Optional: if we can differentiate different errors in order to pinpoint an exact one.

longyulongyu (approved):

PR #3509, opened 2025-09-11
Pix - Improve pix code experience

johnayeni:
## Summary
- Update QR code UI for PIX payment to switch between displaying QR code and copying code text
- Display full bank code for PIX so users can also highlight and copy the code directly
- Decompose QR loader component into smaller components
- Remove checkmark from the Segmented control component and improve the button responsiveness

## Tested scenarios

**Fixed issue**:


packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -228,54 +246,21 @@ class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
...
-                                this.props.onSubmitAnalytics(event);
-
-                                complete();
-                            }}
-                            icon={getImage({ imageFolder: 'components/' })(`${PREFIX}copy`)}
-                            label={i18n.get('button.copy')}
-                        />
-                    </div>
-                )}
+                <QRLoaderDetailsProvider
```
m1aw, line 249: What advantage do we get by using a provider here?

johnayeni: The main advantage here is to reduce prop drilling for the child component. I could also just use render props but one other reason for this was to reduce the number of places I also needed to add props for components. What do you think?

m1aw: I guess it's a choice, but in this case it doesn't avoid any drilling? `useQRLoaderDetails` hook is called in both cases immediately on the next "depth" level. So if we would pass it as props we would have the exact same logic inside of QRDetails / PixDetails without any extra drilling since their children don't call the hook.

johnayeni: I agree as well. I would prefer to leave it as it is, but if you have any strong reasons why we should just move it to props I can do that as well. One other reason I went for this approach is to at some point break down the QR code component and Await component to reusable hooks and components.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -16,15 +15,17 @@ import useImage from '../../../core/Context/useImage';
...
 import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
+import { CountdownTime } from '../Countdown/types';
+import QRDetails from './components/QRDetails';
+import { QRLoaderDetailsProvider } from './QRLoaderDetailsProvider';
 
 const QRCODE_URL = 'utility/v1/barcode.png?type=qrCode&data=';
 
 class QRLoader extends Component<QRLoaderProps, QRLoaderState> {
-    private timeoutId;
+    private timeoutId: NodeJS.Timeout | undefined;
```
ribeiroguilherme: Minor thing - this is not wrong on NodeJS/server side environment, but on Web the type here should be of `number`. 
If you click in the reference of the `setTimeout` function, you get redirected to `lib.dom.d.ts` which shows that

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
m1aw: The only downside, is just the normal of Context API is that the dependencies of those props are unclear till runtime, but I do see why the aesthetic preferences. Just keep it like this.

packages/lib/src/components/helpers/QRLoaderContainer/QRLoaderContainer.tsx
```diff
@@ -41,10 +41,7 @@ class QRLoaderContainer<T extends QRLoaderConfiguration = QRLoaderConfiguration>
                         {...this.props}
                         type={this.constructor['type']}
                         brandLogo={this.props.brandLogo || this.icon}
-                        delay={this.props.delay}
                         onComplete={this.onComplete}
-                        countdownTime={this.props.countdownTime}
-                        instructions={this.props.instructions}
                         onActionHandled={this.onActionHandled}
```
ribeiroguilherme, line 48: How are these props (`delay`, `countdownTime`,  `instructions`) being propagated to the `QRLoader`? Are we now relying on the `{...this.props}`?

One of the issues we currently have in the repo is that there are props being spread in multiple places, and that leads to:
- unexpected/unused props ending in a Component
- it is not clear if/how the Component receives a config property (e.g. `delay`) 

For readability and to avoid propagating unnecessary objects, it is better to avoid spreading and pass the properties directly.

packages/lib/src/components/Pix/Pix.tsx
```diff
@@ -1,10 +1,14 @@
...
+import PixInput from './components/PixInput';
 import { cleanCPFCNPJ } from '../internal/SocialSecurityNumberBrazil/utils';
 import { PixElementData, PixConfiguration } from './types';
 import { TxVariants } from '../tx-variants';
+import QRLoader from '../internal/QRLoader';
+import SRPanelProvider from '../../core/Errors/SRPanelProvider';
+import PixQRDetails from './components/PixQRDetails';
+import './Pix.scss';
 
 class PixElement extends QRLoaderContainer<PixConfiguration> {
```
ribeiroguilherme, line 13: Does it still need to extend the `QRLoaderContainer` class? 🤔  
Maybe not .. but it is also not bad to do it since we can easily track which Components are QRLoader type..

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: Fixed 👍🏿

packages/lib/src/components/helpers/QRLoaderContainer/QRLoaderContainer.tsx
johnayeni: Indeed, I have also seen that we have quite a lot of unused props which can be quite confusing. I restored passing the `delay`, `countdownTime`, and `instructions` props explicitly for this PR but I also don't want to remove spreading the props as I am not very sure which of the props that are spread are not being used. Would have to dig a little bit deeper so I would do this as part of refactoring this component into a functional one.

packages/lib/src/components/Pix/Pix.tsx
johnayeni: Yup, I think when we make `QRLoader` functional, we might not need to extend this class at that point.

packages/lib/src/components/helpers/QRLoaderContainer/QRLoaderContainer.tsx
ribeiroguilherme: All good. My point was just to restore the previous ones, and we can investigate further the remaining props during the refactoring 👍

PR #3511, opened 2025-09-12
feat: add support for sdkData API field

m1aw:
## Summary

Add support for a new API field sdkData. Please check internal ticket for the full details. This PR adds sdkData as an encrypted blob in the `.data()`, this is used both on sessions and advanced flow payment requests.

## Tested scenarios

**Fixed issue**:  
COSDK-514

### DIFF
--- a/packages/lib/src/components/AfterPay/AfterPay.test.ts
+++ b/packages/lib/src/components/AfterPay/AfterPay.test.ts
@@ -142,7 +142,7 @@ describe('AfterPay', () => {
                             stateOrProvince: 'N/A',
                             street: 'Simon Carmilgestraat'
                         },
-                        paymentMethod: { checkoutAttemptId: 'fetch-checkoutAttemptId-failed', type: 'afterpay_default' },
+                        paymentMethod: { checkoutAttemptId: 'fetch-checkoutAttemptId-failed', sdkData: expect.any(String), type: 'afterpay_default' },
                         shopperEmail: 'jose@adyen.com',
                         shopperName: { firstName: 'Jose', lastName: 'Fernandez' },
                         telephoneNumber: '612345678'
@@ -325,7 +325,7 @@ describe('AfterPay', () => {
                             stateOrProvince: 'N/A',
                             street: 'Kinkerstraat'
                         },
-                        paymentMethod: { checkoutAttemptId: 'fetch-checkoutAttemptId-failed', type: 'afterpay_default' },
+                        paymentMethod: { checkoutAttemptId: 'fetch-checkoutAttemptId-failed', sdkData: expect.any(String), type: 'afterpay_default' },
                         shopperEmail: 'jose@adyen.com',
                         shopperName: { firstName: 'Jose', lastName: 'Fernandez' },
                         telephoneNumber: '612345678'
--- a/packages/lib/src/components/ApplePay/ApplePay.test.ts
+++ b/packages/lib/src/components/ApplePay/ApplePay.test.ts
@@ -644,6 +644,7 @@ describe('ApplePay', () => {
                 paymentMethod: {
                     applePayToken: 'InBheW1lbnQtZGF0YSI=',
                     checkoutAttemptId: NO_CHECKOUT_ATTEMPT_ID,
+                    sdkData: expect.any(String),
                     type: 'applepay'
                 }
             });
--- a/packages/lib/src/components/BacsDD/BacsDD.test.ts
+++ b/packages/lib/src/components/BacsDD/BacsDD.test.ts
@@ -49,6 +49,7 @@ describe('Bacs Direct Debit', () => {
                         bankLocationId: '560036',
                         checkoutAttemptId: 'fetch-checkoutAttemptId-failed',
                         holderName: 'David Archer',
+                        sdkData: expect.any(String),
                         type: 'directdebit_GB'
                     },
                     shopperEmail: 'jon@adyen.com',
@@ -167,6 +168,7 @@ describe('Bacs Direct Debit', () => {
                         bankLocationId: '895302',
                         checkoutAttemptId: 'fetch-checkoutAttemptId-failed',
                         holderName: 'Jonny R.',
+                        sdkData: expect.any(String),
                         type: 'directdebit_GB'
                     },
                     shopperEmail: 'jonnyr@adyen.com',
--- a/packages/lib/src/components/Boleto/Boleto.test.ts
+++ b/packages/lib/src/components/Boleto/Boleto.test.ts
@@ -74,7 +74,8 @@ describe('Boleto', () => {
                 data: {
                     paymentMethod: {
                         type: 'boletobancario',
-                        checkoutAttemptId: 'fetch-checkoutAttemptId-failed'
+                        checkoutAttemptId: 'fetch-checkoutAttemptId-failed',
+                        sdkData: expect.any(String)
                     },
                     billingAddress: {
                         city: 'Sao Paulo',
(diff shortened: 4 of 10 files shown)


packages/lib/src/utils/createSdkData.ts
```diff
@@ -0,0 +1,31 @@
...
+    };
+}
+
+/**
+ * Creates the sdkData object with analytics and risk information
+ * @param checkoutAttemptId - The checkout attempt ID from analytics
+ * @param clientData - The client data from risk module
+ * @returns Base64 encoded JSON string of the SDK data object
+ */
+export function createSdkData(checkoutAttemptId: string, clientData: string): string {
```
sponglord: According to the YT ticket this sdkData should also contain a timestamp?

The Analytics utils (`/src/core/Analytics/utils.ts`) contains a `getUTCTimestamp` function - which is what we use in analytics.

m1aw: Yeah, indeed. This has been added to the ticket, it wasn't there when I originally made the change. I'm going to be adding it, thanks for the tip.

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -87,15 +88,25 @@ abstract class BaseElement<P extends BaseElementProps> implements IBaseElement {
...
+        // first one used for the clientData field in the payment request
         const clientData = getProp(this.props, 'modules.risk.data');
+        // second one used for the sdkData field in the payment request
+        const clientDataUnencoded = getProp(this.props, 'modules.risk.dataUnencoded');
         const checkoutAttemptId = getProp(this.props, 'modules.analytics.getCheckoutAttemptId')?.() ?? NO_CHECKOUT_ATTEMPT_ID; // NOTE: we never expect to see this "failed" value, but, just in case...
         const order = this.state.order || this.props.order;
         const componentData = this.formatData();
 
+        // Create sdkData when both analytics and risk data are available
+        const sdkData = checkoutAttemptId && clientDataUnencoded ? createSdkData(checkoutAttemptId, clientDataUnencoded) : undefined;
```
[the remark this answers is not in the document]
m1aw: This is expected behaviour.

packages/lib/src/core/RiskModule/RiskModule.tsx
```diff
@@ -89,6 +89,14 @@ export default class RiskElement extends BaseElement<RiskModuleProps> {
         return false;
     }
 
+    get dataUnencoded(): object | false {
```
johnayeni: I am not a big fan here of returning either an object or false from the same function. We could make it more singular purpose by just retuning null as the negative case


```suggestion
    get dataUnencoded(): Record<string | any> | null {
```

packages/lib/src/core/RiskModule/RiskModule.tsx
```diff
@@ -89,6 +89,14 @@ export default class RiskElement extends BaseElement<RiskModuleProps> {
...
+        if (this.isValid) {
+            return { version: RISK_DATA_VERSION, ...this.state.data };
+        }
+
+        return false;
```
johnayeni:
```suggestion
        return null;
```

packages/lib/src/utils/createSdkData.ts
```diff
@@ -0,0 +1,33 @@
...
+    };
+}
+
+/**
+ * Creates the sdkData object with analytics and risk information
+ * @param checkoutAttemptId - The checkout attempt ID from analytics
+ * @param clientData - The client data from risk module
+ * @returns Base64 encoded JSON string of the SDK data object
+ */
+export function createSdkData(checkoutAttemptId: string, clientData: object): string {
```
johnayeni:
```suggestion
export function createSdkData(checkoutAttemptId: string, clientData: Record<string, any>): string {
```

Object is also not the most easy to work with in typescript. I propose using `Record<>` instead

packages/lib/src/core/RiskModule/RiskModule.tsx
sponglord: Is there a demand for this? i.e. the risk data in an "unencoded" form?
Or is this just for us - so we don't have to base64 decode the object?

m1aw: This is just for us, because we are going to encode it again when creating `sdkData` blob.

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -87,15 +88,25 @@ abstract class BaseElement<P extends BaseElementProps> implements IBaseElement {
...
+        // first one used for the clientData field in the payment request
         const clientData = getProp(this.props, 'modules.risk.data');
+        // second one used for the sdkData field in the payment request
+        const clientDataUnencoded = getProp(this.props, 'modules.risk.dataUnencoded') as Record<string, any> | null;
         const checkoutAttemptId = getProp(this.props, 'modules.analytics.getCheckoutAttemptId')?.() ?? NO_CHECKOUT_ATTEMPT_ID; // NOTE: we never expect to see this "failed" value, but, just in case...
         const order = this.state.order || this.props.order;
         const componentData = this.formatData();
 
+        // Create sdkData when both analytics and risk data are available
+        const sdkData = checkoutAttemptId && clientDataUnencoded ? createSdkData(checkoutAttemptId, clientDataUnencoded) : undefined;
```
sponglord: What happens if `clientDataUnencoded` doesn't exist e.g. if the merchant has disabled the riskModule?
We would still want to send an `sdkData` with the `checkoutAttemptId`

m1aw: That's a fair point. We would still want to send, I will change this part.

m1aw: I have made the change, together with a few other changes to the schema. It was a misunderstanding that we need to send the unencoded blob. Can you review again?

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -87,15 +88,22 @@ abstract class BaseElement<P extends BaseElementProps> implements IBaseElement {
      * Note: this does not ensure validity, check isValid first
      */
     public get data(): PaymentData {
+        // first one used for the clientData field in the payment request
         const clientData = getProp(this.props, 'modules.risk.data');
         const checkoutAttemptId = getProp(this.props, 'modules.analytics.getCheckoutAttemptId')?.() ?? NO_CHECKOUT_ATTEMPT_ID; // NOTE: we never expect to see this "failed" value, but, just in case...
         const order = this.state.order || this.props.order;
         const componentData = this.formatData();
 
+        const sdkData = createSdkData(checkoutAttemptId, clientData);
```
sponglord, line 97: So now you're no longer base64 decoding the risk data, correct?
You're just passing the encoded string and then it gets encoded a second time?

m1aw: We were never "decoding", we were just reading the plain object.
RiskModule encodes what the risk sdk gives us. Now we are just encoding it twice, one in the RiskModule and when creating sdkData object.

sponglord: And the double-decoding isn't a problem for the backend?
(I hope it's not - I do think this approach of just passing on what we've already created i.e. the encoded blob, is much simpler and cleaner)

m1aw: That was exactly the confusion that we had in the first place, backend wasn't sure about it because Android and iOS don't send it.

PR #3512, opened 2025-09-15
[wip] Adding analytics to Script

ribeiroguilherme:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/lib/src/utils/Script.ts
+++ b/packages/lib/src/utils/Script.ts
@@ -5,19 +5,13 @@ interface IScript {
     remove(): void;
 }
 
-/**
- * Creates a script element from a certain source in the passed node selector.
- * If no selector is passed it will add the script element in the body.
- *
- * @example
- * ```
- * const script = new Script('https://example.com/script.js', '.container');
- * script.load().then(doSomething);
- *
- * // To clean up just call the remove method
- * script.remove();
- * ```
- */
+interface IScriptProps {
+    src: string;
+    node?: string;
+    attributes?: Partial<HTMLScriptElement>;
+    dataAttributes?: Record<string, string | undefined>;
+}
+
 class Script implements IScript {
     private readonly src: string;
     private readonly node: string;
@@ -31,7 +25,7 @@ class Script implements IScript {
     public static readonly RETRY_DELAY = 1000;
     public static readonly MAX_NUMBER_OF_RETRIES = 3;
 
-    constructor(src: string, node = 'body', attributes: Partial<HTMLScriptElement> = {}, dataAttributes: Record<string, string | undefined> = {}) {
+    constructor({ src, node = 'body', attributes, dataAttributes }: IScriptProps) {
         this.src = src;
         this.node = node;
         this.attributes = attributes;
--- a/packages/lib/src/utils/Script.ts
+++ b/packages/lib/src/utils/Script.ts
@@ -1,4 +1,6 @@
 import AdyenCheckoutError from '../core/Errors/AdyenCheckoutError';
+import { AnalyticsInfoEvent, InfoEventType } from '../core/Analytics/AnalyticsInfoEvent';
+import { AnalyticsModule } from '../types/global-types';
 
 interface IScript {
     load(): Promise<void>;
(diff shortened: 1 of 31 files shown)


packages/lib/src/utils/Script.ts
```diff
@@ -1,28 +1,29 @@
...
- * script.load().then(doSomething);
- *
- * // To clean up just call the remove method
- * script.remove();
- * ```
- */
+interface IScriptProps {
+    src: string;
+    component: string;
+    // TODO: The optional flag will be fixed on PR #3464
```
ribeiroguilherme, line 13: This will be fixed on #3464

packages/lib/src/utils/Script.ts
```diff
@@ -48,21 +59,27 @@ class Script implements IScript {
             this.rejectLoadPromise = reject;
             let attempts = 0;
 
+            this.trackEvent(InfoEventType.sdkDownloadInitiated);
+
             const loadScriptWithRetry = async (): Promise<void> => {
                 try {
                     attempts++;
                     await this.loadScript();
```
johnayeni, line 67: [Question] Should this be an info event or an error event? It perharps depends on the use case we have for this event but I think it might be useful to have it as an error. If the SDK keeps failing to load, that should be an issue

packages/lib/src/utils/Script.ts
```diff
@@ -48,21 +59,27 @@ class Script implements IScript {
...
                     if (this.loadPromise === null) {
                         return;
                     }
 
+                    this.trackEvent(InfoEventType.sdkDownloadFailed);
+
                     this.removeScript();
 
                     if (attempts < Script.MAX_NUMBER_OF_RETRIES) {
                         setTimeout(() => void loadScriptWithRetry(), Script.RETRY_DELAY);
```
johnayeni, line 80: [Suggestion] Not in this scope of this PR but maybe we could consider an exponential backoff for these retries. They can be useful for making more efficient and successful retries https://en.wikipedia.org/wiki/Exponential_backoff

packages/lib/src/components/internal/ClickToPay/services/sdks/VisaSdk.test.ts
```diff
@@ -34,28 +36,28 @@ afterEach(() => {
 
 describe('SDK urls', () => {
     test('should load sdk script with correct URL for live', async () => {
-        const sdk = new VisaSdk('live', { dpaLocale: 'en-US', dpaPresentationName: 'MyStore' });
+        const sdk = new VisaSdk('live', { dpaLocale: 'en-US', dpaPresentationName: 'MyStore' }, mockAnalytics);
```
m1aw: For all this tests that take the same props should we create and reuse a constant and the spread/override in each test?

packages/lib/src/utils/Script.ts
```diff
@@ -31,11 +37,16 @@ class Script implements IScript {
...
+    constructor({ src, component, node = 'body', attributes, dataAttributes, analytics }: IScriptProps) {
         this.src = src;
+        this.component = component;
         this.node = node;
         this.attributes = attributes;
         this.dataAttributes = dataAttributes;
+        this.analytics = analytics;
+
+        const url = new URL(this.src);
+        this.baseUrl = url.origin + url.pathname;
```
m1aw: It's not obvious what this is doing. Is this to remove query params based on the comment above, but maybe explain the code here so it's more obvious.

packages/lib/src/utils/Script.ts
ribeiroguilherme: The definition of the Error event is:

> Error events warn us whenever the SDK is broken and action is required from our side to fix the Checkout SDK or API.

In this case, there is nothing 'broken' on our side. We are just reporting that a third party SDK failed to load even after the retrials. We can use this data to follow up with the payment method provider, but there is nothing else we can do to mitigate the problem!

packages/lib/src/utils/Script.ts
ribeiroguilherme: Good one!
I discussed that with @longyulongyu when I implemented this, and we decided that it wasn't needed in this case. Perhaps can be added in the next iteration if we still face issues.

packages/lib/src/utils/Script.ts
ribeiroguilherme: I updated the code

packages/lib/src/components/internal/ClickToPay/services/sdks/VisaSdk.test.ts
ribeiroguilherme: I updated the code

PR #3529, opened 2025-09-22
PayTo: Allow leading zeros for phone number field

johnayeni:
## Summary

Allow leading zeros for phone number field of `PayTo` component

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @sponglord @m1aw @ribeiroguilherme @longyulongyu @johnayeni
+* @adyen/checkout-web


packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -10,8 +10,6 @@ import usePhonePrefixes from '../../../internal/PhoneInput/usePhonePrefixes';
 function MBWayInput(props: MBWayInputProps) {
     const { i18n, loadingContext } = useCoreContext();
 
-    const phoneInputRef = useRef(null);
```
johnayeni, line 13: I removed this since the phone input form was not using the ref anyway

packages/lib/src/components/PayTo/components/PayIDInput.tsx
```diff
@@ -85,9 +85,7 @@ export default function PayIDInput({ setComponentRef, defaultData, placeholders,
                 onSelectedIdentifier={handleChangeFor('selectedIdentifier')}
                 selectedIdentifier={data.selectedIdentifier}
             />
-            {data.selectedIdentifier === PayToIdentifierEnum.phone && (
-                <PayToPhone onChange={handleChangeFor('phone', 'blur')} data={data} form={form} />
-            )}
+            {data.selectedIdentifier === PayToIdentifierEnum.phone && <PayToPhone form={form} />}
```
m1aw, line 88: Good catch on the `onChange` not being needed there getting from the `form` in the component.

johnayeni:
> Just to confirm if I understand the whole change we are doing here.
> 
> Most of it is just clean up right? Besides that only logic changes are the Regex + Test?

Yes, indeed. The other changes are just small cleanups

m1aw (approved):

PR #3535, opened 2025-09-26
fix: remove if clause to allow E2E on renovate

m1aw:
## Summary
It seems like this line is the culprit of E2E not running on renovate. However, I don't understand how on the same day a renovate PR has been merged successfully: #3534 

## Tested scenarios

**Fixed issue**:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,7 +6,6 @@ on:
 
 jobs:
   build-then-e2e:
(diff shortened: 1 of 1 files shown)


atmamont:
I've seen some comms it [will be renamed](https://github.com/renovatebot/renovate/discussions/37842), maybe this is the cause

m1aw:
Interesting. I still don't see how this was working before, but the timeline matches.

ribeiroguilherme (approved):

PR #3543, opened 2025-09-30
feat(bankTransfer): support new countries and new fields

m1aw:
## Summary

This adds support for new `bankTransfer` variants and also support for 2 new fields `bankCode` and `branchCode`.

## Tested scenarios

**Fixed issue**:


packages/lib/src/components/BankTransfer/BankTransfer.test.ts
```diff
@@ -79,5 +79,70 @@ describe('BankTransfer', () => {
...
+            });
+
+            render(bankTransfer.render());
+            expect(await screen.findByText('A$100.00')).toBeInTheDocument();
+            expect(await screen.findByText('Beneficiary name')).toBeInTheDocument();
+            expect(await screen.findByText(mockResult.beneficiary)).toBeInTheDocument();
+            expect(await screen.findByText('Account number')).toBeInTheDocument();
+            expect(await screen.findByText(mockResult.accountNumber)).toBeInTheDocument();
+            expect(await screen.findByText('Branch code')).toBeInTheDocument();
+            expect(await screen.findByText(mockResult.branchCode)).toBeInTheDocument();
```
johnayeni, line 109: To test for accessibility better here we should find the value based on the label i.e

```ts
expect(screen.getByLabelText('Branch code').innerText).toBe(mockResult.branchCode)
```

m1aw: You mean to make sure it's a semantic label? I'm going to test, but I think is a limitation of `testing-library` and it not considering that definition lists are `labels`, but my memory could be wrong. Honestly just copy and pasted from the code above.

m1aw: Yeah, it's not a label so what we do for example in PayTo.test.ts:L303 is:
```
const mandateAmount = screen.getByText('Amount').nextSibling;
expect(mandateAmount).toHaveTextContent('Up to A$40.01 per transaction');
```

m1aw: I think this also might be related: https://github.com/testing-library/dom-testing-library/issues/140

johnayeni: Two options I would like to propose. So I see that indeed, the bank instructions table component does not tie the label to the fields. This is actually not great accessibility wise.

Could we take this opportunity to either
- Refactor the markup to use a grid with each label actually using a label element with an id to link it to its value. One drawback here is that we might break backwards compatibility with the styles perhaps we can keep the classnames and still do this
- Simply just use `aria-labeledby` and `id` props with the current markup. Then we can use the `getByLabelText ` tests

What do you think?

m1aw: Well, I don't agree that is not great a11y wise. We don't need implicit association for definition lists. This is more of a problem with testing-library that refuse to add support for it.

These are 2 blogs that dive quite a lot into this topic:
- https://www.oidaisdes.org/description-list-html-element.en/
- https://adrianroselli.com/2025/01/updated-brief-note-on-description-list-support.html

There's some caveats to this implementation, which is not perfect on iOS, but semantically speaking they are the correct option in this case. For now I don't think we should change, specially since we are getting the results of the audit this Wednesday 8th, we can change depending on their findings/advice.

johnayeni: Got it. Thanks for explaining. This is fine for now then 👍🏿

johnayeni (approved):
🚀

PR #3546, opened 2025-10-01
Upgrade to yarn 4

johnayeni:
## Summary
This PR upgrades the yarn package manager version we use for this project from version 1 to version 4. This provides a number of [benefits](https://yarnpkg.com/migration/overview) especially when it comes to installs and our install time for packages seems to be reduced by an average of about 10 seconds.
As part of this upgrade we are now going to also be using [corepack](https://nodejs.org/download/release/v19.9.0/docs/api/corepack.html) which can help us better manage package managers if we decide to switch to a different one.

Some other improvements have been included in this PR as well
- Caching nodule modules to skip the install step for installing npm packages if no new package have been added
- Caching playwright binaries for chromium and firefox browsers to allow skipping the installation for e2e tests
- Only install the necessary browser dependency for each CI e2e job

## CI perfomance benchmarking

| Yarn 4 install   | Yarn 1 install |
| -------- | ------- |
| ~26s  | ~35s    |

| E2E Chromium, API v71 (with cache)   | E2E Chromium, API v71 (without cache) |
| -------- | ------- |
| 8m 7s  | 9m 36s    |

| E2E Firefox, API v71 (with cache)   | E2E Firefox, API v71 (without cache) |
| -------- | ------- |
| 7m 35s  | 10m 56s   |

## Tested scenarios

- [x] Local development and builds
- [x] CI build
- [x] Unit and e2e tests

**Fixed issue**:

### DIFF
--- a/.changeset/famous-cows-hug.md
+++ b/.changeset/famous-cows-hug.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fix accessibility issue with opening klarna widget with enter button
--- a/.changeset/five-pears-fold.md
+++ b/.changeset/five-pears-fold.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+Apple Pay - Adding 'domainName' configuration property which can be used to set which domain is performing the merchant validation request (Useful when embedding component within iframe)
--- a/packages/lib/src/components/ApplePay/ApplePay.test.ts
+++ b/packages/lib/src/components/ApplePay/ApplePay.test.ts
@@ -203,8 +203,7 @@ describe('ApplePay', () => {
             await new Promise(process.nextTick);
             expect(jest.spyOn(ApplePayService.prototype, 'begin')).toHaveBeenCalledTimes(1);
 
-            // Trigger ApplePayService onValidateMerchant property
-            // @ts-ignore ApplePayService is mocked
+            // @ts-ignore  ApplePayService is mocked. Trigger ApplePayService onValidateMerchant property
             const onValidateMerchant = ApplePayService.mock.calls[0][1].onValidateMerchant;
             const resolveMock = jest.fn();
             const rejectMock = jest.fn();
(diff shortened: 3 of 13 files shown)


.github/workflows/automated-a11y.yml
```diff
@@ -32,16 +32,47 @@ jobs:
...
+    - name: Cache node modules
+      id: node-modules-cache
+      uses: actions/cache@v4
+      env:
+        cache-name: node-modules-v1
+      with:
+        path: node_modules
+        key: ${{ runner.os }}-cache-${{ env.cache-name }}-node-${{ matrix.node-version }}-yarn-${{ hashFiles('**/yarn.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-cache-${{ env.cache-name }}-node-${{ matrix.node-version }}-yarn-
```
m1aw: Do we want the restore-keys? If there's no match on the hash for `yarn.lock` should we not just rebuild?

.github/workflows/automated-a11y.yml
```diff
@@ -32,16 +32,47 @@ jobs:
...
-    
+      if: steps.node-modules-cache.outputs.cache-hit != 'true'
+      run: yarn install --immutable
+
+    - name: Extract Playwright version
+      id: playwright-version
+      run: echo "PLAYWRIGHT_VERSION=$(jq -r '.devDependencies["@playwright/test"]' packages/e2e-playwright/package.json)" >> $GITHUB_ENV
+
+    - name: Cache Playwright
+      # Using a cache for the webkit binary files causes this error on subsequent runs
```
m1aw, line 57: Also what is the difference in performance we get here, can you make the comparison? 

https://playwright.dev/docs/ci#caching-browsers

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
m1aw, line 1: Also would it be beneficial to change it to pnp or pnpm for example?

.github/workflows/automated-a11y.yml
johnayeni: Indeed, there is no need for this. I'll remove this.

.github/workflows/automated-a11y.yml
johnayeni: | E2E Chromium, API v71 (with cache)   | E2E Chromium, API v71 (without cache) |
| -------- | ------- |
| 8m 7s  | 9m 36s    |

| E2E Firefox, API v71 (with cache)   | E2E Firefox, API v71 (without cache) |
| -------- | ------- |
| 7m 35s  | 10m 56s   |

We get about a 1 to 2 minute reduction in time it takes for the tests on Firefox and Chromium which are currently the only tests which cached playwright browsers.

.yarnrc.yml
johnayeni: We could but there few things still to consider. One major one is local development. Since we do not have node modules anymore, some code editors have issues with resolving the source path. This was with `pnp` however so I would experiment with `pnpm`

.github/workflows/automated-a11y.yml
m1aw: I meant more the difference for this exact step. But I went to check this on other PRs and this job took about ~58s on Chromium and now is about ~4s. 

It gets my worried about the OS dependencies but if nothing is broken let's not fix it.

m1aw (approved):

PR #3560, opened 2025-10-03
Klarna: Fix accessibility issue with opening klarna widget

johnayeni:
## Summary

-  Fix accessibility issue with opening klarna widget
- Fix type for `Button` component `onClick` prop

## Tested scenarios

- [x] Manual test

**Fixed issue**:

### DIFF
--- a/packages/lib/src/components/internal/Button/Button.tsx
+++ b/packages/lib/src/components/internal/Button/Button.tsx
@@ -12,20 +12,14 @@ class Button extends Component<ButtonProps, ButtonState> {
         disabled: false,
         label: '',
         inline: false,
-        target: '_self',
-        onClick: () => {},
-        onMouseEnter: () => {},
-        onMouseLeave: () => {},
-        onFocus: () => {},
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/internal/Button/Button.tsx
```diff
@@ -12,20 +12,14 @@ class Button extends Component<ButtonProps, ButtonState> {
         disabled: false,
         label: '',
         inline: false,
-        target: '_self',
-        onClick: () => {},
-        onMouseEnter: () => {},
-        onMouseLeave: () => {},
-        onFocus: () => {},
-        onBlur: () => {},
-        onKeyPress: () => {}
```
johnayeni, line 21: These default handlers were simply just adding event listeners to the button element which did nothing

e.g 

Before
Screenshot 2025-10-03 at 15 36 22

Now
Screenshot 2025-10-03 at 15 35 40

ribeiroguilherme (approved):

Issue #3408, opened 2025-07-28
ApplePay instant payment method is not show and selected/choosen


jureperak:
**Describe the bug**
When using Apple Pay as an instant payment method and dismissing the Apple Pay payment sheet, the onError callback shows the wrong payment method as selected/clicked.

**To Reproduce**
1. Go to https://www.mystoredemo.io/
2. Select products and proceed to checkout
3. Open the developer console
4. Verify that PayPal is selected by default (leave it selected)
5. Click on Apple Pay
6. Close the Apple Pay modal
7. Check the developer console log (as shown in the screenshot), which will contain the data.paymentMethod object

**Expected behavior**
Console.log should show `applepay` as payment method. 
If we do the same for GooglePay, it will show `paywithgoogle` in `data.paymentMethod` (screenshot attached)

**Screenshots**

<img width="1335" height="1072" alt="Image" src="https://github.com/user-attachments/assets/d14699ba-9f5d-4c64-a52c-dc5d3d82dad7" />

<img width="406" height="837" alt="Image" src="https://github.com/user-attachments/assets/9773e78b-9c07-41f4-b121-4b886bf02726" />

<img width="364" height="761" alt="Image" src="https://github.com/user-attachments/assets/6dc38a0d-a719-45eb-bf3e-459a09e0022e" />

**Additional context**
I know that demo shop is on old version, but same happens on 6.18 version

ribeiroguilherme:
Hey @jureperak ,

Thanks for flagging this issue. The team took a look into it and indeed it is not working as expected. We will look into fixing it on our side.

Cheers

jureperak:
Do you have a rough estimate on when the fix could be available? No rush, just checking.

ribeiroguilherme:
Hi @jureperak  - we don't have an estimate yet unfortunately

dimitri-dev:
@ribeiroguilherme was this fixed as part of 6.24.0?

ribeiroguilherme:
@dimitri-dev yes. I forgot to close this thread.  Cheers

PR #3581, opened 2025-10-09
Apple Pay - Fix issue where payment is not selected if used as Instant payment

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

The click event on the Apple Pay button was not being propagated to the `InstantPaymentMethods` `onClick` handler. I suspect that the Apple Pay button stops the propagation, whereas the Google Pay button works as expected.

This PR forces the capture step when clicking on the button, so we can set the Apple Pay as the active payment method when using it as Instant Payment within Drop-in

---

## 🧪 Tested scenarios

---

**Closes**:  #3408 

---


packages/lib/src/components/Dropin/components/PaymentMethod/InstantPaymentMethods.tsx
```diff
@@ -11,8 +11,7 @@ function InstantPaymentMethods({ paymentMethods, onSelect }: InstantPaymentMetho
     return (
         <ul className="adyen-checkout__instant-payment-methods-list">
             {paymentMethods.map(pm => (
-                // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
-                <li key={pm._id} data-testid={pm.type} onClick={() => onSelect(pm)}>
+                <li key={pm._id} data-testid={pm.type} onClickCapture={() => onSelect(pm)}>
```
longyulongyu, line 18: Maybe put a comment why we need to use `onClickCapture` instead of `onClick`.

PR #3583, opened 2025-10-10
feat: added responsive behavior on input fields for card and stored card

longyulongyu:
## 📋 Pull Request Checklist

---

## 📝 Summary
This PR makes the expiry date and security code fields stack vertically on smaller screens, ensuring a consistent, responsive layout that matches other payment methods.

---

## 🧪 Tested scenarios

https://github.com/user-attachments/assets/3d188eb8-55e2-40f4-a436-3848cfcca6c0

---

**Closes**:  COSDK-270

---

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
@@ -61,7 +61,7 @@ export default function CardFields({
             <AvailableBrands activeBrand={brand} brands={allowedBrands} />
 
             <div
-                className={classNames('adyen-checkout__card__exp-cvc adyen-checkout__fieldset__fields', {
+                className={classNames('adyen-checkout__card__exp-cvc', {
                     'adyen-checkout__card__exp-cvc__exp-date__input--hidden': expiryDatePolicy === DATE_POLICY_HIDDEN
                 })}
             >
--- a/packages/lib/src/components/Card/components/CardInput/components/CardFields.scss
+++ b/packages/lib/src/components/Card/components/CardInput/components/CardFields.scss
@@ -1,7 +1,11 @@
-@import 'styles/variable-generator';
+@use 'styles/index';
 
 .adyen-checkout__card__exp-cvc {
-    .adyen-checkout__field {
-        margin-bottom: 0;
+    @include index.fieldset-fields-layout;
+
+    @include index.screen-s-and-up {
+        .adyen-checkout__field {
+            margin-bottom: 0;
+        }
     }
 }
--- a/packages/lib/src/components/internal/FormFields/Fieldset/Fieldset.scss
+++ b/packages/lib/src/components/internal/FormFields/Fieldset/Fieldset.scss
@@ -47,14 +47,7 @@
 }
 
 .adyen-checkout__fieldset__fields {
-    display: flex;
-    flex-wrap: wrap;
-    justify-content: space-between;
-    width: 100%;
-
-    > .adyen-checkout__field:last-of-type {
-        margin-bottom: 0;
-    }
+    @include index.fieldset-fields-layout;
 }
 
 .adyen-checkout__fieldset--readonly .adyen-checkout__fieldset__fields {
--- a/packages/lib/src/styles/mixins.scss
+++ b/packages/lib/src/styles/mixins.scss
@@ -142,7 +142,7 @@ $adyen-checkout-media-query-l-min: 1024px;
 }
 
 @mixin adyen-checkout-text-caption-stronger {
-   @include adyen-checkout-text-caption;
+    @include adyen-checkout-text-caption;
 
     font-weight: token(text-caption-stronger-font-weight);
 }
@@ -173,4 +173,15 @@ $adyen-checkout-media-query-l-min: 1024px;
 
 @mixin adyen-checkout-component-loading {
     pointer-events: none;
+}
+
+@mixin fieldset-fields-layout {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: space-between;
+    width: 100%;
+
+    > .adyen-checkout__field:last-of-type {
+        margin-bottom: 0;
+    }
 }


packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
```diff
@@ -60,33 +60,33 @@ export default function CardFields({
             <AvailableBrands activeBrand={brand} brands={allowedBrands} />
 
             <div
-                className={classNames('adyen-checkout__card__exp-cvc adyen-checkout__field-wrapper', {
+                className={classNames('adyen-checkout__card__exp-cvc adyen-checkout__fieldset__fields', {
```
johnayeni: Just a question here,  but should we use the `Fieldset` component here instead? the reason this class works is because the `Fieldset` component is imported somewhere in the `Card` component but it is not explicit from this component and `adyen-checkout__fieldset__fields` is not a global classname

packages/lib/src/components/Card/components/CardInput/components/StoredCardFields.tsx
```diff
@@ -34,12 +34,11 @@ export default function StoredCardFields({
 
     return (
         <div className="adyen-checkout__card__form adyen-checkout__card__form--oneClick" aria-label={ariaLabel}>
-            <div className="adyen-checkout__card__exp-cvc adyen-checkout__field-wrapper">
+            <div className="adyen-checkout__card__exp-cvc adyen-checkout__fieldset__fields">
```
johnayeni: Same comment as here https://github.com/Adyen/adyen-web/pull/3583/files#r2425684207

packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
longyulongyu: That's a good point, I'll take a look.

longyulongyu: So I was checking this, I do think we should use `Fieldset` for better semantic dom, but replacing the div with the `Fieldset` would break for users who have customized the layout using the `.adyen-checkout__card__exp-cvc` class. 

E.g: if they've applied something like below, it won't work anymore. Because the `Fieldset` component renders its own internal structure - `<div class="adyen-checkout__fieldset__fields">`, now the layout will be controlled by `adyen-checkout__fieldset__fields` class instead of `.adyen-checkout__card__exp-cvc`: 
```css
.adyen-checkout__card__exp-cvc {
    display: gird
}
```
Besides this, i also need to pass the class `.adyen-checkout__card__exp-cvc` to `Fieldset` as we cannot remove it.

Let me know your thoughts.

longyulongyu:
@johnayeni @m1aw i've rebased from `main`, the e2e tests should pass now, and PR is ready for review. 
Can you both take a look at this [comment](https://github.com/Adyen/adyen-web/pull/3583#discussion_r2428281376) and let me know what you think?

packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
johnayeni: If we can pass the `adyen-checkout__card__exp-cvc` class to the `Fieldset` component, that would be great. I like the idea. If this would however breaksomething else then I think we can proceed as it is

longyulongyu: I think it's possible and non breaking, i've updated my code.

I've refactored some css, maybe @m1aw you can do a sanity check on [this commit](https://github.com/Adyen/adyen-web/pull/3583/commits/f44e51992efd3a429cad7d2bb40cc4bda68fbfd2)?

The reason is the following css rule has a very high specificity, which makes it difficult to override, and changes the card number field margin to 0 for my PR when i try to use `Fieldset` component. 

I took a look and it seems `.adyen-checkout__field-group` is only used in the `Address` component. Hence i've moved some css there. 

On top of that, i think the css meant to set bottom margin to 0 to the last `.adyen-checkout__field` element that is a direct child of a non `.adyen-checkout__field-group` element, i think it can be replaced by [this](https://github.com/Adyen/adyen-web/blob/f44e51992efd3a429cad7d2bb40cc4bda68fbfd2/packages/lib/src/components/internal/FormFields/Fieldset/Fieldset.scss#L55).

```css
:not(.adyen-checkout__field-group) > .adyen-checkout__field:last-of-type {
    margin-bottom: 0;
}
```

packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
```diff
@@ -59,10 +60,13 @@ export default function CardFields({
 
             <AvailableBrands activeBrand={brand} brands={allowedBrands} />
 
-            <div
-                className={classNames('adyen-checkout__card__exp-cvc adyen-checkout__fieldset__fields', {
-                    'adyen-checkout__card__exp-cvc__exp-date__input--hidden': expiryDatePolicy === DATE_POLICY_HIDDEN
-                })}
+            <Fieldset
```
m1aw: Perhaps I'm missing something from the other discussions, and a bit late to the topic.
I don't think this should be a fieldset, there is no semantic meaning in this div. A fieldset has semantic meaning and this purely a styling div.
The fieldset for Card is the whole form.
@longyulongyu @johnayeni

packages/lib/src/components/Card/components/CardInput/components/StoredCardFields.tsx
```diff
@@ -34,7 +35,7 @@ export default function StoredCardFields({
 
     return (
         <div className="adyen-checkout__card__form adyen-checkout__card__form--oneClick" aria-label={ariaLabel}>
-            <div className="adyen-checkout__card__exp-cvc adyen-checkout__fieldset__fields">
+            <Fieldset classNamesFields={['adyen-checkout__card__exp-cvc']}>
```
m1aw, line 38: However I do believe this one makes sense.

packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
longyulongyu: Hmm that actually makes sense, it presents the payment details as a whole semantic section. I think i will roll this change back🙃

johnayeni: In that case, should we just add new styles for the card fields instead of reusing the fieldset class?

longyulongyu: Would creating a mixin and reusing it in both places be a better and more maintainable approach?

johnayeni: That would actually be a good idea. I think we can create mixins for things we want to be able to reuse

e.g flex box styles, full width.

I think this can be a follow up though. For this PR, I am fine with us proceeding as it is if we can pick it up later. If a mixin can be added quickly now that would be great.

longyulongyu: It seems not a big change. I pushed it to this PR.

johnayeni (approved):
🚀

PR #3589, opened 2025-10-17
Fixing issue with actions in onAdditionalDetails when initiated from createFromAction. FINAL

sponglord:
## 📋 Pull Request Checklist

---

### Description:
Some payment actions that lead to a call to `onAdditionalDetails`, when initiated via `checkout.createFromAction`, were seeing `onAdditionalDetails` being called without the `actions` argument containing the `resolve` & `reject` functions. These are needed to complete the flow, as detailed in our documentation.

This applies to `threeDS`, `await` & `qrCode` actions. 

Calling `checkout.createFromAction` means the component that is instantiated just receives the raw, merchant defined, `onAdditionalDetails` callback, and not the "wrapped" version of this callback that is generated by `UIElement.handleAction`. 

It is this latter, wrapped, version, called within ` UIElement.handleAdditionalDetails`, that creates a `Promise` that can then be finalised with the `resolve` & `reject` functions that are passed, as the third argument, to the merchant's own `onAdditionalDetails` callback.
Our documentation states that the merchant callback should expect to receive these 2 functions and that they should be called to properly finalise the flow & lead to `onPaymentCompleted` or `onPaymentFailed` being called.

### Fix:
This PR fixes this problem by:
- `UIElement.onComplete` _only_ calling `this.handleAdditionalDetails`, with no additional checks.
- Changing what we map when we handle an action in `actionTypes.ts`. 
    - Most actions do not need an `onComplete` function mapped for them now. 
    - The affected action components (`Await` & `QRCode` based components) naturally call their `onComplete` function (inherited from `UIElement`). 
    - It is only the 3DS2 action that _might_  need to call an alternative `onComplete` function  (for the case of the MDFlow)
- `ThreeDS2DeviceFingerprint` and `ThreeDS2Challenge` then, overwrite their inherited `onComplete` function, so they can check for the presence of `this.props.onComplete`. If this exists they call it in preference to `super.onComplete`

## Tested scenarios
- Manually tested `threeDS`, `await` (via `MBWay`) & `qrCode` (via `PromptPay`) actions, both with `handleAction` & `createFromAction`, to see them complete correctly in both scenarios.

- Also tested `threeDS` in the "MDFlow", to see that the `onComplete` function passed by _this_ flow is called _instead_ of any others.

- Tested the `redirect` action to see that this situation didn't arise (`checkout.submitDetails`, which is called after a redirect, creates its own `Promise`, with its own `resolve` & `reject` functions that can then be use to finalise the payment)

- Tested the `sdk` action with the Klarna widget to see that the `onComplete` for the `KlarnaPayments` component directly calls it's `super.handleAdditionalDetails` function. So this issue doesn't arise.

- Tested that this doesn't apply to `bankTransfer` actions. (They are concluded in a 3rd party app and not via `onAdditionalDetails`)

- Tested that this doesn't apply to `voucher` actions, since they don't conclude in a call to `onAdditionalDetails`

- Also tested how a card and storedCard within the Dropin performed, to ensure the `elementRef` (which points to the `DropinElement`) is persisted through the 3DS2 flow and reaches the `ThreeDS2Challenge` component.

- Tested that both `await` & `qrCode` components also performed correctly within the Dropin

- All e2e tests pass
- Additional unit tests added

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
+++ b/packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
@@ -56,7 +56,7 @@ const actionTypes = {
             token: action.token,
             paymentData,
             onActionHandled: props.onActionHandled,
-            onComplete: props.onComplete, // NOTE: We only expect props.onComplete to be defined for the MDFlow
+            onComplete: props.onComplete, // NOTE: We only ever expect props.onComplete to be defined for the MDFlow
             onError: props.onError,
             isDropin: !!props.isDropin,
             loadingContext: props.loadingContext,


packages/playground/src/pages/ThreeDS/ThreeDS.js
```diff
@@ -1,23 +1,70 @@
...
 (async () => {
     const checkout = await AdyenCheckout({
         countryCode,
         locale: shopperLocale,
         environment: 'test',
-        clientKey: process.env.__CLIENT_KEY__,  
-
+        clientKey: process.env.__CLIENT_KEY__,
         ...environmentUrlsOverride,
+        onSubmit: async (state, component, actions) => {
```
longyulongyu, line 16: Can we move this to storybook? Switching between different playgrounds is a bit inconvenient.

Maybe worth to create a similar story for QR code as well.

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -146,6 +146,14 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
         return this;
     }
 
+    /**
+     * Function that can be called for core.ts - forms part of the "contract" between core and UIElement
+     * @param state -
+     */
+    public _internalHandleAdditionalDetails(state: AdditionalDetailsData): void {
```
longyulongyu: I'm not really a fan of this suggestion from Gemini. It feels like a sugar-coated workaround to pass the TypeScript check, and it compromises the encapsulation of handleAdditionalDetails as a protected function.

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -377,8 +385,7 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
 
         const paymentAction = this.core.createFromAction(action, {
             ...this.elementRef.props,
-            ...props,
-            onAdditionalDetails: this.handleAdditionalDetails
+            ...props
```
longyulongyu, line 380: Why we need to remove this?

packages/lib/src/core/core.ts
```diff
@@ -264,7 +264,15 @@ class Core implements ICore {
 
             const props = {
                 ...this.getCorePropsForComponent(),
-                ...options
+                onComplete: (state: AdditionalDetailsData, component?: UIElement) => {
```
longyulongyu: I like the clear mapping of `onComplete: props.onComplete` in the `actionTypes.ts` file. 
Since we map it correctly now, I am wondering maybe we can move this piece of logic inside UIElement? 
(just an idea, not fully tested)
```js
    protected onComplete(state): void {
        if (this.props.onComplete) {
            this.props.onComplete(state, this.elementRef);
        } else {
            this.handleAdditionalDetails(state);
        }
    }
```
Doing so, we can also get rid of the TS warning and no need to create `_internalHandleAdditionalDetails`

packages/playground/src/pages/ThreeDS/ThreeDS.js
sponglord: Is this a blocker?

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: But that was exactly gemini's point that doing it the other way broke encapsulation!
And, tbh, I thought the the other way was incredibly unreadable (and ugly) - all just to satisfy the TypeScript checker

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: It used to be there so it could get passed on to `actionTypes`, but it's no longer necessary since we wrap it in the `onComplete` function we generate in `core`

packages/lib/src/core/core.ts
sponglord: Which piece of logic are you talking about moving?

sponglord: yeah, could work. I'll take a look

packages/playground/src/pages/ThreeDS/ThreeDS.js
longyulongyu: It's not a blocker, but it shouldn't take too long to move this 3ds example into Storybook.

Creating similar stories for the QR code can be handled in a separate PR.

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: And with the latest batch of changes it's not required because an action will always hit `UIElement.onComplete` where we make the decision whether to call `this.handleAdditionalDetails` directly

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: This function, or the ugly type code, no longer exists

packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
```diff
@@ -20,7 +20,7 @@ const getActionHandler = statusType => {
         const config = {
             ...props,
             ...action,
-            onComplete: props.onAdditionalDetails,
+            onComplete: props.onComplete ? props.onComplete : undefined,
```
longyulongyu: I think the ternary operator is redundant because it effectively does what the JS engine already does.
If `props.onComplete` is not defined on the props object, accessing it evaluates to `undefined` by default.

packages/lib/src/core/ProcessResponse/PaymentAction/actionTypes.ts
```diff
@@ -57,7 +57,9 @@ const actionTypes = {
             token: action.token,
             paymentData,
             onActionHandled: props.onActionHandled,
-            onComplete: props.isMDFlow ? props.onComplete : props.onAdditionalDetails,
+            // Be clear about the fact that we only define an onComplete handler if one has been explicitly passed to the action (which in this case
+            // we only expect to happen for the MDFlow)
+            onComplete: props.onComplete ? props.onComplete : undefined,
```
longyulongyu: same here

packages/lib/src/components/internal/UIElement/UIElement.test.ts
```diff
@@ -66,17 +66,6 @@ describe('UIElement', () => {
         });
     });
 
-    describe('onComplete()', () => {
```
longyulongyu, line 69: With the new change:
```
   protected onComplete(state): void {
        this.handleAdditionalDetails(state);
    }
```
This unit test is no longer needed. 
@ribeiroguilherme tagging you for a quick sanity check in case we’ve missed anything.

sponglord: We're basically saying that, in the normal course of events,  we don't expect a `UIElement` to be instantiated with an `onComplete` prop...

packages/playground/src/pages/ThreeDS/ThreeDS.js
longyulongyu: We can resolve this for now. 
As agreed, it will be moved to storybook in a separate PR.

PR #3591, opened 2025-10-20
Refactor  QRLoader to functional component

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

- Refactor `QRLoader` loader component from class component to functional component
- Add missing types to functions in `Await` component
- Remove unused `ref` prop from `Await` and `QRLoader` component

---

## 🧪 Tested scenarios

- [x] Manual testing
- [x] E2e tests
- [x] Unit tests 

---

**Closes**:  

---

### DIFF
--- a/.github/workflows/.size-limit.yml
+++ b/.github/workflows/.size-limit.yml
@@ -8,7 +8,7 @@ jobs:
 
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Use Node.js 22.18.0
         uses: actions/setup-node@v4
--- a/.github/workflows/automated-a11y.yml
+++ b/.github/workflows/automated-a11y.yml
@@ -26,7 +26,7 @@ jobs:
       API_VERSION: ${{matrix.api-version}}
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
     
     - name: Setup Node
       uses: actions/setup-node@v4
--- a/.github/workflows/check-release-build.yml
+++ b/.github/workflows/check-release-build.yml
@@ -13,7 +13,7 @@ jobs:
       trigger_release: ${{ steps.check_commit_message.outputs.trigger_release }}
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Get commit message
         id: get_commit_message
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v2
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -26,7 +26,7 @@ jobs:
       API_VERSION: ${{matrix.api-version}}
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Setup Node
       uses: actions/setup-node@v4
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -8,7 +8,7 @@ jobs:
   publish-npm:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       # Setup .npmrc file to publish to npm
       - uses: actions/setup-node@v4
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -11,7 +11,7 @@ jobs:
   build-umd-and-translations:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       # Setup .npmrc file to publish to npm
       - uses: actions/setup-node@v4
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup Node.js 22.18.0
         uses: actions/setup-node@v4
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -16,7 +16,7 @@ jobs:
         node-version: [22.18.0]
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0
 
--- a/.github/workflows/validate-remote-assets.yml
+++ b/.github/workflows/validate-remote-assets.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Getting SF_VERSION in the codebase
         run: |
--- a/packages/e2e-playwright/package.json
+++ b/packages/e2e-playwright/package.json
@@ -16,7 +16,7 @@
     "@adyen/adyen-web-server": "1.0.0",
     "@axe-core/playwright": "^4.10.1",
     "@playwright/test": "1.50.0",
-    "dotenv": "17.2.2",
+    "dotenv": "17.2.3",
     "ts-loader": "9.4.4",
     "typescript": "5.2.2"
   }
--- a/packages/lib/package.json
+++ b/packages/lib/package.json
@@ -104,7 +104,7 @@
         "@types/jest": "29.5.14",
         "autoprefixer": "10.4.21",
         "cssnano": "7.1.1",
-        "dotenv": "17.2.2",
+        "dotenv": "17.2.3",
         "enzyme": "3.11.0",
         "enzyme-adapter-preact-pure": "4.1.0",
         "eslint": "9.35.0",
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -25,7 +25,7 @@
         "autoprefixer": "^10.4.9",
         "core-js-pure": "^3.25.3",
         "css-loader": "^6.0.0",
-        "dotenv": "17.2.2",
+        "dotenv": "17.2.3",
         "html-webpack-plugin": "5.6.3",
         "postcss": "^8.4.31",
         "postcss-loader": "^7.0.0",
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -22,7 +22,7 @@
         "start:storybook": "node start.js storybook"
     },
     "devDependencies": {
-        "dotenv": "17.2.2",
+        "dotenv": "17.2.3",
         "express": "^5.0.0"
     },
     "dependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -36,7 +36,7 @@ __metadata:
     autoprefixer: "npm:^10.4.9"
     core-js-pure: "npm:^3.25.3"
     css-loader: "npm:^6.0.0"
-    dotenv: "npm:17.2.2"
+    dotenv: "npm:17.2.3"
     html-webpack-plugin: "npm:5.6.3"
     postcss: "npm:^8.4.31"
     postcss-loader: "npm:^7.0.0"
@@ -59,7 +59,7 @@ __metadata:
     "@adyen/adyen-web-server": "npm:1.0.0"
     "@axe-core/playwright": "npm:^4.10.1"
     "@playwright/test": "npm:1.50.0"
-    dotenv: "npm:17.2.2"
+    dotenv: "npm:17.2.3"
     ts-loader: "npm:9.4.4"
     typescript: "npm:5.2.2"
   languageName: unknown
@@ -69,7 +69,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "@adyen/adyen-web-server@workspace:packages/server"
   dependencies:
-    dotenv: "npm:17.2.2"
+    dotenv: "npm:17.2.3"
     express: "npm:^5.0.0"
     serverless-http: "npm:^4.0.0"
   languageName: unknown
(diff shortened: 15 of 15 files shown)


packages/lib/src/components/internal/QRLoader/components/QRCountdown.tsx
```diff
@@ -9,7 +9,7 @@ const QRCountdown = ({ countdownTime, timeToPay, onTick, onCompleted }: QRCountd
     const timeToPayString = i18n.get(timeToPay).split('%@');
 
     return (
-        <span aria-live="polite" className="adyen-checkout__qr-loader__countdown">
```
johnayeni, line 12: No need for this aria label here since we already have the SRPanel

packages/lib/src/components/Blik/Blik.tsx
```diff
@@ -64,9 +64,6 @@ class BlikElement extends UIElement<AwaitConfiguration> {
                 <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources}>
                     <SRPanelProvider srPanel={this.props.modules.srPanel}>
                         <Await
-                            ref={ref => {
```
johnayeni, line 67: Both QR loader and the Await component does not use this ref. The ref is usually used to hook into a component's `setStatus` and `showValidation` functions, both of which these components don't implement. This makes passing this ref unnecessary. If we ever need this component's to do so, we can add this later.

packages/lib/src/components/ANCV/ANCV.tsx
```diff
@@ -99,9 +99,6 @@ export class ANCVElement extends UIElement<ANCVConfiguration> {
                 <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources}>
                     <SRPanelProvider srPanel={this.props.modules.srPanel}>
                         <Await
-                            ref={ref => {
```
johnayeni, line 102: Both QR loader and the Await component does not use this ref. The ref is usually used to hook into a component's `setStatus` and `showValidation` functions, both of which these components don't implement. This makes passing this ref unnecessary. If we ever need this component's to do so, we can add this later.

packages/lib/src/components/helpers/QRLoaderContainer/QRLoaderContainer.tsx
```diff
@@ -35,9 +35,6 @@ class QRLoaderContainer<T extends QRLoaderConfiguration = QRLoaderConfiguration>
             <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources}>
                 <SRPanelProvider srPanel={this.props.modules.srPanel}>
                     <QRLoader
-                        ref={ref => {
```
johnayeni, line 37: Same comment as here https://github.com/Adyen/adyen-web/pull/3591/files#r2444886264

packages/lib/src/components/PayTo/PayTo.tsx
```diff
@@ -131,9 +131,6 @@ export class PayToElement extends UIElement<PayToConfiguration> {
                 <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources}>
                     <SRPanelProvider srPanel={this.props.modules.srPanel}>
                         <Await
-                            ref={ref => {
```
johnayeni, line 134: Same comment as here https://github.com/Adyen/adyen-web/pull/3591/files#r2444886264

packages/lib/src/components/Pix/Pix.tsx
```diff
@@ -50,9 +50,6 @@ class PixElement extends QRLoaderContainer<PixConfiguration> {
             <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources}>
                 <SRPanelProvider srPanel={this.props.modules.srPanel}>
                     <QRLoader
-                        ref={ref => {
```
johnayeni, line 53: Same comment as here https://github.com/Adyen/adyen-web/pull/3591/files#r2444886264

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -104,9 +104,6 @@ class UPI extends UIElement<UPIConfiguration> {
             case 'qrCode':
                 return (
                     <QRLoader
-                        ref={ref => {
```
johnayeni, line 107: Same comment as here https://github.com/Adyen/adyen-web/pull/3591/files#r2444886264

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -121,9 +118,6 @@ class UPI extends UIElement<UPIConfiguration> {
             case 'await':
                 return (
                     <Await
-                        ref={ref => {
```
johnayeni, line 124: Same comment as here https://github.com/Adyen/adyen-web/pull/3591/files#r2444886264

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,244 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
+        setPercentage(time.percentage);
     };
 
-    private onTimeUp = (): void => {
-        this.setState({ expired: true });
-        clearTimeout(this.timeoutId);
-        this.props.onError(new AdyenCheckoutError('ERROR', 'Payment Expired'));
+    const onTimeUp = (): void => {
+        setExpired(true);
+        clearTimeout(timeoutRef.current);
```
longyulongyu: It seems we can remove this line `clearTimeout(timeoutRef.current);`? Because we are handling it the the `useEffect`.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,244 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
     };
 
-    render({ amount, showAmount, url, brandLogo, brandName, countdownTime, type, onActionHandled }: QRLoaderProps, { expired, completed, loading }) {
-        const { i18n, loadingContext } = useCoreContext();
-        const getImage = useImage();
+    useEffect(() => {
+        void checkStatus();
+    }, []);
+
+    useEffect(() => {
```
longyulongyu, line 103: Potentially this can move to a custom hook and but we can do that in a separate PR, we already have a tech debt ticket. This is a nice prep for it.

johnayeni: Indeed, this is the idea. The next thing i plan to work on is to move things that can be reused between `Await` and `QRLoader` into hooks and or components

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: In this case, we would need to keep it actually because technically, the time could be up before an already set timeout is executed.

longyulongyu: Hmm true, but it should already be handled by `useEffect` no? 

Because when `onTimeUp` is called: `setExpired(true)` is executed, this triggers a re-render of the `QRLoader` component. 

During this re-render, the `useEffect` executes its cleanup function. This cleanup function clears the timeoutId that was set in the previous render right?

After the cleanup, the `useEffect` itself will re-evaluate. Since `expired` is now `true`, it will `clearTimeout(timeoutRef.current)` again.

johnayeni: Indeed! you are right. It is unnecessary there 👍🏿 . I also removed the clear timeout in other places where it was unnecessary. The clean up function handles it just fine.

longyulongyu (dismissed):

packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -49,7 +49,7 @@ function Await(props: AwaitComponentProps) {
                 }
             };
             // Send success response to onAdditionalDetails
-            return props.onComplete(state, this);
+            return props.onComplete(state);
```
ribeiroguilherme: Does removing 'this' imply that in the 'onAdditionalDetails' , the 'component' property is undefined?

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,238 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
+                    paymentData: props.paymentData
                 }
             };
-            this.props.onComplete(state, this);
+            props.onComplete(state);
         }
 
         const error = new AdyenCheckoutError('ERROR', 'error result with no payload in response');
-        return this.props.onError(error);
+        return props.onError(error);
```
ribeiroguilherme: There is no need to return here.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,238 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
+    const onQrCodeLoad = () => {
+        onActionHandled?.({
+            componentType: props.type,
+            actionDescription: 'qr-code-loaded'
+        });
+    };
 
+    const finalState = (image: string, message: string) => {
+        const status = i18n.get(message);
+        useA11yReporter(status);
```
ribeiroguilherme: React hooks shouldn't be used conditionally. I see that this is not a 'new' change to it, but this can lead to misbehavior if I am not mistaken.

Could that be moved to the top of the Component?

packages/lib/src/components/internal/Await/Await.tsx
```diff
@@ -25,7 +25,7 @@ function Await(props: AwaitComponentProps) {
     const [percentage, setPercentage] = useState(100);
     const [timePassed, setTimePassed] = useState(0);
     const [hasAdjustedTime, setHasAdjustedTime] = useState(false);
-    const [storedTimeout, setStoredTimeout] = useState(null);
+    const [storedTimeout, setStoredTimeout] = useState<NodeJS.Timeout | number | null>(null);
```
m1aw, line 28: I thought that was this discussion in another PR about this, but I couldn't find it now.
Why are we using `NodeJS.Timeout`? This would be the NodeJS interface, and wouldn't be available in the browser, but I get that it feel better type safety then number, it just feels misleading.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,238 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
-                    handleCopy={handleCopy}
-                >
-                    {this.props.children ? this.props.children : <QRDetails />}
-                </QRLoaderDetailsProvider>
+                <Spinner />
             </div>
         );
     }
+
+    const qrSubtitleRef = useAutoFocus();
```
ribeiroguilherme: Same hook comment applies here too

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,238 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
         if (status.props.payload) {
             const state = {
                 data: {
                     details: { payload: status.props.payload },
-                    paymentData: this.props.paymentData
+                    paymentData: props.paymentData
                 }
             };
-            this.props.onComplete(state, this);
+            props.onComplete(state);
```
m1aw, line 72: Think out load, should we stop sending the component here? I'm not sure there's any use for it but comparing to the code on the Await component we still send the component on the `onComplete `.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -19,251 +18,238 @@ import { AnalyticsInfoEvent } from '../../../core/Analytics/AnalyticsInfoEvent';
...
     };
 
-    private checkStatus = () => {
-        const { paymentData, clientKey, loadingContext, throttledInterval } = this.props;
+    const checkStatus = async (): Promise<void> => {
+        const { paymentData, clientKey, throttledInterval } = props;
 
         return checkPaymentStatus(paymentData, clientKey, loadingContext, throttledInterval)
             .then(processResponse)
             .catch(response => ({ type: 'network-error', props: response }))
```
m1aw, line 84: Also out of the scope, but I was going over Await and noticed there's a slight difference in implementation of this function, it checks for i18n translation key for the error. Unsure if it's necessary. 
EDIT: Feel free to just resolve this issue immediately and tackle when we merge Await and QRLoader, I didn't see the comment bellow.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
m1aw: Adding some more to this. 
After checking both implementations, I not sure if we should have this a 2 separate components that handle almost same timer logic in fast different ways. Indeed we should probably refactor this into a hook.

m1aw (dismissed):

ribeiroguilherme (approved):

packages/lib/src/components/internal/Await/Await.tsx
johnayeni: Yeah, but if we set the type to just `number` then typescript complains about it. We might need to overwrite the type maybe in a `d.ts` file so that we can just use `number`.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: It is also removed from `Await` in this PR as it is also unneccessary there.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: Indeed, I can look into this in a follow up PR 👍🏿

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: Fixed

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: Fixed, I moved this JSX into a seperate component

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: Fixed

PR #3598, opened 2025-10-21
Analytics - Sending render events part 1

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

This PR changes the mechanism how we dispatch 'render' events.  This is the first part of a series of changes.

Core changes:
- A `beforeRender`  hook was created on `UIElement`.  It is executed every time the 'render()' is called. We can use this `beforeRender` to dispatch render events, but it can be extended to more things in the future
- `beforeRender` can be overiden by payment methods, so if a payment method has a specific payload for the render analytics event, that can be changed in a single place
- `AnalyticsInfoEvent` is responsible to generate the `configData` and sanitize it accordingly for 'rendered' events
- Drop-in does not send 'rendered' events on behalf of other payment methods anymore. All the logic is contained on each payment method
- `InstantPaymentMethods` and `PaymentMethodDetails` components are caching their content by using `memo`. This let us not trigger 'render()' multiple times when Drop-in receives state updates
- `PaymentMethods` class will parse the payment methods response from the API, and it will generate an unique ID per non-stored payment method (`_id`) . This is useful since there is a use-case where there are multiple payment methods with the same tx variant in the array (e.g. giftcards, scheme ). The UI element has to find the correct payment method to when building the Component props, and the only way to achieve that is by generating an unique identifier.

This change above touches a very core function (`render()`) , therefore many tests had to be adjusted since they are trying to dispatch 'rendered' events. To accomodate that:
- Created a `setup-core-mock.ts` which creates a mock of Core on demand. The mock can be customized via props. The mock exposes the Analytics module mocked, which is useful since we don't have to manually create and pass Analytics module as props when rendering payment methods

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/GooglePay/GooglePay.tsx
+++ b/packages/lib/src/components/GooglePay/GooglePay.tsx
@@ -135,7 +135,6 @@ class GooglePay extends UIElement<GooglePayConfiguration> {
     }
 
     protected override beforeRender(configSetByMerchant: GooglePayConfiguration) {
-        console.log('render');
         const event = new AnalyticsInfoEvent({
             type: InfoEventType.rendered,
             component: this.type,
--- a/packages/lib/src/core/Analytics/AnalyticsInfoEvent.ts
+++ b/packages/lib/src/core/Analytics/AnalyticsInfoEvent.ts
@@ -104,7 +104,15 @@ export class AnalyticsInfoEvent extends AnalyticsEvent {
      */
     private get configDataExcludedFields() {
         const DROPIN_FIELDS = ['paymentMethodsConfiguration'];
-        const FIELDS_INJECTED_BY_DROPIN = ['elementRef', 'isDropin', 'oneClick', 'storedPaymentMethodId', 'isInstantPayment', 'type'];
+        const FIELDS_INJECTED_BY_DROPIN = [
+            'elementRef',
+            'isDropin',
+            'oneClick',
+            'storedPaymentMethodId',
+            'paymentMethodId',
+            'isInstantPayment',
+            'type'
+        ];
         const PII_FIELDS = ['data', 'shopperEmail', 'telephoneNumber'];
 
         /**
@@ -125,19 +133,24 @@ export class AnalyticsInfoEvent extends AnalyticsEvent {
         const MAX_STRING_LENGTH = 128;
         const result = {};
 
-        for (const [key, value] of Object.entries(config)) {
-            if (!this.configDataExcludedFields.includes(key)) {
-                if (typeof value === 'function') {
-                    result[key] = 'function';
-                } else if (typeof value === 'object' && value !== null) {
-                    result[key] = JSON.stringify(value).substring(0, MAX_STRING_LENGTH);
-                } else {
-                    result[key] = value;
+        try {
+            for (const [key, value] of Object.entries(config)) {
+                if (!this.configDataExcludedFields.includes(key)) {
+                    if (typeof value === 'function') {
+                        result[key] = 'function';
+                    } else if (typeof value === 'object' && value !== null) {
+                        result[key] = JSON.stringify(value).substring(0, MAX_STRING_LENGTH);
+                    } else {
+                        result[key] = value;
+                    }
                 }
             }
-        }
 
-        return result;
+            return result;
+        } catch (error: unknown) {
+            if (process.env.NODE_ENV === 'development') console.warn('AnalyticsInfoEvent: Error when creating configData\n', error);
+            return result;
+        }
     }
 
     public getEventCategory(): string {


.github/pull_request_template.md
```diff
@@ -8,8 +8,9 @@ Check the checkboxes that are relevant for your pull request.
 - [ ] I have added unit tests to cover my changes.
 - [ ] I have added or updated Storybook stories where applicable.
 - [ ] I have tested the changes manually in the local environment.
+- [ ] I have checked that no PII data is being sent on analytics events
```
m1aw, line 11: How are we going to check this?

ribeiroguilherme: That is more like a reminder when creating new payment methods.  
You should check on 'rendered' event if the component is propagating PII data.

packages/lib/src/components/Dropin/components/PaymentMethod/InstantPaymentMethods.tsx
```diff
@@ -1,4 +1,5 @@
 import { h } from 'preact';
+import { memo } from 'preact/compat';
```
sponglord, line 2: Do we really want to start using `preact/compat` just so we can do this?

How about:
```
import { memo } from 'preact';

function InstantPaymentMethods({ paymentMethods, onSelect }: InstantPaymentMethodsProps) {
    return (
        <ul className="adyen-checkout__instant-payment-methods-list">
            {paymentMethods.map(pm => (
                // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
                <li key={pm._id} data-testid={pm.type} onClick={() => onSelect(pm)}>
                    {pm.render()}
                </li>
            ))}
        </ul>
    );
}

// Export the memoized version
export default memo(InstantPaymentMethods);
```

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -90,21 +134,26 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+        return this.core.paymentMethodsResponse.findStoredPaymentMethod(storedPaymentMethodId);
+    }
+
     /**
      *  Get the payment method from the paymentMethodsResponse
      *
      * @param type - The type of the payment method to get. (This prop is passed by Drop-in OR Standalone components containing the property 'type' as part of their configuration)
      */
-    protected getPaymentMethodFromPaymentMethodsResponse(type?: string): PaymentMethod {
+    protected getPaymentMethodFromPaymentMethodsResponse(type?: string, paymentMethodId?: string): RawPaymentMethod {
```
longyulongyu, line 155: Just nitpicking can you update the JSDoc for `paymentMethodId`?

packages/lib/src/components/GooglePay/GooglePay.tsx
```diff
@@ -65,7 +59,9 @@ class GooglePay extends UIElement<GooglePayConfiguration> {
      * @param type
      * @returns
      */
-    protected override getPaymentMethodFromPaymentMethodsResponse(type?: string): PaymentMethod {
+    protected override getPaymentMethodFromPaymentMethodsResponse(type?: string, paymentMethodId?: string): RawPaymentMethod {
```
longyulongyu, line 62: same here, can you update the JSDoc for `paymentMethodId`?

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -180,8 +229,7 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
     protected submitAnalytics(analyticsObj: AnalyticsEvent) {
         try {
             analyticsObj.component = this.getComponent(analyticsObj);
-
-            this.props.modules.analytics.sendAnalytics(analyticsObj);
+            this.analytics.sendAnalytics(analyticsObj);
```
longyulongyu, line 232: A minor improvement, and not related to your change: it would be nice to avoid mutating the `analyticsObj` directly.

packages/lib/src/components/internal/UIElement/UIElement.tsx
longyulongyu, line 305: Not related to your change, but i think we can send the analytics outside of the promise.

```js
    private submitUsingAdvancedFlow(): Promise<CheckoutAdvancedFlowResponse> {
        // Call analytics endpoint
        const event = new AnalyticsLogEvent({
            type: ANALYTICS_SUBMIT_STR,
            message: 'Shopper clicked pay'
        });
        this.submitAnalytics(event);

        return new Promise<CheckoutAdvancedFlowResponse>((resolve, reject) =>
            this.props.onSubmit({ data: this.data, isValid: this.isValid }, this.elementRef, { resolve, reject })
        );
    }
```

packages/lib/src/core/Analytics/AnalyticsInfoEvent.ts
```diff
@@ -83,6 +98,61 @@ export class AnalyticsInfoEvent extends AnalyticsEvent {
         return this;
     }
 
+    /**
+     * Set of properties that must not be included when creating the configData for Analytics
+     * @private
+     */
+    private get configDataExcludedFields() {
```
longyulongyu, line 105: It could be a static getter:
```ts
private static get configDataExcludedFields() {}
```

sponglord:
As it stands this PR is completely killing the `configData` sent by the `storedCard` and `Card` component.
Whereas they used to send a `configData` object populated with all their configured props (35+) when they created a "rendered" event - now this object is barely populated at all. (I can share examples of the differences)

I know this PR is part of a series of changes, but is the above behaviour expected at this stage?

packages/lib/src/components/Dropin/components/PaymentMethod/InstantPaymentMethods.tsx
ribeiroguilherme: Yes. As of now, we are not allowed to use `forwardRef`, but other utilities can be used. `memo` is quite useful in this specific scenario

packages/lib/src/components/internal/UIElement/UIElement.tsx
ribeiroguilherme: I will update in the next PR 👍

packages/lib/src/components/GooglePay/GooglePay.tsx
ribeiroguilherme: I will update in the next PR 👍

packages/lib/src/components/internal/UIElement/UIElement.tsx
ribeiroguilherme: That will be done in the following PR's 👍

packages/lib/src/components/internal/UIElement/UIElement.tsx
ribeiroguilherme: I will update in the next PR 👍

packages/lib/src/core/Analytics/AnalyticsInfoEvent.ts
ribeiroguilherme: I can change that, although I am not sure if there are any benefits in doing that. As it is used internally, seems like it can be kept as is

ribeiroguilherme:
> As it stands this PR is completely killing the `configData` sent by the `storedCard` and `Card` component. Whereas they used to send a `configData` object populated with all their configured props (35+) when they created a "rendered" event - now this object is barely populated at all. (I can share examples of the differences)
> 
> I know this PR is part of a series of changes, but is the above behaviour expected at this stage?

Yes, that is intentional. The configData is intended to "give details of the merchant-set configuration for the rendered component" . It is not the case currently, as it is spamming all props being used by the Component. (props set by the merchant but also internal props/default props).

In the next PR's I will adjust properly the data that must be passed by all payment method components. If you want to perform some tests to check, I suggest to check the final PR once it is ready

PR #3605, opened 2025-10-23
Fix Gift Card error message display issue

m1aw:
## 📋 Pull Request Checklist

---

## Problem
Gift card error messages (no-balance, card-error, currency-error) stopped displaying because `UIElement.handleError()` always resets status to 'ready', overriding the specific error status set by the gift card component.

**Root Cause**: The `handleError()` method calls `setElementStatus('ready')` which overwrites the component's internal status, preventing specific error messages from being displayed to users.

## Solution
Implemented a new error handling approach that separates balance check errors from the component's main status, allowing error messages to display correctly while maintaining proper error callbacks.

### 1. **Enhanced Error Handling Architecture**
- **Before**: Balance check errors were set via `setStatus(error.message)` and immediately overridden by `handleError()`
- **After**: Balance check errors are managed through a unified `transformedErrors` state that merges SFP (SecuredFields) errors with balance check errors

### 2. **Updated `GiftcardComponent.tsx`**
- Added `updateTransformedErrors()` method to merge errors
- Created `generateBalanceCheckErrors()` to return errors in the same format as `useForm` errors
- Updated `getCardErrorMessage()` to use the unified `transformedErrors` structure
- Enhanced `setBalanceCheckErrors()` to trigger error updates when called from parent component

## 🧪 Tested scenarios

- All existing tests pass
- New test patterns that check for this error messages in the UI
- SR Panel tested manually, also working

---

**Closes**:  

---

### DIFF
--- a/.changeset/loose-islands-wave.md
+++ b/.changeset/loose-islands-wave.md
@@ -1,0 +1,6 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Move StoredCardFields related styles into its own scss file.
+Refactor internal styles.
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.scss
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.scss
@@ -69,22 +69,9 @@
     padding-left: token(spacer-060);
 }
 
-.adyen-checkout__card__exp-date__input--oneclick {
-    line-height: token(text-title-line-height);
-    font-weight: token(text-body-font-weight);
-    text-overflow: ellipsis;
-    overflow: hidden;
-    white-space: nowrap;
-    text-align: left;
-}
-
-.adyen-checkout__field--storedCard,
-.adyen-checkout__field--expiryDate,
 .adyen-checkout__card__holderName {
     margin-bottom: 0;
 }
-
-
 
 .adyen-checkout__card__holderName,
 .adyen-checkout__card__kcp-authentication,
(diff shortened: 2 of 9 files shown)


packages/lib/src/components/Giftcard/components/GiftcardComponent.tsx
```diff
@@ -91,6 +94,55 @@ class Giftcard extends Component<GiftcardComponentProps> {
...
+        // This is the field that the error is associated with, only used for SR logic
+        const fieldToAnnounce = 'encryptedCardNumber';
+
+        // If errorType is null, clear errors
+        if (errorType === null) {
+            return balanceCheckErrors;
+        }
+
+        // Use provided errorType or get from component state
+        if (errorType) {
```
longyulongyu: I think we can remove the check for `errorType` existence.

packages/lib/src/components/Giftcard/Giftcard.tsx
longyulongyu, line 82: Not sure if it's intended but if the previous card already received `no-balance` error, i can keep submitting the same card number (keep calling the `handleBalanceCheck`).

packages/lib/src/components/Giftcard/types.ts
```diff
@@ -47,3 +47,22 @@ export interface GiftCardConfiguration extends UIElementProps {
...
+    errorI18n?: string;
+    error?: string; // Original error code
+}
+
+export type GiftCardBalanceCheckErrorType = 'no-balance' | 'card-error' | 'currency-error';
+
+/**
+ * Balance check errors for gift card components
+ */
+export interface GiftCardBalanceCheckErrors {
```
longyulongyu: The interface seems not used anywhere.

packages/lib/src/components/Giftcard/Giftcard.tsx
m1aw: Yeah, I didn't want to change this flow. That's how it used to work on V5, you should be able to re-submit the same card several times.

longyulongyu (approved):

PR #3618, opened 2025-10-30
Cosdk 684 stop propagation for copy button on enter key press

longyulongyu:
## 📋 Pull Request Checklist

---

## 📝 Summary

- Created a shared copy button component
- Fixed some style issues

---

## 🧪 Tested scenarios

---

**Closes**:  COSDK-684

---

### DIFF
--- a/packages/lib/src/components/internal/Button/CopyButton.tsx
+++ b/packages/lib/src/components/internal/Button/CopyButton.tsx
@@ -6,8 +6,9 @@ import { useCoreContext } from '../../../core/Context/CoreProvider';
 import copyToClipboard from '../../../utils/clipboard';
 import { PREFIX } from '../Icon/constants';
 import useImage from '../../../core/Context/useImage';
+import { stopPropagationForActionKeys } from './stopPropagationForActionKeys';
 
-export interface CopyButtonProps extends Omit<ButtonProps, 'onClickCompletedLabel'> {
+export interface CopyButtonProps extends Omit<ButtonProps, 'variant' | 'onClickCompletedLabel'> {
     /**
      * String that will get copied to the clipboard
      */
@@ -29,27 +30,14 @@ const CopyButton = (props: CopyButtonProps) => {
         [props.text, props.onClick]
     );
 
-    // todo: we should remove it from the whole sdk: onKeyPress is deprecated.
-    const onKeyPress = useCallback((event: KeyboardEvent) => {
-        if (event.key === 'Enter' || event.code === 'Enter' || event.key === ' ' || event.code === 'Space') {
-            event.stopPropagation();
-        }
-    }, []);
-
-    const onKeyDown = useCallback((event: KeyboardEvent) => {
-        if (event.key === 'Enter' || event.code === 'Enter' || event.key === ' ' || event.code === 'Space') {
-            event.stopPropagation();
-        }
-    }, []);
-
     return (
         <Button
             {...props}
             variant="action"
             onClick={onClick}
             // It's ok to have both, browsers will fire only one click event for enter key pressed.
-            onKeyPress={onKeyPress}
-            onKeyDown={onKeyDown}
+            onKeyPress={stopPropagationForActionKeys}
+            onKeyDown={stopPropagationForActionKeys}
             icon={props.icon ?? getImage({ imageFolder: 'components/' })(`${PREFIX}copy`)}
             label={props.label ?? i18n.get('button.copy')}
             onClickCompletedLabel={props.copiedLabel ?? i18n.get('button.copied')}
--- a/.changeset/large-stamps-appear.md
+++ b/.changeset/large-stamps-appear.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Introduced a shared `CopyButton` component with unified clipboard logic and styling enhancements to the `Voucher` component.
--- a/packages/lib/src/components/internal/Button/CopyIconButton.tsx
+++ b/packages/lib/src/components/internal/Button/CopyIconButton.tsx
@@ -6,6 +6,7 @@ import { useCoreContext } from '../../../core/Context/CoreProvider';
 import copyToClipboard from '../../../utils/clipboard';
 import { SingletonTooltipProvider, useTooltip } from '../Tooltip/SingletonTooltipProvider';
 import './CopyIconButton.scss';
+import { stopPropagationForActionKeys } from './stopPropagationForActionKeys';
 
 export interface CopyIconButtonProps extends ButtonProps {
     /**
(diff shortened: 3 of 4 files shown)


packages/lib/src/components/internal/Button/CopyButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+        },
+        [props.text, props.onClick]
+    );
+
+    // todo: we should remove it from the whole sdk: onKeyPress is deprecated.
+    const onKeyPress = useCallback((event: KeyboardEvent) => {
+        if (event.key === 'Enter' || event.code === 'Enter' || event.key === ' ' || event.code === 'Space') {
+            event.stopPropagation();
+        }
+    }, []);
```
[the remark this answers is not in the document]
longyulongyu: It will be addressed in a different PR

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -135,15 +134,12 @@ function QRLoader(props: QRLoaderProps) {
...
-        void copyToClipboard(props.qrCodeData);
-
+    const handleCopy = () => {
         const event = new AnalyticsInfoEvent({
             type: ANALYTICS_DOWNLOAD_STR,
             target: ANALYTICS_QR_CODE_DOWNLOAD
         });
         props.onSubmitAnalytics(event);
-        complete();
     };
```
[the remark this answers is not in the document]
longyulongyu: It's handled by the `CopyButton` component.

packages/lib/src/components/internal/Voucher/Voucher.scss
```diff
@@ -11,7 +12,7 @@
     background: token(color-background-primary);
     border: token(border-width-s) solid token(color-separator-primary);
 
-    .adyen-checkout__details-table{
```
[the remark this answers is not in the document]
longyulongyu: There is a space in the change..

packages/lib/src/components/internal/Button/CopyButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const onKeyDown = useCallback((event: KeyboardEvent) => {
+        if (event.key === 'Enter' || event.code === 'Enter' || event.key === ' ' || event.code === 'Space') {
+            event.stopPropagation();
+        }
+    }, []);
+
+    return (
+        <Button
+            inline
```
johnayeni: Seems we are not yet passing the rest of the button props to the component that will allow the look of this button to be customized.

```tsx
<Button {...restProps} />
```

Also the button on the QRLoader component no longer uses the `inline` button. I would suggest we don't set the button as line default here but allow places where an inline button to should be used to pass the prop. The `variant="action"` prop could still be left by default but the spreading of the remaining props  `{...restProps}` should be able to override it.

packages/lib/src/components/internal/Button/Button.stories.tsx
```diff
@@ -40,7 +42,7 @@ const coreProps = {
             confirmPreauthorization: 'Confirm preauthorization'
         }
     }),
-    resources: global.resources
+    resources: new Resources('https://checkoutshopper-test.cdn.adyen.com/checkoutshopper/')
```
johnayeni: Is there a way for us not to hardcode this value? Maybe by using the environment variables

packages/lib/src/components/internal/QRLoader/QRLoader.scss
```diff
@@ -100,6 +100,9 @@
 }
 
 .adyen-checkout__qr-loader__actions {
+    display: flex;
```
johnayeni, line 103: I wonder if at some point we would consider having a solution for our css like tailwind or at least writing some reusable utilities. Adding more of these styles would continue to increase our css bundle.

packages/lib/src/components/internal/Button/Button.stories.tsx
longyulongyu: You're right, I think we can use environment variables together with the `resolveEnvironments` function.

packages/lib/src/components/internal/Button/CopyButton.tsx
longyulongyu: Nice catch, I’ll update it.

packages/lib/src/components/internal/QRLoader/QRLoader.scss
longyulongyu: 👍 I am open to use tailwind and clean up our css code. Maybe a topic for the tech talk?

packages/lib/src/components/internal/Button/CopyButton.tsx
longyulongyu: Updated

packages/lib/src/components/internal/Button/Button.stories.tsx
longyulongyu: Updated.

packages/lib/src/components/internal/Button/CopyButton.tsx
```diff
@@ -0,0 +1,60 @@
+import { h } from 'preact';
+import { useCallback } from 'preact/hooks';
+import Button from './Button';
+import { ButtonProps } from './types';
+import { useCoreContext } from '../../../core/Context/CoreProvider';
+import copyToClipboard from '../../../utils/clipboard';
+import { PREFIX } from '../Icon/constants';
+import useImage from '../../../core/Context/useImage';
+
+export interface CopyButtonProps extends Omit<ButtonProps, 'onClickCompletedLabel'> {
```
johnayeni:
```suggestion
export interface CopyButtonProps extends Omit<ButtonProps,'variant' | 'onClickCompletedLabel'> {
```

Since `variant="action` cannot be overridden,  let's remove it from the props that can be passed for this component

packages/lib/src/components/internal/Button/CopyButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+            void copyToClipboard(props.text);
+            complete();
+            props.onClick?.(e);
+        },
+        [props.text, props.onClick]
+    );
+
+    // todo: we should remove it from the whole sdk: onKeyPress is deprecated.
+    const onKeyPress = useCallback((event: KeyboardEvent) => {
+        if (event.key === 'Enter' || event.code === 'Enter' || event.key === ' ' || event.code === 'Space') {
```
johnayeni: Might make sense to move this to a reusuable function since we use this logic in quite some places now e.g `isFocusedElementKeyPress(event)`

packages/lib/src/components/internal/Button/CopyButton.tsx
longyulongyu: Done

packages/lib/src/components/internal/Button/CopyButton.tsx
longyulongyu: Done

johnayeni (approved):

PR #3638, opened 2025-11-11
[QR loader and Await]:  Share timer logic between await and qr component

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

- Move timer logic for QR loader and await component to a shared hook
- Use named exports for QR loader and Await components

---

## 🧪 Tested scenarios

- [x] Manual testing
- [x] Unit tests

---

**Closes**:  

---

### DIFF
--- a/.github/workflows/.size-limit.yml
+++ b/.github/workflows/.size-limit.yml
@@ -11,7 +11,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Use Node.js 22.18.0
-        uses: actions/setup-node@v5
+        uses: actions/setup-node@v6
         with:
           node-version: 22.18.0
           package-manager-cache: false
--- a/.github/workflows/automated-a11y.yml
+++ b/.github/workflows/automated-a11y.yml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/checkout@v5
     
     - name: Setup Node
-      uses: actions/setup-node@v5
+      uses: actions/setup-node@v6
       with:
         node-version: ${{ matrix.node-version }}
         package-manager-cache: false
@@ -75,7 +75,7 @@ jobs:
     
     - name: Archive test result artifacts
       if: always()
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v5
       with:
         name: a11y-report-${{matrix.api-version}}-${{matrix.runner-browser}}-${{ github.sha }}
         path: packages/e2e-playwright/playwright-report
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/checkout@v5
 
     - name: Setup Node
-      uses: actions/setup-node@v5
+      uses: actions/setup-node@v6
       with:
         node-version: ${{ matrix.node-version }}
         package-manager-cache: false
@@ -75,7 +75,7 @@ jobs:
 
     - name: Archive test result artifacts
       if: always()
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@v5
       with:
         name: playwright-report-${{matrix.api-version}}-${{matrix.runner-browser}}-${{ github.sha }}
         path: packages/e2e-playwright/playwright-report
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -11,7 +11,7 @@ jobs:
       - uses: actions/checkout@v5
 
       # Setup .npmrc file to publish to npm
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: 22.18.0
           registry-url: https://registry.npmjs.org/
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@v5
 
       # Setup .npmrc file to publish to npm
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: 22.18.0
           registry-url: https://registry.npmjs.org/
@@ -33,7 +33,7 @@ jobs:
 
       # Upload UMD and translations
       - run: cp packages/server/translations/ packages/lib/dist/umd -r
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v5
         with:
           name: umd-and-translations
           path: packages/lib/dist/umd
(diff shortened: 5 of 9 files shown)


packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
```diff
@@ -0,0 +1,148 @@
...
+        setExpired(true);
+        props.onError(new AdyenCheckoutError('ERROR', 'Payment Expired'));
+    };
+
+    const onComplete = (status: PaymentStatusObject): void => {
+        setCompleted(true);
+        setLoading(false);
+
+        if (status.props.payload) {
+            const state = {
```
ribeiroguilherme: I think this `state` can be typed.  We have the type `AdditionalDetailsData` , perhaps it can be used here.

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
```diff
@@ -0,0 +1,148 @@
...
+            props.onError(new AdyenCheckoutError('ERROR', 'successful result, but no payload in response'));
+        }
+    };
+
+    const onError = (status: PaymentStatusObject): void => {
+        setExpired(true);
+        setLoading(false);
+
+        if (status.props.payload) {
+            const state = {
```
ribeiroguilherme: Same `type` related comment here (related to [this](https://github.com/Adyen/adyen-web/pull/3638#discussion_r2514543819) )

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
```diff
@@ -0,0 +1,148 @@
...
+        } else {
+            const error = new AdyenCheckoutError('ERROR', 'error result with no payload in response');
+            props.onError(error);
+        }
+    };
+
+    const checkStatus = async (): Promise<void> => {
+        const { paymentData, clientKey, throttleInterval, pollStatus } = props;
+
+        const pollStatusFunction = pollStatus ?? (() => checkPaymentStatus(paymentData, clientKey, props.loadingContext, throttleInterval));
```
ribeiroguilherme, line 72: The return type of the `pollStatusFunction` is `Promise<any>` right now. We should make it more strict. I believe we can do that in this PR.

First thing would be to set the strict return type for the `checkPaymentStatus` function (right now it is `Promise<any>` ) . Then we also need to use the same defined type and apply to the hook prop `pollStatus`.

Right now, `pollStatus` prop is typed as `Promise<RawPaymentResponse>` , but I guess it could be changed to something like `Promise<RawStatusCheckResponse>` (that seems more correct IMO , although fields might overlap). 

And then perhaps we can use the `RawStatusCheckResponse` instead of the new `PaymentStatusObject` type defined on `global-types.ts`

packages/lib/src/types/global-types.ts
```diff
@@ -399,3 +399,14 @@ export type DecodeObject = {
...
+interface PaymentStatusObjectProps {
+    payload?: string | null;
+    resultCode: string;
+    type: string;
+}
+
+export interface PaymentStatusObject {
+    type: string;
+    props: PaymentStatusObjectProps;
+}
```
ribeiroguilherme: Creating types and adding suffix as `Object` seems a bit redundant, since everything is an object in Javascript 😅 

I believe the comment I left about typing the [pollStatus](https://github.com/Adyen/adyen-web/pull/3638#discussion_r2514599705)  might impact these types.

packages/lib/src/hooks/usePaymentStatusTimer/types.ts
```diff
@@ -0,0 +1,36 @@
+import { CountdownTime } from '../../components/internal/Countdown/types';
+import { AdyenCheckoutError, RawPaymentResponse } from '../../types';
+
+export interface PaymentCompeteStatus {
```
ribeiroguilherme: Is that a typo? "Payment**Compete**Status"

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+        if (!onPollingStartedActionHandled) {
+            props.onActionHandled?.({ componentType: props.type, actionDescription: 'polling-started' });
+            setOnPollingStartedActionHandled(true);
+        }
+
+        return pollStatusFunction()
+            .then(processResponse)
+            .catch(response => ({
```
ribeiroguilherme: We should try to type the `response` here if possible

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
johnayeni: [fixed] I also renamed the name of the variable to `additionalDetailsData` as it is closer to what the data means

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
johnayeni: [fixed] I also renamed the name of the variable to `additionalDetailsData` as it is closer to what the data means

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
johnayeni: I created a new `RawStatusCheckResponse ` type for this and I have also removed the `PaymentStatusObject ` type. I also went further to rename the `processResponse ` function to `processPaymentStatusResponse` as that is what this utility is being used for actually. Lemme know what you think about the reorganisation here

packages/lib/src/types/global-types.ts
johnayeni: Addressed with this https://github.com/Adyen/adyen-web/pull/3638#discussion_r2518494565

packages/lib/src/hooks/usePaymentStatusTimer/types.ts
johnayeni: This type has been removed because of this https://github.com/Adyen/adyen-web/pull/3638#discussion_r2518494565

packages/lib/src/hooks/usePaymentStatusTimer/usePaymentStatusTimer.tsx
johnayeni: This is a catch block actually so this would not always be a response e.g network errors. So for now, I have types this as `unknown`

PR #3640, opened 2025-11-11
fix: added box shadow to all brand icons and used the black checkmark

longyulongyu:
## 📋 Pull Request Checklist

---

## 📝 Summary
- Added the box shadow to all brands in dropin, card and bancontact
- Removed onclick handler on img element, and moved it to the parent element
- The cursor inside card number, exp date and cvc fields is now `text`
- Removed `DualBrandingIcon`, we can reuse just `BrandIcon`
- `BrandIcon` is reusing `Brand` now
- Added `adyen-checkout-icon-shadow` mixin to reuse
- Used the black checkmark

---

## 🧪 Tested scenarios

---

**Closes**:  COSDK-701

---

### DIFF
--- a/packages/lib/src/components/internal/Brand/Brand.tsx
+++ b/packages/lib/src/components/internal/Brand/Brand.tsx
@@ -12,9 +12,10 @@ interface BrandProps {
     showOnError?: boolean;
 }
 
-const Brand = ({ url, alt, wrapperClassName = '', imgClassName = '' }: BrandProps) => {
+const Brand = ({ url, alt, wrapperClassName = '', imgClassName = '', showOnError }: BrandProps) => {
     const [hasError, setHasError] = useState(false);
-    const classes = cx('adyen-checkout-brand-wrapper', wrapperClassName, { 'adyen-checkout-brand-wrapper--error': hasError });
+    const classesOnError = showOnError ? {} : { 'adyen-checkout-brand-wrapper--error': hasError };
+    const classes = cx('adyen-checkout-brand-wrapper', wrapperClassName, classesOnError);
 
     return (
         <span className={classes}>
--- a/packages/lib/src/components/Card/components/CardInput/components/AvailableBrands/AvailableBrands.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/components/AvailableBrands/AvailableBrands.tsx
@@ -25,7 +25,13 @@ const AvailableBrands = ({ brands, activeBrand }: PaymentMethodBrandsProps) => {
             })}
         >
             {brands.map(({ name, icon }) => (
-                <Brand wrapperClassName={'adyen-checkout__card__brands__brand-wrapper'} key={name} url={icon} alt={getFullBrandName(name)}></Brand>
+                <Brand
+                    showOnError
+                    wrapperClassName={'adyen-checkout__card__brands__brand-wrapper'}
+                    key={name}
+                    url={icon}
+                    alt={getFullBrandName(name)}
+                ></Brand>
             ))}
         </span>
     );


packages/lib/src/components/Card/components/CardInput/components/DualBrandingIcon/DualBrandingIcon.tsx
```diff
@@ -13,13 +13,11 @@ const DualBrandingIcon = ({ brand, onClick, dataValue, brandsConfiguration = {}
     };
 
     return (
-        // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-noninteractive-element-interactions
```
longyulongyu, line 16: `onClick` handler is moved to the parent, we shouldn't add it because image is a non interactive element. Same goes for other similar changes in the PR.

packages/lib/src/components/Dropin/components/PaymentMethod/PaymentMethodIcon.tsx
```diff
@@ -11,23 +10,16 @@ interface PaymentMethodIconProps {
 
     /** Type of the payment method*/
     type: string;
-
```
longyulongyu, line 14: Remove because it's not used anywhere.

packages/lib/src/components/Card/components/CardInput/components/BrandIcon.tsx
```diff
@@ -2,32 +2,18 @@ import { h } from 'preact';
...
-        setHasLoaded(true);
-    };
-
-    const fieldClassnames = classNames({
+    const imgClassName = classNames({
         'adyen-checkout-card-input__icon': true,
-        'adyen-checkout__card__cardNumber__brandIcon': true,
-        'adyen-checkout-card-input__icon--hidden': !hasLoaded
+        'adyen-checkout__card__cardNumber__brandIcon': true
     });
```
johnayeni:
```suggestion
const imgClassName = `adyen-checkout-card-input__icon adyen-checkout__card__cardNumber__brandIcon`;
```

packages/lib/src/components/Card/components/CardInput/components/CardNumber.test.tsx
```diff
@@ -1,51 +1,51 @@
-import { mount } from 'enzyme';
 import { h } from 'preact';
+import { render, screen } from '@testing-library/preact';
```
johnayeni, line 2: 🔥  Thank you!

packages/lib/src/components/internal/Brand/Brand.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface BrandProps {
+    url: string;
+    alt: string;
+    wrapperClassName?: string;
+    imgClassName?: string;
+}
+
+const Brand = ({ url, alt, wrapperClassName = '', imgClassName = '' }: BrandProps) => {
+    return (
+        <span className={cx('adyen-checkout-brand-wrapper', [wrapperClassName])}>
```
johnayeni:
```suggestion
        <span className={cx('adyen-checkout-brand-wrapper', wrapperClassName)}>
```

johnayeni:
Thanks for the changes here @longyulongyu !

packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.scss
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.adyen-checkout__input-wrapper {
+    .adyen-checkout__field__exp-date_hint_wrapper {
+        position: absolute;
+        right: 0;
+        transition: opacity 0.1s linear;
+        margin-right: token(spacer-060);
+        width: 27px;
+        height: 18px;
```
johnayeni, line 24: Do we have any spacer values we could use here?

packages/lib/src/components/Card/components/CardInput/components/DualBrandingIcon/DualBrandingIcon.tsx
```diff
@@ -1,28 +0,0 @@
-import { h } from 'preact';
```
sponglord, line 1: Observation: With the removal of this component - we no longer have any error detection if the brand fails to load.
Which could lead to results like this
image
 \- if the `brand` has not been set correctly in the `/paymentMethods` response or '/binLookup`
 This is an edge-case - but has happened in the past, hence the original fix

sponglord:
I notice that now, when the card number is in error, it keeps the brand icon:
image
whereas in the past that disappeared:
image

Just checking this is a deliberate change?

packages/lib/src/components/Card/components/CardInput/components/ExpirationDate.scss
longyulongyu: Unfortunately there isn't any tokens we can reuse 😢.

packages/lib/src/components/Card/components/CardInput/components/DualBrandingIcon/DualBrandingIcon.tsx
longyulongyu: Strange it should be handled by Img component, let me recheck.

longyulongyu:
> I notice that now, when the card number is in error, it keeps the brand icon: image whereas in the past that disappeared: image
> 
> Just checking this is a deliberate change?

No it's not, will check.

packages/lib/src/components/Card/components/CardInput/components/DualBrandingIcon/DualBrandingIcon.tsx
longyulongyu: How can I reproduce the issue? I tried to mock binLookup with a non exist brand, it seems working.

sponglord: With a single brand it seems to work - although it looks on my screen that it leaves the box shadow behind:
image

With dual branding: I get the broken image:
image

longyulongyu: I used this card number but I cannot reproduce the same behavior, should i put the network offline or simulate the binlookup response?
I had the feeling what needs to be done but just wanted to confirm.

sponglord: You have to mock it. I'll DM you

packages/lib/src/components/Card/components/CardInput/components/BrandIcon.tsx
longyulongyu: Fixed

packages/lib/src/components/internal/Brand/Brand.tsx
longyulongyu: Fixed

packages/lib/src/components/Card/components/CardInput/components/DualBrandingIcon/DualBrandingIcon.tsx
longyulongyu: Both issues are [fixed](https://github.com/Adyen/adyen-web/pull/3640/commits/41e8bdc98a0daf5c52c8d6ff4f83378425bf8345).

sponglord:
What's the difference I'm supposed to be seeing with the box shadow? It might just be my monitor - but I'm not seeing any change to, for instance, the cvc hint

longyulongyu:
> What's the difference I'm supposed to be seeing with the box shadow? It might just be my monitor - but I'm not seeing any change to, for instance, the cvc hint

There was no box-shadow before, now we are applying this mixin to all brands, cvc hint, etc..

You can try this in the mixin `adyen-checkout-icon-shadow`, it will show the box shadow more obvious.
```css
// box-shadow: token(shadow-low);
    box-shadow:
        0 2px 4px red,
        0 1px 2px red;
```

PR #3642, opened 2025-11-12
Analytics - Making events immutable

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary
- Events are now immutable: all properties are private and readonly, and they can not be changed. They must be set when the event is created
- Moved all analytic event types to their own respective files (Info event types moved to the Info event file; Error event types moved to the Error event file; etc) and converted them to enum's
- Simplified UIElement by removing the `getComponent` method
- Moved the Analytic events classes to the `events` folder
- Cleaned up the 'Stored Card story' : we don't have to spread all props into the component itself; we can just pass the stored payment method ID and the Component is able to pick up its data

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
+++ b/packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
@@ -11,6 +11,7 @@ import useImage from '../../../../core/Context/useImage';
 import './Enrollment.scss';
 import { LabelOnlyDisclaimerMessage } from '../../../internal/DisclaimerMessage/DisclaimerMessage';
 import { ComponentMethodsRef, UIElementStatus } from '../../../internal/UIElement/types';
+import { TxVariants } from '../../../tx-variants';
 
 const TERMS_CONDITIONS_URL = 'https://www.adyen.com/pt_BR/legal/termo-de-consentimento-br';
 
@@ -92,7 +93,7 @@ function Enrollment(props: EnrollmentProps) {
                         payButton={props.payButton}
                         showPayButton={true}
                         ref={issuerListRef}
-                        type={'paybybank_pix'}
+                        type={TxVariants.paybybank_pix}
                     ></IssuerList>
                     <span className="adyen-checkout-disclaimer__label">
                         <LabelOnlyDisclaimerMessage message={i18n.get('paybybankpix.issuerList.disclaimer')} urls={[TERMS_CONDITIONS_URL]} />
--- a/packages/lib/src/components/internal/IssuerList/IssuerList.tsx
+++ b/packages/lib/src/components/internal/IssuerList/IssuerList.tsx
@@ -100,7 +100,7 @@ function IssuerList({
                 props.onSubmitAnalytics(event);
             }
         },
-        [props.type]
+        [props.type, props.onSubmitAnalytics]
     );
 
     const debounceSearchAnalytics = useRef(debounce(props.onSubmitAnalytics, ANALYTICS_SEARCH_DEBOUNCE_TIME));
--- a/packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
+++ b/packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
@@ -47,8 +47,8 @@ export class AnalyticsLogEvent extends AbstractAnalyticsEvent {
         this.type = props.type;
         this.message = props.message;
 
-        if (props.subType !== undefined) this.subType = props.subType;
-        if (props.result !== undefined) this.result = props.result;
+        if (props.subType) this.subType = props.subType;
+        if (props.result) this.result = props.result;
     }
 
     public getEventCategory(): string {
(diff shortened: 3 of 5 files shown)


packages/lib/src/components/Card/components/Fastlane/FastlaneSignup.test.tsx
```diff
@@ -170,7 +204,13 @@ test('should not render the UI if there are missing configuration fields', () =>
     jest.spyOn(console, 'warn').mockImplementation(consoleMock);
 
     const { container } = customRender(
-        <FastlaneSignup {...fastlaneConfiguration} onChange={onChangeMock} currentDetectedBrand="visa" onSubmitAnalytics={jest.fn()} />
+        <FastlaneSignup
+            {...fastlaneConfiguration}
+            onChange={onChangeMock}
+            currentDetectedBrand="visa"
+            onSubmitAnalytics={jest.fn()}
+            type={'scheme'}
```
johnayeni, line 212: I see that we use this method `propName={'string'}` and `propName="string"` for string props. Its a small thing but would be nice to keep things consistent. Maybe nice to have  prettier rule that can automatically fix this.

packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
```diff
@@ -0,0 +1,74 @@
...
+    private readonly message?: string;
+    private readonly result?: string;
+
+    constructor(props: AnalyticsLogEventProps) {
+        super(props.component);
+
+        this.type = props.type;
+        this.message = props.message;
+
+        if (props.subType !== undefined) this.subType = props.subType;
```
johnayeni: Do we want to check for only undefined here or all falsy values e.g 

```
if (!!props.subType)
```

packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (props.result !== undefined) this.result = props.result;
```
johnayeni: Do we want to check for only undefined here or all falsy values e.g 

```
if (!!props.result)
```

packages/lib/src/components/internal/IssuerList/IssuerList.tsx
```diff
@@ -93,23 +86,27 @@ function IssuerList({
...
+            if (isOpen) {
+                const event = new AnalyticsInfoEvent({
+                    component: props.type,
+                    type: InfoEventType.displayed,
+                    target: UiTarget.list
+                });
+                props.onSubmitAnalytics(event);
+            }
+        },
+        [props.type]
```
longyulongyu: The `props.onSubmitAnalytics` is missing in its dependency array.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -136,8 +135,9 @@ function QRLoader(props: QRLoaderProps) {
 
     const handleCopy = () => {
         const event = new AnalyticsInfoEvent({
-            type: ANALYTICS_DOWNLOAD_STR,
-            target: ANALYTICS_QR_CODE_DOWNLOAD
+            component: props.type,
+            type: InfoEventType.download,
```
longyulongyu, line 139: I am wondering if this is a historical overlooked issue. I think the target and type should be copy button related?

packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
```diff
@@ -92,6 +92,7 @@ function Enrollment(props: EnrollmentProps) {
                         payButton={props.payButton}
                         showPayButton={true}
                         ref={issuerListRef}
+                        type={'paybybank_pix'}
```
longyulongyu: Maybe use `TxVariants.paybybank_pix` or pass it to the prop?

packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public getEventCategory(): string {
+        return ANALYTICS_EVENT.log;
+    }
+
+    public static getSubtypeFromActionType(type: PaymentAction['type']): LogEventSubtype {
+        switch (type) {
```
longyulongyu: We can just return `LogEventSubtype[type]`?

longyulongyu (approved):

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
ribeiroguilherme: That is a good question tbh. I will keep as it is for now, but I will investigate if we have to update this event

packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
ribeiroguilherme: Oh yeah, good one!

packages/lib/src/components/PayByBankPix/components/Enrollment/Enrollment.tsx
ribeiroguilherme: Yup

packages/lib/src/components/internal/IssuerList/IssuerList.tsx
ribeiroguilherme: Done

packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
ribeiroguilherme: For majority of props it is ok to do `if (props.subtype)` , but there are few that are boolean values . For those, I will keep the 'undefined' only check

packages/lib/src/components/Card/components/Fastlane/FastlaneSignup.test.tsx
ribeiroguilherme: There is this rule which we could use: https://eslint.style/rules/jsx-curly-brace-presence

PR #3649, opened 2025-11-14
Fix: adds missing error messages for UPI and CtP

m1aw:
## 📋 Pull Request Checklist

- [X] I have added or updated Storybook stories where applicable.
- [X] I have tested the changes manually in the local environment.
- [X] All interfaces and types introduced or updated are strictly typed.

---

## 📝 Summary

Fixes missing error message text for UPI ID and Click to Pay OTP. They now both use the generic getMessage. I have also updated the getMessage to allow to not change the text to lowercase.

---

## 🧪 Tested scenarios

---

**Closes**:   COSDK-742

---

### DIFF
--- a/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
+++ b/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
@@ -101,18 +101,18 @@ const CtPOneTimePasswordInput = (props: CtPOneTimePasswordInputProps): h.JSX.Ele
     }, [data, valid, errors]);
 
 
-    const errorMessageLogic = () => {
-        if(!isOtpFielDirty) return null;
+    const getOtpErrorMessage = useCallback(() => {
+        if (!isOtpFielDirty) return null;
+
         return resendOtpError || props.errorMessage || getErrorMessage(i18n, errors.otp, i18n.get('ctp.otp.fieldLabel'));
-    }
-    const errorMessage = errorMessageLogic(); 
+    }, [isOtpFielDirty, resendOtpError, props.errorMessage, errors.otp, i18n]);
 
     return (
         <div className={'adyen-checkout-ctp__otp-field-wrapper'}>
             <Field
                 name="oneTimePassword"
                 label={i18n.get('ctp.otp.fieldLabel')}
-                errorMessage={errorMessage}
+                errorMessage={getOtpErrorMessage()}
                 classNameModifiers={['otp']}
             >
                 <InputText
--- a/packages/lib/src/utils/getErrorMessage.ts
+++ b/packages/lib/src/utils/getErrorMessage.ts
@@ -1,11 +1,11 @@
 import { ERROR_FIELD_INVALID, ERROR_FIELD_REQUIRED } from '../core/Errors/constants';
 
-export const getErrorMessage = (i18n, error, label?: string, lowercaseLabel: boolean = true): string | boolean => {
+export const getErrorMessage = (i18n, error, label?: string, lowerCaseLabel: boolean = true): string | boolean => {
     if (error?.errorMessage) {
         const errorKey = error.errorMessage;
         const shouldBuildErrorMessage = [ERROR_FIELD_REQUIRED, ERROR_FIELD_INVALID].includes(errorKey);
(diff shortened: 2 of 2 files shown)


packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
```diff
@@ -99,12 +100,19 @@ const CtPOneTimePasswordInput = (props: CtPOneTimePasswordInputProps): h.JSX.Ele
         props.onChange({ data, valid, errors, isValid });
     }, [data, valid, errors]);
 
+
+    const errorMessageLogic = () => {
```
johnayeni: Should we be having the term 'logic' in function names? A function should already be logic no?n

Maybe we can just use `getErrorMessage`

packages/lib/src/utils/getErrorMessage.ts
```diff
@@ -1,11 +1,11 @@
 import { ERROR_FIELD_INVALID, ERROR_FIELD_REQUIRED } from '../core/Errors/constants';
 
-export const getErrorMessage = (i18n, error, label?: string): string | boolean => {
+export const getErrorMessage = (i18n, error, label?: string, lowercaseLabel: boolean = true): string | boolean => {
```
johnayeni:
```suggestion
export const getErrorMessage = (i18n, error, label?: string, lowerCaseLabel: boolean = true): string | boolean => {
```

Because JS already uses `toLowerCase`. This is just a personal peference

packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
```diff
@@ -99,12 +100,19 @@ const CtPOneTimePasswordInput = (props: CtPOneTimePasswordInputProps): h.JSX.Ele
...
+        if(!isOtpFielDirty) return null;
+        return resendOtpError || props.errorMessage || getErrorMessage(i18n, errors.otp, i18n.get('ctp.otp.fieldLabel'));
+    }
+    const errorMessage = errorMessageLogic();
```
ScottiBR: can it be a simple const without the function, since this function is not going to be reused?

```js
const errorMessage = isOtpFielDirty ? resendOtpError || props.errorMessage || getErrorMessage(i18n, errors.otp, i18n.get('ctp.otp.fieldLabel')) : null
```

m1aw: That was exactly what I was trying to avoid, I really don't like long ternary logic, it's way hard to read in my opinion, but yeah it could be. I just don't like it.

packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
m1aw: I agree, I honestly wanted to avoid using a function entirely but I know people don't like self-calling anonymous functions, and `getErrorMesssage` was taken any clever alternatives?

packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPOneTimePasswordInput.tsx
```diff
@@ -99,12 +100,19 @@ const CtPOneTimePasswordInput = (props: CtPOneTimePasswordInputProps): h.JSX.Ele
         props.onChange({ data, valid, errors, isValid });
     }, [data, valid, errors]);
 
+
+    const getOtpErrorMessage = useCallback(() => {
+        if (!isOtpFielDirty) return null;
+
+        return resendOtpError || props.errorMessage || getErrorMessage(i18n, errors.otp, i18n.get('ctp.otp.fieldLabel'));
+    }, [isOtpFielDirty, resendOtpError, props.errorMessage, errors.otp, i18n]);
```
ScottiBR, line 108: Nice, this way is better than my previous idea, i agree with you

johnayeni (approved):

PR #3650, opened 2025-11-14
Fix: adds roles to CtP OTP button

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

This attempts to fix the missing live regions for the OTP button / content that informs how much time users have.

- `role=alert` for the "Code resent" message;
- `role=timer` for the "Timer" part

Timer and alert are in the suggestion from level access:
> Indicate live regions for dynamically changing content. Live regions can be created by adding a role attribute set to "log", "status", "alert", "progressbar", "marquee", or "timer" as appropriate. Alternatively, custom behavior can be created by using the aria-live, aria-atomic, and aria-relevant attributes. Text injected into this live region element will be announced by screen readers.

Also additional documentation to check the support of these features:
- https://useragentman.com/enable/timer.php
- https://useragentman.com/enable/alert.php
- https://a11ysupport.io/tech/aria/alert_role

This is quite a different approach from what we use for the SRPanel, but it's quite simpler and if given the greenlight by level access could simplify our Timer logic too.

---

## 🧪 Tested scenarios

---

**Closes**:  COSDK-748, CW-31

---

### DIFF
--- a/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPResendOtpLink.tsx
+++ b/packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPResendOtpLink.tsx
@@ -67,7 +67,7 @@ const CtPResendOtpLink = ({ onError, onResendCode, disabled }: CtPResendOtpLinkP
 
     if (showConfirmation) {
         return (
-            <div className="adyen-checkout-ctp__otp-resend-code--confirmation">
+            <div role="alert" className="adyen-checkout-ctp__otp-resend-code--confirmation">
                 {i18n.get('ctp.otp.codeResent')}
                 <Icon type={`${PREFIX}checkmark_black`} height={14} width={14} />
             </div>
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/internal/ClickToPay/components/CtPOneTimePassword/CtPOneTimePasswordInput/CtPResendOtpLink.tsx
```diff
@@ -76,7 +76,7 @@ const CtPResendOtpLink = ({ onError, onResendCode, disabled }: CtPResendOtpLinkP
 
     if (counter > 0) {
         return (
-            <div className="adyen-checkout-ctp__otp-resend-code--disabled">
+            <div role="timer" className="adyen-checkout-ctp__otp-resend-code--disabled">
```
ribeiroguilherme, line 79: Does it work as is? I tried testing it but the VoiceOver didn't announce the timer nor the 'code resent' text.  Also, isn't `aria-live` needed for this?

m1aw: Yeah, we think this is more a VoiceOver / Mac problem and works on other screen readers, we are going to ask LevelAccess what they think about it.

PR #3663, opened 2025-11-18
Fix: Click to Pay accessibility label

ScottiBR:
## 📋 Pull Request Checklist

---

## 📝 Summary

Improves accessibility for the Click to Pay card selector for screen readers. Removed redundant subtitles for single/empty card states while preserving spacing.

### accessibility
- Added programmatic <label> association to the card selector dropdown using Field component's label prop and allowIdOnButton={true}
- Impact: Screen readers now announce "Select a card to use" when focusing the dropdown. Voice control users can activate it by name.

### Layout: Removed Redundant Subtitle (Aligned with Arjen)
- Empty Card List: No subtitle (doesn't make sense without cards)
- Single Card: No subtitle (no selection needed)
- Multiple cards: Subtitle rendered as accessible label for dropdown
Consistent spacer-070 spacing maintained across all scenarios via CSS.

### Before (Single Card):
O I@VISA

### After  (Single Card):
Complete payment with Click to Pay

### multiple cards
image

---

## 🧪 Tested scenarios

- Updated button queries in tests to use accessible label name, 

---

**Closes**:  COSDK-745

---

### DIFF
--- a/.changeset/loud-areas-leave.md
+++ b/.changeset/loud-areas-leave.md
@@ -1,5 +1,0 @@
----
-'@adyen/adyen-web': minor
----
-
-FIX: Click to Pay accessibility labels and subtitle logic Added programmatic labels to the card selector for screen readers. Removed redundant subtitles for single/empty card states while preserving spacing.


johnayeni (approved):
🚀

.changeset/tall-eyes-sniff.md
```diff
@@ -0,0 +1,6 @@
+---
+'@adyen/adyen-web': patch
+---
+
+- Fixed Click to Pay accessibility label at the card selector for screen readers.
```
johnayeni: One small thing here is that I think we try to keep messages in this change set files to one line because the formatting in the release branch ends up being weird

@ribeiroguilherme when we have multiple changes for a PR should we generate two changesets then?

ScottiBR: Changeset updated

ribeiroguilherme (approved):

PR #3666, opened 2025-11-20
fix: cancel script load handler when this.script is null

fabiolmorato:
## 📋 Pull Request Checklist

---

## 📝 Summary

We noticed recently this error in our client error reporter: `ErrorEvent: Cannot read properties of null (reading 'setAttribute')`.

In our specific usage, we mount the script on a sidepane and unmount it when user closes the sidepane. Unmounting the script causes `removeScript` to be called setting `this.script` to `null`. So these events happen in this order:
1. Sidepane opens
2. Script gets mounted (not loaded yet)
3. User closes sidepane
4. Script is unmounted
5. `removeScript` is called setting `this.script` to `null`
6. Sidepane is unmounted
7. Load event is fired
8. Load event handler tries to run `this.script.setAttribute('data-script-loaded', 'true');` but `this.script` is `null`
9. Error is thrown

---

## 🧪 Tested scenarios

Added a test case that simulates the script being mounted, unmounted before load event, asserting the promise still rejects and the dispatch of the load event does not cause an error to be thrown.

---

https://github.com/Adyen/adyen-web/issues/3681

**Closes**:  

#3681 

---


ScottiBR (changes requested):
Nice catch on the guard clause! While `if(!this.script)` definitely stops the crash, isn't it technically just a band-aid? 

The real root cause seems to be that our load and error listeners are still sticking around even after we verify the script is removed.

If we just ensure `cleanupListeners()` is called right alongside `script.remove()`, we wouldn't even need this check because `handleOnLoad` would never fire on a dead component.

fabiolmorato:
@ScottiBR fala Guilherme!

Thanks for taking the time to check this out. While I do agree the initial solution was a band-aid to the problem, I didn't want to introduce too many changes to the code. I updated the pull request in accordance to the suggestion

ScottiBR (approved):
Fala Fabio, Well Done!
 I'll cherryPick your commit and this fix will be available at our next release.

ScottiBR:
https://github.com/Adyen/adyen-web/pull/3709

PR #3669, opened 2025-11-21
UIlement: Move CoreProvider to UIElement

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

Right now every component has to import `CoreProvider` and pass the props for all the modules which it needs. This is quite repetitive and usually changes to the `CoreProvider` means we have to change multiple files. Also it is very hard to have logic across UI elements without needing to change all the extensions of `UIElement`. Below is an architectural view of how the `CoreProvider` is setup in the SDK

image

**Cons of the current architecture**

A lot of repeated code on the SDK for using the core provider which we need to keep adding when working on new payment methods

Add a new module or prop to the core provider would require a lot of file changes, slowing down these kind of developments e.g [this PR for adding analytics module to core provider](https://github.com/Adyen/adyen-web/pull/3512)

Suggested Solution

A solution for this would be to render a jsx in the `UIElement` class with a `render` method which all component inherit from and move the `CoreProvider` there. Each component that inherits from `UIElement` would change the signature of their `render` method to `componentToRender` and remove the `CoreProvider`  in those methods.

image

**Why are we doing we doing this?**

There a bunch of benefits of switching to this architecture
- This would help reduce the amount of effort we need to make changes to providers or modules in the future. Since this would be done at a central point, this would reduce potential for introducing bugs and makes the codebase easier to understand
- Slightly smaller bundle size. I don't expect the savings here to be significant but anything that reduces bundle size is always welcome.
- Since UI element would now also include a render function, having effects or preact specific logic that we want shared across all payment methods components would be easier to do.

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,12 +1,13 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-    "extends": ["config:base", ":disableDependencyDashboard"],
+    "extends": ["config:recommended", ":disableDependencyDashboard"],
     "labels": ["dependencies"],
     "minimumReleaseAge": "21 days",
     "internalChecksFilter": "strict",
     "rebaseWhen": "behind-base-branch",
     "prConcurrentLimit": 5,
     "updateNotScheduled": false,
+    "ignorePaths": ["packages/playground/**"],
     "packageRules": [
         {
             "matchDatasources": ["npm"],
@@ -14,11 +15,13 @@
             "groupName": "Minor and Patch Updates"
         },
         {
-            "matchPackageNames": ["@adyen/**", "node", "cheerio"],
+            "matchPackageNames": ["@adyen/**", "cheerio"],
             "enabled": false
         },
         {
+            "description": "Disable node-version updates",
             "matchDatasources": ["node-version"],
+            "matchPackageNames": ["actions/node-versions"],
             "enabled": false
         },
         {


packages/lib/src/components/ThreeDS2/ThreeDS2DeviceFingerprint.tsx
```diff
@@ -48,7 +48,7 @@ class ThreeDS2DeviceFingerprint extends UIElement<ThreeDS2DeviceFingerprintConfi
         this.unmount(); // re. fixing issue around back to back fingerprinting calls
     }
 
-    render() {
+    protected override componentToRender(): h.JSX.Element {
```
longyulongyu, line 51: It seems this ui element doesn't need core provider, I think we can fully override it.
```js
public override render() {}
```

johnayeni: Fixed

packages/lib/src/components/ApplePay/ApplePay.tsx
```diff
@@ -350,7 +350,11 @@ class ApplePayElement extends UIElement<ApplePayConfiguration> {
         }
     }
 
-    render() {
+    protected componentToRender(): h.JSX.Element {
+        return null;
```
ribeiroguilherme: Why not wrap it too? I guess it is fine even if it does not use CoreProvider.  At least we keep it consistent.

packages/lib/src/components/GooglePay/GooglePay.tsx
```diff
@@ -316,7 +316,11 @@ class GooglePay extends UIElement<GooglePayConfiguration> {
         return this.props.icon ?? this.resources.getImage()('googlepay');
     }
 
-    render() {
+    protected componentToRender(): h.JSX.Element {
```
ribeiroguilherme: Same comment as ApplePay

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -605,6 +606,18 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
                 });
             });
     }
+
+    protected componentToRender(): h.JSX.Element | Error {
```
ribeiroguilherme: I guess this can be marked as `abstract` , so we are always forced to supply its implementation. That is a minor thing though

packages/lib/src/components/ThreeDS2/ThreeDS2DeviceFingerprint.tsx
```diff
@@ -48,7 +48,11 @@ class ThreeDS2DeviceFingerprint extends UIElement<ThreeDS2DeviceFingerprintConfi
         this.unmount(); // re. fixing issue around back to back fingerprinting calls
     }
 
-    render() {
+    protected componentToRender(): h.JSX.Element {
```
ribeiroguilherme: Same as the Apple Pay comment!

packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -209,21 +208,19 @@ class DropinElement extends UIElement<DropinConfiguration> implements IDropin {
...
+                <DropinComponent
+                    {...this.props}
+                    core={this.core}
+                    elementRef={this.elementRef}
+                    onCreateElements={this.handleCreate}
+                    ref={dropinRef => {
+                        this.dropinRef = dropinRef;
+                    }}
+                />
+            </SRPanelProvider>
```
ribeiroguilherme, line 223: Some thoughts that are related to this Provider wrapping, but don't have to be addressed here:

- Drop-in is wrapped in the CoreProvider, and all its payment methods being re-rendered are also wrapped in the CoreProvider. It does not seem to cause any harm (it always worked like that), but it does not feel that it is the correct way. Maybe it does make sense to make sure that there is a single CoreProvider when using Drop-in and its child components just re-use it.
- We used to have preact devtools extension, so we could spot what I just mentioned above in an easier way. Since we move to ES modules (on v6), we couldn't make it work with the repository anymore. Perhaps it is worth taking a look and see if we can bring that back, so we can easily spot those kind of scenarios

packages/lib/src/components/ApplePay/ApplePay.tsx
johnayeni: Done

packages/lib/src/components/GooglePay/GooglePay.tsx
johnayeni: Done

packages/lib/src/components/ThreeDS2/ThreeDS2DeviceFingerprint.tsx
johnayeni: Done

packages/lib/src/components/Dropin/Dropin.tsx
johnayeni: I have added a change here that basically checks in the UI element if there is already a core context in scope. If one exists, then it renders the component without the Core provider. You can see the difference below from the preact dev tools (this already works by the way)

Before
Screenshot 2025-11-24 at 14 05 59


After
Screenshot 2025-11-24 at 15 46 21


I am not fully certain if this is the best approach but you can let me know what you think.

packages/lib/src/components/internal/UIElement/UIElement.tsx
johnayeni: Done

ribeiroguilherme (approved):

PR #3675, opened 2025-11-24
Analytics - Converted module to class based

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

We are moving the attempt ID request to be earlier in the flow. In order to keep capturing the 'flavor' , we need to send another request to report that eventually. 

As we have to touch the Analytics module to support this behavior accordingly, I decided to refactor the Analytics module to class based since it might be the best opportunity to do it, before actually adding new features to it.

Core changes:
- Refactored `Analytics.ts` to be class based. 
- Adjusted `core.ts` to request the ID as soon as the analytics module is created
- Removed the Analytics setup part from the `UIEelement`
- Removed `experiments` and `payload` property configurations from the Analytics (these values were used in the previous analytics and not in the current one)

Next PR:
- Adjust `Analytics.ts` to report the flavor

Disclaimer: There are quite some files changes, but majority is just renaming the type `AnalyticsModule` to `IAnalytics`.

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/core/Analytics/Analytics.ts
+++ b/packages/lib/src/core/Analytics/Analytics.ts
@@ -1,144 +1,135 @@
-import CollectId from '../Services/analytics/collect-id';
+import CollectId, { AttemptIdPayload } from '../Services/analytics/collect-id';
 import EventsQueue, { EventsQueueModule } from './EventsQueue';
-import { AnalyticsEventCategory, AnalyticsInitialEvent, AnalyticsProps } from './types';
-import { ANALYTIC_LEVEL, ANALYTICS_INFO_TIMER_INTERVAL, ANALYTICS_PATH, ANALYTICS_EVENT } from './constants';
+import { ANALYTICS_PATH } from './constants';
 import { debounce } from '../../utils/debounce';
-import { AnalyticsModule } from '../../types/global-types';
 import { processAnalyticsData } from './utils';
 import { AbstractAnalyticsEvent } from './events/AbstractAnalyticsEvent';
 import Storage from '../../utils/Storage';
-import { CheckoutAttemptIdSession } from '../Services/analytics/types';
-
-let capturedCheckoutAttemptId: string | null = null;
-let sendEventsTimerId = null;
+import type { CheckoutAttemptIdSessionStorage } from '../Services/analytics/types';
+import type { AnalyticsOptions } from './types';
 
 /**
  * If the checkout attempt ID was stored more than fifteen minutes ago, then we should request a new ID.
  * More here: COWEB-1099
  */
-function isSessionCreatedUnderFifteenMinutes(session: CheckoutAttemptIdSession): boolean {
+function isSessionCreatedUnderFifteenMinutes(session: CheckoutAttemptIdSessionStorage): boolean {
     const FIFTEEN_MINUTES_IN_MS = 15 * 60 * 1000;
     if (!session?.timestamp) return false;
     const fifteenMinutesAgo = Date.now() - FIFTEEN_MINUTES_IN_MS;
     return session.timestamp > fifteenMinutesAgo;
 }
 
-const Analytics = ({ locale, clientKey, analytics, analyticsContext }: AnalyticsProps): AnalyticsModule => {
-    const defaultProps = {
-        enabled: true
-    };
+const ANALYTICS_INFO_TIMER_INTERVAL = process.env.NODE_ENV === 'development' ? 5000 : 10000;
 
-    const props = { ...defaultProps, ...analytics };
+export interface AnalyticsProps {
+    locale: string;
+    clientKey: string;
+    analyticsContext: string;
+    enabled?: boolean;
+    analyticsData?: AnalyticsOptions['analyticsData'];
+}
 
-    const level = props.enabled ? ANALYTIC_LEVEL.all : ANALYTIC_LEVEL.initial;
+export interface IAnalytics {
+    setUp({ sessionId, checkoutStage }: { sessionId?: string; checkoutStage?: 'precheckout' | 'checkout' }): Promise<void>;
+    flush(): void;
+    sendAnalytics(event: AbstractAnalyticsEvent): void;
+    getEventsQueue(): EventsQueueModule;
+}
 
-    const collectId = CollectId({
-        analyticsContext,
-        clientKey,
-        locale,
-        analyticsPath: ANALYTICS_PATH
-    });
+class Analytics implements IAnalytics {
+    private readonly enabled: boolean = true;
+    private readonly analyticsData?: AnalyticsOptions['analyticsData'];
 
-    const storage = new Storage<CheckoutAttemptIdSession>('checkout-attempt-id', 'sessionStorage');
-    const eventsQueue: EventsQueueModule = EventsQueue({ analyticsContext, clientKey, analyticsPath: ANALYTICS_PATH });
+    private readonly eventsQueue: EventsQueueModule;
+    private readonly storage = new Storage<CheckoutAttemptIdSessionStorage>('checkout-attempt-id', 'sessionStorage');
+    private readonly collectId: (attemptIdPayload: AttemptIdPayload) => Promise<string>;
+    private readonly debouncedSendEventsFunction: () => void;
 
-    const sendAnalyticsEvents = () => {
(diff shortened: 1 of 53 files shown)


packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,144 +1,135 @@
...
+            };
+
+            this.capturedCheckoutAttemptId = await this.collectId(attemptIdPayload);
+
+            this.storage.set({
+                id: this.capturedCheckoutAttemptId,
+                timestamp: isSessionReusable ? checkoutAttemptIdSession.timestamp : Date.now()
+            });
+        } catch (error: unknown) {
+            console.warn(error);
```
ScottiBR, line 89: Maybe we should improve the error handle here, adding more context to the error message like:
- Merchant enabled events `this.enabled`
- checkoutAttemptId

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,144 +1,135 @@
...
-         */
-        if (eventCat === ANALYTICS_EVENT.log || eventCat === ANALYTICS_EVENT.error) {
-            clearTimeout(sendEventsTimerId); // clear any timer that might be about to dispatch the info events array
-
-            debounce(sendAnalyticsEvents)();
+    public sendAnalytics(event: AbstractAnalyticsEvent): void {
+        if (!this.enabled) {
+            return;
+        }
+        this.addEventToQueue(event);
```
ScottiBR, line 97: I suggest adding a try catch block here to ensures that if event processing fails (queue issues, timer problems), the payment process or any other process continues uninterrupted.

ScottiBR (changes requested):

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,144 +1,135 @@
...
+     * Info events don't have high priority, therefore  we add a delay in order to dispatch a batch of events
+     * Log/Error events are sent almost immediately. There is a debounce mechanism in place but the delay is very minimal
+     *
+     * @param event
+     * @private
+     */
+    private addEventToQueue(event: AbstractAnalyticsEvent): void {
+        this.eventsQueue.add(event);
+
+        if (event.getEventCategory() === 'info') {
```
longyulongyu, line 118: Very minor and optional: I think we could have a `debouncedSendInfoEvents` in the constructor so that we don't need to manually manage the timeout id here. 
```js
// constructor
this.debouncedSendInfoEvents = debounce(this.sendEvents.bind(this), ANALYTICS_INFO_TIMER_INTERVAL);
```

```js
    private addEventToQueue(event: AbstractAnalyticsEvent): void {
        this.eventsQueue.add(event);

        if (event.getEventCategory() === 'info') {
            this.debouncedSendInfoEvents();
        } else {
            this.debouncedSendEventsFunction();
        }
    }

```

ribeiroguilherme: Oh yeah, that is very smart approach. I will do that in the next PR

packages/lib/src/core/Analytics/Analytics.ts
ribeiroguilherme: I will add these changes in the follow up PR which will have big changes on this part and in the other you mentioned above. I will tag you on that

packages/lib/src/core/Analytics/events/AbstractAnalyticsEvent.ts
```diff
@@ -10,7 +10,7 @@ export abstract class AbstractAnalyticsEvent {
      */
     private readonly component: string;
 
-    public abstract getEventCategory(): string;
+    public abstract getEventCategory(): 'info' | 'log' | 'error';
```
johnayeni, line 13: We could add a type for this event category here

`type AnalyticsEventCategory = 'info' | 'log' | 'error'`

packages/lib/src/core/Analytics/events/AnalyticsErrorEvent.ts
```diff
@@ -67,7 +66,7 @@ export class AnalyticsErrorEvent extends AbstractAnalyticsEvent {
         if (props.message) this.message = props.message;
     }
 
-    public getEventCategory(): string {
-        return ANALYTICS_EVENT.error;
+    public getEventCategory(): 'error' {
```
johnayeni, line 69: Here we could extract the error option from the union type 

```suggestion
    public getEventCategory(): Extract<AnalyticsEventCategory "error"> {
```

Related to this comment https://github.com/Adyen/adyen-web/pull/3675/files#r2564924803

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
```diff
@@ -174,7 +173,7 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
         }
     }
 
-    public getEventCategory(): string {
-        return ANALYTICS_EVENT.info;
+    public getEventCategory(): 'info' {
```
johnayeni, line 176: Same comment as here https://github.com/Adyen/adyen-web/pull/3675/files#r2564931145

packages/lib/src/core/Analytics/events/AnalyticsLogEvent.ts
```diff
@@ -51,8 +50,8 @@ export class AnalyticsLogEvent extends AbstractAnalyticsEvent {
         if (props.result) this.result = props.result;
     }
 
-    public getEventCategory(): string {
-        return ANALYTICS_EVENT.log;
+    public getEventCategory(): 'log' {
```
johnayeni, line 53: Same comment as here https://github.com/Adyen/adyen-web/pull/3675/files#r2564931145

packages/lib/src/core/Analytics/events/AbstractAnalyticsEvent.ts
johnayeni: It also maybe makes sense to an enum type. Makes it easier to refactor since we can just target all of the enum values rather than `error` or `info` strings in code which can be quite common values.

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -1,144 +1,135 @@
...
 
-const Analytics = ({ locale, clientKey, analytics, analyticsContext }: AnalyticsProps): AnalyticsModule => {
-    const defaultProps = {
-        enabled: true
-    };
-
-    const props = { ...defaultProps, ...analytics };
-
-    const level = props.enabled ? ANALYTIC_LEVEL.all : ANALYTIC_LEVEL.initial;
+const ANALYTICS_INFO_TIMER_INTERVAL = process.env.NODE_ENV === 'development' ? 5000 : 10000;
```
johnayeni, line 22:
```suggestion
const ANALYTICS_INFO_TIMER_INTERVAL = process.env.NODE_ENV === 'development' ? 5_000 : 10_000;
```

Just a preference here, but I find that numberic seperators on large numbers make them easier to read

packages/lib/src/core/Analytics/events/AbstractAnalyticsEvent.ts
ribeiroguilherme: Sounds good. I will add that in the next PR

packages/lib/src/core/Analytics/events/AnalyticsErrorEvent.ts
ribeiroguilherme: Sounds good. I will add that in the next PR

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
ribeiroguilherme: Sounds good. I will add that in the next PR

packages/lib/src/core/Analytics/Analytics.ts
ribeiroguilherme: Sounds good. I will add that in the next PR

johnayeni (approved):
Thanks @ribeiroguilherme. Approving now since you will be adding the suggestions in a follow up PR.

PR #3689, opened 2025-11-27
Analytics - Adding AnalyticsService to perform async tasks

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

- Added `AnalyticsService` which performs the async requests related to analytics.
- Refactored EventQueue to be just a EventQueue and not perform any async requests from within it. Requests now are done from the `AnalyticsService` 
- Components now report their flavor once they are instantiated. If Drop-in is instantiated, it will report the flavor 'dropin' . Remaining components will report 'components'. The flavor can be reported a single time only.
- Added enum for the analytics event category
- Wrapped methods in try-catch block to assure exceptions will be caught

---

## 🧪 Tested scenarios
- Added tests for core.ts doing initial setup call
- Added tests for the payment method components reporting the flavor
- Added tests for the event queue and analytics service class
- Adjusted tests for the analytics module

### DIFF
--- a/packages/lib/src/core/Services/http.ts
+++ b/packages/lib/src/core/Services/http.ts
@@ -34,7 +34,7 @@ function isAdyenApiErrorResponse(data: any): data is AdyenApiErrorResponse {
     return data && data.errorCode && data.errorType && data.message && data.status;
 }
 
-export function http<T>(options: HttpOptions, data?: any): Promise<T> {
+export function http<T>(options: HttpOptions, payload?: any): Promise<T> {
     const {
         headers = [],
         errorLevel = 'warn',
@@ -58,7 +58,7 @@ export function http<T>(options: HttpOptions, data?: any): Promise<T> {
         redirect: 'follow',
         referrerPolicy: 'no-referrer-when-downgrade',
         ...(AbortSignal?.timeout && { signal: AbortSignal?.timeout(timeout) }),
-        ...(data && { body: JSON.stringify(data) })
+        ...(payload && { body: JSON.stringify(payload) })
     };
 
     const url = `${loadingContext}${path}`;
@@ -71,8 +71,9 @@ export function http<T>(options: HttpOptions, data?: any): Promise<T> {
                     return;
                 }
 
+                const data = await response.json();
+
                 if (response.ok) {
-                    const data = await response.json();
                     return data;
                 }
 


packages/lib/src/core/Services/http.ts
```diff
@@ -66,9 +66,13 @@ export function http<T>(options: HttpOptions, data?: any): Promise<T> {
     return (
         fetch(url, request)
             .then(async response => {
-                const data = await response.json();
+                // Handle empty responses (e.g., 204 No Content)
+                if (response.status === 204) {
+                    return;
+                }
```
ScottiBR, line 72: Since this is an heavily used service, should we create a specific unit test for 204 scenarios?

packages/lib/src/core/Services/http.ts
```diff
@@ -66,9 +66,13 @@ export function http<T>(options: HttpOptions, data?: any): Promise<T> {
...
 
                 if (response.ok) {
+                    const data = await response.json();
```
ScottiBR: There is any case that `response.ok` will be falsy? Then `data` variable will not exists at line 79 `isAdyenApiErrorResponse(data) `

packages/lib/src/core/Analytics/AnalyticsService.test.ts
```diff
@@ -0,0 +1,156 @@
...
+import AdyenCheckoutError from '../Errors/AdyenCheckoutError';
+import { AnalyticsInfoEvent, InfoEventType } from './events/AnalyticsInfoEvent';
+
+jest.mock('../Services/http', () => ({
+    httpPost: jest.fn()
+}));
+
+const mockHttpPost = httpPost as jest.MockedFunction<typeof httpPost>;
+
+describe('AnalyticsService', () => {
```
ScottiBR, line 12: maybe add one test regarding 204 response? like `should handle 204 response in analytics service integration` and mock an 204 http result for sendEvents()

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -107,6 +108,10 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
         this.analytics.sendAnalytics(event);
     }
 
+    protected reportIntegrationFlavor(): void {
+        void this.analytics.sendFlavor('components');
```
ScottiBR, line 112: Do we need to worry about concurrent `sendFlavor()` calls if multiple components spin up at once? I'm wondering if that might trip up the initialization (setUp) or queueing logic.

packages/lib/src/core/Services/http.ts
ribeiroguilherme: There is already one covering that

packages/lib/src/core/Services/http.ts
ribeiroguilherme: Oh, good catch.  There is `data` object defined on line `37` which made the linter not complain about this. I will fix that

packages/lib/src/core/Analytics/AnalyticsService.test.ts
ribeiroguilherme: I believe it does not add value doing this since the `http` code is mocked and isolated from this test file.  `204` responses are being tested in the other file, and the `sendEvents()` do not expect any data back.

packages/lib/src/components/internal/UIElement/UIElement.tsx
ribeiroguilherme: Good point - concurrent calls might happen indeed, but the Analytics module handles that and just accept the first flavor report and ignore the next ones.

PR #3696, opened 2025-12-01
Fix aria-live announcement at Button component

ScottiBR:
## 📋 Pull Request Checklist

---

## 📝 Summary

When forms were submitted without moving focus to the Button (e.g., via keyboard shortcuts, programmatic submission, or pressing Enter in a focused field), screen readers received no announcement that the action entered a loading state. Because the loading feedback was only rendered within the button’s visible content, assistive technologies missed it when the button wasn’t focused, resulting in a silent wait and poor user feedback. 

image

Adding a dedicated, visually hidden live region (role="status" with aria-live="polite") ensures loading and redirect updates are announced regardless of focus, so users relying on screen readers are informed immediately after submission.

---

### Click To Pay, PayTo & CoBranded Cards
- Fill the one time code field, navigate to the continue button and press enter to activate.
- Enable a screen reader.
- Using the Down Arrow key or Tab key, navigate to the dynamically updating content. Trigger it with the Enter key if necessary.
- Review any announcements given by the screen reader in response to the dynamic content update.

---

[COSDK-785](https://youtrack.is.adyen.com/issue/COSDK-785/Ensure-that-loading-status-messages-can-be-determined-programmatically-without-receiving-focus)

**Closes**: COSDK-785:

---

### DIFF
--- a/packages/lib/src/components/internal/Button/Button.tsx
+++ b/packages/lib/src/components/internal/Button/Button.tsx
@@ -1,4 +1,4 @@
-import { Component, h } from 'preact';
+import { Component, h, Fragment } from 'preact';
 import classNames from 'classnames';
 import Spinner from '../Spinner';
 import { useCoreContext } from '../../../core/Context/CoreProvider';
@@ -28,6 +28,15 @@ class Button extends Component<ButtonProps, ButtonState> {
         setTimeout(() => {
             this.setState({ completed: false });
         }, delay);
+    };
+
+    private readonly buttonStatusSRLabel = (status: string): string => {
+        const srLabels: Record<string, string> = {
+            loading: 'loading',
+            redirect: 'payButton.redirecting'
+        };
+
+        return srLabels[status] || '';
     };
 
     render() {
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/internal/Button/Button.tsx
```diff
@@ -109,24 +117,29 @@ class Button extends Component<ButtonProps, ButtonState> {
...
+                    onMouseLeave={onMouseLeave}
+                    onKeyDown={onKeyDown}
+                    onFocus={onFocus}
+                    onBlur={onBlur}
+                    onKeyPress={onKeyPress}
+                >
+                    {buttonText}
+                    {status !== 'loading' && status !== 'redirect' && this.props.children}
+                </button>
+                <div role="status" aria-live="polite" className="adyen-checkout__button__text--sr-only">
```
ScottiBR, line 139: The aria status should be a visually hidden container div always present when the page loads. The div must be empty unless a visual loading indicator is visible.

role="status" is the correct ARIA pattern for announcing dynamic, non-critical updates (like loading/redirect) to screen readers, with implicit polite live announcements across ATs.
"output" is for form result/calculation values and isn’t intended for live status updates; it doesn’t guarantee screen reader announcements in this scenario.

PR #3708, opened 2025-12-08
Add IRIS payment component

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

IRIS Payment is a real-time bank transfer system, popular in Greece, allowing direct payments from a customer's bank account via their mobile banking app, often by scanning a QR code or by selecting an bank issuer to be redirected to. This PR adds proper support for this Payment method to the SDK.

Screenshot 2025-12-17 at 15 05 42
Screenshot 2025-12-17 at 15 05 49
Screenshot 2025-12-17 at 15 06 48

---

## 🧪 Tested scenarios

Both the redirect flow and the qr code flow for the IRIS payment method have been covered by the following tests
- [x] Manual tests
- [x] Unit tests
- [x] E2E tests

---

COSDK-861

**Closes**:  

---


packages/lib/src/components/Iris/components/IrisComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+        props.setComponentRef(irisRef.current);
+    }, [props.setComponentRef, irisRef.current]);
+
+    useEffect(() => {
+        if (props.issuers.length === 0) {
+            handleModeChange(IrisMode.QR_CODE);
+        }
+    }, [props.issuers]);
+
+    if (props.issuers.length === 0) {
```
ScottiBR: if (!props.issuers.length)

packages/lib/src/components/Iris/components/IrisComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    useEffect(() => {
+        props.setComponentRef(irisRef.current);
+    }, [props.setComponentRef, irisRef.current]);
+
+    useEffect(() => {
+        if (props.issuers.length === 0) {
+            handleModeChange(IrisMode.QR_CODE);
+        }
```
ScottiBR: should we move this if inside the `handleModeChange` and use an early return approach like
```
if (props.issuers.length) return 
```

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
```diff
@@ -134,7 +134,7 @@ QRLoader.defaultProps = {
     countdownTime: DEFAULT_PAYMENT_STATUS_TIMER_COUNTDOWN_TIME_MIN,
     delay: DEFAULT_PAYMENT_STATUS_TIMER_DELAY_MS,
     throttleTime: DEFAULT_PAYMENT_STATUS_TIMER_THROTTLE_TIME_MS,
-    throttledInterval: DEFAULT_PAYMENT_STATUS_TIMER_THROTTLE_INTERVAL_MS,
+    throttleInterval: DEFAULT_PAYMENT_STATUS_TIMER_THROTTLE_INTERVAL_MS,
```
ScottiBR: since we are not going to expose the throttle prop in this PR, should we also not fix the typo?

packages/lib/src/styles/overrides.scss
```diff
@@ -1,7 +1,7 @@
 // Override the Bento font sizes. Calculation is based on a root font-size of 16 pixel.
 $text-overrides: (
-    // 12px
-    'text-caption-font-size': 0.75rem,
+    // 13px
+    'text-caption-font-size': 0.813rem,
```
ScottiBR: are this right?

packages/lib/src/components/Iris/components/IrisComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const { i18n } = useCoreContext();
+
+    const [mode, setMode] = useState<IrisMode>(props.defaultMode);
+    const [status, setStatus] = useState<UIElementStatus>('ready');
+    const segmentedControlOptions = useMemo(() => {
+        const options = getIrisSegmentedControlOptions(i18n);
+
+        if (isMobile()) {
+            return options.reverse();
+        }
```
ScottiBR: Since the logic for isMobile already exists in the main class, we should avoid duplicating the business rule here. Can we simply use the mode to define the initial state? 
`this.mode = isMobile() ? IrisMode.BANK_LIST : IrisMode.QR_CODE;`

packages/lib/src/components/Iris/components/IrisComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }, [props.issuers]);
+
+    if (props.issuers.length === 0) {
+        return <IrisGenerateQRCode showPayButton={props.showPayButton} payButton={props.payButton} status={status} />;
+    }
+
+    return (
+        <div>
+            <SegmentedControl onChange={handleModeChange} selectedValue={mode} disabled={status === 'loading'} options={segmentedControlOptions} />
```
m1aw: I see this is how we handle this logic in UPI, but I'm not the biggest fan of it. 
There's lots of prop drilling just to pass `onModechange`, for PayTo `selectedInput` is passed to the component data. I find it a more clear solution since we only ever need to use this data inside `formatData()`.

packages/lib/src/components/Iris/components/IrisComponent.module.scss
```diff
@@ -0,0 +1,6 @@
+@use 'styles/mixins';
+@import 'styles/variable-generator';
+
+.list {
```
m1aw: I think we could make this more explicit what it is. `.bankList` would be a suggestion. I know it's linked via the import, but intellisense was not working for this.

packages/lib/src/components/Iris/Iris.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+    beforeEach(() => {
+        jest.clearAllMocks();
+        isMobileMock.mockReturnValue(false); // Default to desktop
+    });
+
+    test('should pre-select "QR Code" mode on Desktop', async () => {
+        isMobileMock.mockReturnValue(false);
+        const core = setupCoreMock();
+
+        const iris = new Iris(core, {
```
m1aw: You can wrap this in a setup function, saves editing these lines for every tests.

packages/lib/src/styles/overrides.scss
johnayeni: I saw that the design now uses a 13px font size for this text. I have reverted it for now so I can confirm with Arjen later on it.

packages/lib/src/components/internal/QRLoader/QRLoader.tsx
johnayeni: Good point. Reverted for now

packages/lib/src/components/Iris/components/IrisComponent.tsx
johnayeni: I spilt this into two places as handleChange mode should only run on the life cycle update with the issuer list is empty. If this is also part of the early return in the JSX then that would run on every render.

packages/lib/src/components/Iris/components/IrisComponent.tsx
johnayeni: [done]

packages/lib/src/components/Iris/components/IrisComponent.tsx
johnayeni: I have refactored the `getIrisSegmentedControlOptions` function to take into account the `defaultOption` set on the UIElement level. This way we can easily change this if needed in the future quite easily without worrying about breaking business logic.

packages/lib/src/components/Iris/components/IrisComponent.tsx
johnayeni: `PayTo` seems to kinda be doing the same but just with the `onChange` prop. If we do the same here for IRIS I would still need to pass that function as a prop for it to work.

packages/lib/src/components/Iris/components/IrisComponent.module.scss
johnayeni: I have changed this to `bankList`. There needs to be a few more changes for scss modules but it seems especially regarding the intellisense but would do that seperately.

packages/lib/src/components/Iris/Iris.test.tsx
ScottiBR: Nice catch, done!

packages/lib/src/components/Iris/components/IrisComponent.tsx
ScottiBR: We have added an tech debt for refactoring state logic for payment methods creating providers for it

packages/e2e-playwright/tests/e2e/issuerList/iris.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+        await interceptAndValidatePaymentResponse({ page, expectedActionType: 'redirect' });
+        await iris.switchToBankListMode();
+        expect(await iris.isBankListModeSelected()).toBe(true);
+        
+        // The combobox should be visible in Bank List mode
+        await expect(iris.selectorCombobox).toBeVisible();
+        
+        // Select an issuer from the dropdown (opens the listbox)
+        await iris.selectIssuerOnSelectorDropdown('Piraeus Bank');
+        await iris.pay();
```
ribeiroguilherme, line 15: Are we sure that the `interceptAndValidatePaymentResponse()` is being executed and performing its assertions right after the  `await iris.pay()` is called?

Asking because it seems weird by reading the code . I would expect to see the assertions of the payments response right after the payment step, but they are added as first thing in the function block. Feels like they might be ignored.

packages/lib/src/components/helpers/QRLoaderContainer/types.ts
```diff
@@ -12,6 +12,16 @@ export interface QRLoaderConfiguration extends UIElementProps {
      */
     countdownTime?: number;
 
+    /**
+     * Number of milliseconds that the component will switch to throttled mode
+     */
+    throttleTime?: number;
```
ribeiroguilherme, line 21: Maybe we add in the JSDoc what is the default value of these fields

packages/lib/src/components/Iris/types.ts
```diff
@@ -0,0 +1,15 @@
+import { IssuerListConfiguration, QRLoaderConfiguration } from '../types';
+
+export type IrisConfiguration = IssuerListConfiguration & Pick<QRLoaderConfiguration, 'countdownTime' | 'qrCodeData' | 'paymentData' | 'delay'>;
```
ribeiroguilherme, line 3: Is it expected to not expose the `throttleTime` and `throttleInterval`?

packages/lib/src/components/Iris/types.ts
```diff
@@ -0,0 +1,15 @@
...
+    paymentMethod: {
+        type: string;
+        issuer?: string;
+    };
+}
+
+export enum IrisMode {
+    QR_CODE = 'qrCode',
+    BANK_LIST = 'bankList'
+}
```
ribeiroguilherme, line 23: This types will be exposed in the public typescript definitions, although they are not needed to be there. 

Perhaps add a description marking them as `@internal` and we add this [feature in the near future](https://www.typescriptlang.org/tsconfig/#stripInternal) , or move these types to a different file.

packages/lib/src/components/index.ts
```diff
@@ -47,6 +47,7 @@ export { default as BillDeskOnline } from './BillDesk/BillDeskOnline';
 export { default as BillDeskWallet } from './BillDesk/BillDeskWallet';
 export { default as Dotpay } from './Dotpay'; // add other tx variants
 export { default as EPS } from './EPS';
+export { default as Iris } from './Iris';
```
ribeiroguilherme, line 50: ⚠️  Component is being exported here, but it is not added to the `component-map.ts` and `component-name-map.ts`  files.

packages/lib/src/components/Iris/components/IrisComponent.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const irisRef = useRef<ComponentMethodsRef>({
+        setStatus: setStatus
+    });
+
+    useEffect(() => {
+        props.setComponentRef(irisRef.current);
+    }, [props.setComponentRef, irisRef.current]);
+
+    useEffect(() => {
+        if (!props.issuers.length) {
```
ribeiroguilherme: Is it even possible that the issuer list is empty when using this payment method ? 

If that is not the case, I would rather not add custom logic for this scenario. That also simplifies upstream on `Iris.tsx` where we are checking a boolean flag to know if analytic events need to be sent or not.

packages/lib/src/components/internal/SegmentedControl/SegmentedControlRegion.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export interface SegmentedControlRegionProps {
+    id: string;
+    ariaLabelledBy: string;
+    className?: string;
+    children: h.JSX.Element | h.JSX.Element[];
+}
+
+export const SegmentedControlRegion = ({ id, ariaLabelledBy, className, children }: SegmentedControlRegionProps) => {
+    return (
```
ribeiroguilherme, line 11: There is a sonarcloud issue about this specific `div` . Is it something that we need to look into?

packages/lib/src/components/Iris/types.ts
ScottiBR: Yes, that is intentional. We decided to handle the throttleTime and throttleInterval exports in a separate PR, similar to how we managed the recent onKeyPress fix.

This approach allows us to keep this PR focused while we refine the logic for other payment methods. Specifically, we want to ensure we implement guardrails, such as enforcing a minimum throttle interval and validating related props, before exposing them to the public API.

packages/lib/src/components/index.ts
ScottiBR: Nice catch, Done!

packages/e2e-playwright/tests/e2e/issuerList/iris.spec.ts
ScottiBR: The assertions do run  `page.route()` sets up an interceptor that fires when the request is made during iris.pay(). So the callback executes at the right time.

However, you raise a valid concern: if the request never happens, the test would pass silently with no assertions executed.

packages/lib/src/components/Iris/components/IrisComponent.tsx
ScottiBR: While it's rare for IRIs to have zero issuers, it is possible due to back-end filtering (e.g., amount limits or restrictions per issuer / merchant). By defaulting to the QR flow, we ensure a successful checkout path is always available.

packages/lib/src/components/internal/SegmentedControl/SegmentedControlRegion.tsx
johnayeni: I looked into this and this issue has actually existed since when we had UPI. I tried a different solution already but that still also had issues. Since this is now a reusable component, I think this can be fixed in a separate PR and would cover all the places where this component is used.

packages/lib/src/components/Iris/types.ts
johnayeni: Done

packages/lib/src/components/helpers/QRLoaderContainer/types.ts
johnayeni: Done

ribeiroguilherme (approved):

ScottiBR (dismissed):

PR #3741, opened 2026-01-02
Sessions - Allowing amount to be updated

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

This PR introduces the feature of patching the amount when using Sessions flow. The demo can be seen on Storybook.

Core changes:
-  Enhanced the `core.update` functionality to allow updating the amount without reseting the UI: `core.update(amount, { shouldRecreateDomElements: false });` 
- Introduced `AmountProvider` context, that receives the new amount and propagate it to all Preact elements that are consuming the `useAmount` hook
- When amount update happens, the new amount is propagated to the `core` , to all `UIElements` , and to the `AmountProvider` which propagates to the preact components
- The amount can be updated multiple times, but the patch request is done a single time in the `beforeSubmit` callback, before making the actual /payments call
- In the `beforeSubmit` callback, the merchant can get the most recent session data in order to perform the patch request

Disclaimer:
- Many file changes are just adding Typescript types or removing `amount` prop drilling and adding the `useAmount` hook instead

## TODO
- Add analytics to track the usage of the update amount flow
---

## 🧪 Tested scenarios
- Created Storybook stories showcasing Drop-in and Components flow
- Checked on CA that the amount is getting updated accordingly when doing payment
- Checked that amount gets updated for Wallets such Google Pay, CashApp pay

### DIFF
--- a/.changeset/warm-items-exist.md
+++ b/.changeset/warm-items-exist.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+New: Allowing amount updates without re-initializing Drop-in/Components


packages/lib/src/core/core.ts
```diff
@@ -278,24 +280,61 @@ class Core implements ICore {
...
+                return this;
             });
-            return this;
+        }
+
+        const { amount, secondaryAmount } = props;
+
+        if (amount || secondaryAmount) {
+            this.triggerAmountUpdate(amount, secondaryAmount);
+        }
```
ScottiBR, line 308: When `shouldRecreateDomElements: false`, the method returns undefined instead of a Promise<this> as is the current method contract

ribeiroguilherme: I updated the code

johnayeni:
@ribeiroguilherme everything looks good to me. The only thing I am still thinking about is the `shouldRecreateDomElements` prop. I am wondering if we should name it something like `shouldReinitializeCheckout` since that is what it technically does in general i.e reload the modules, calls setup and remounts the elements. `shouldRecreateDomElements` seems more like only remounting the elements and this might not be telling the SDK users what it does completely

packages/e2e-playwright/tests/automated/visual/visual.spec.ts
```diff
@@ -6,13 +6,23 @@ import { StorybookIndex } from '../types';
 // This is relative to playwright root: adyen-web/lib/e2e-playwright/
 const STORYBOOK_INDEX_PATH = '../lib/storybook-static/index.json';
 
+const EXCLUDED_STORIES: string[] = [
+    /**
+     * Demos
+     */
+    'demos-sessionpatching--with-dropin',
+    'demos-sessionpatching--with-components'
```
johnayeni: Instead of this, you can add the `no-automated-visual-test` tag to the stories itself and it would be excluded from the visual tests

See here https://github.com/Adyen/adyen-web/blob/main/packages/lib/storybook/stories/PaypalReviewPage.stories.tsx#L8

ribeiroguilherme: Oh yeah, nice. I updated the code

ScottiBR (approved):

PR #3752, opened 2026-01-07
Add Windsurf AI Rules for adyen-web

ScottiBR:
## 📝 Summary

Adds a .windsurfrules.md file to provide AI coding assistants (Windsurf/Cascade) with project-specific context and guidelines for the adyen-web SDK.

AI assistants work better with explicit project context. This file:

- Reduces hallucinations about React vs Preact
- Enforces PCI compliance rules in generated code
- Ensures consistent patterns across AI-assisted contributions
- Speeds up onboarding for AI-assisted development

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/.windsurf/rules/code-guide.md
+++ b/.windsurf/rules/code-guide.md
@@ -50,6 +50,17 @@ trigger: always_on
 | `src/utils/` | Pure functions | ZERO side effects. No DOM, no API calls, no state mutations. |
 | `src/language/` | Translation system | Add keys to `packages/server/translations/[locale].json`. |
 
+### Components Directory
+
+**Rules**: 
+- **DO**: treat every component folder as a self-contained unit containing `index.ts`, `[Name].tsx`, and `[Name].test.tsx`.
+- **DO**: Every component folder must have an `index.ts`.
+- **DO**: External files should **only** import from the `index.ts` of the folder, never the `.tsx` file directly.
+- **DO**: extract component-specific helper logic into `utils.ts` or `validate.ts` *inside* the component's own folder.
+- **DO NOT** import directly from a `.tsx` file; external files must only import from the folder's `index.ts`.
+- **DO NOT** use `export *` in `index.ts` to ensure clear APIs and better tree-shaking.
+- **DO NOT** allow sub-component logic (tests, utils, styles) to live outside its specific nested folder.
+
 ### Architectural Hierarchy
 
 All payment components extend `UIElement`:
@@ -140,7 +151,6 @@ When maintaining existing components using global SCSS:
 **Rules**:
 - **DO** group imports with blank lines between categories
 - **DO** use `import type` for type-only imports
-- **DO NOT** use relative imports traversing more than 2 levels up
 
 ---
 
@@ -177,8 +187,11 @@ submit() → makePaymentsCall() → handleResponse()
 
 ## Coding Standards
 
+### Export Strategy
+- **UIElements (Classes):** **DO** use `default export`. These are the main entry points for payment methods.
+- **Preact Components (Functional):** **DO** use `named exports`.
+
 ### Component Structure
-- **DO** use default exports for components
 - **DO** define props interfaces above the component
 - **DO** provide default values via destructuring
 - **DO** colocate: `Component.tsx`, `Component.test.tsx`, `ComponentName.module.scss`, `types.ts`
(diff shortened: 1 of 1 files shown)


.windsurf/rules/code-guide.md
```diff
@@ -0,0 +1,678 @@
...
+- **DO** validate props early, throw `AdyenCheckoutError` for misconfiguration
+- **DO NOT** make API calls in constructor (use `isAvailable()`)
+- **DO NOT** bypass `submit()` flow
+
+---
+
+## Coding Standards
+
+### Component Structure
+- **DO** use default exports for components
```
johnayeni: I think here we only need default exports for UIElements. For preact components, named exports are better because of the following
- tree shaking is easier to handle for bundlers
- named exports make it easier maintain components by preventing issues like using a different name for the component when imported

.windsurf/rules/code-guide.md
```diff
@@ -0,0 +1,678 @@
...
+### Accessibility (A11Y)
+- **DO** use semantic HTML (`<button>`, `<label>`, `<input>`)
+- **DO** add ARIA attributes: `aria-label`, `aria-describedby`, `role`
+- **DO** support keyboard navigation (Enter, Space, Escape, Arrow keys)
+- **DO** provide focus indicators (`@include b-focus-ring` mixin)
+
+### Localization
+- **DO** use `i18n.get('key')` for all user-facing strings
+- **DO** use descriptive keys: `card.number.label` not `cardNumber`
+- **DO NOT** hardcode English strings
```
johnayeni, line 239: One thing I think we can have here which is also an potential improvement point for our code base is passing translation keys as props. We should instead pass the result of `i18n.get` as this gives more options to parent components and makes it obvious where a string is used as a translation key

.windsurf/rules/code-guide.md
```diff
@@ -0,0 +1,678 @@
...
+2. Third-party libraries (classnames, etc.)
+3. Core/Utils (absolute imports)
+4. Components (internal imports)
+5. Types (`import type` statements)
+6. Styles (SCSS — always last)
+
+**Rules**:
+- **DO** group imports with blank lines between categories
+- **DO** use `import type` for type-only imports
+- **DO NOT** use relative imports traversing more than 2 levels up
```
ribeiroguilherme: I think it should be fine to traverse more than 2 levels up. Otherwise, what would it do?

.windsurf/rules/code-guide.md
```diff
@@ -0,0 +1,678 @@
...
+
+---
+
+## Coding Standards
+
+### Component Structure
+- **DO** use default exports for components
+- **DO** define props interfaces above the component
+- **DO** provide default values via destructuring
+- **DO** colocate: `Component.tsx`, `Component.test.tsx`, `ComponentName.module.scss`, `types.ts`
```
ribeiroguilherme, line 197: We need to mention that we also add a index.ts file exporting the Component from its folder

ribeiroguilherme (dismissed):

.windsurf/rules/code-guide.md
ScottiBR: make sense, this is probably some pattern picked from the last commits, maybe none of then had more than two levels up

.windsurf/rules/code-guide.md
ScottiBR: Good point, that also prevents AI from rename a component during import

.windsurf/rules/code-guide.md
ScottiBR: Good point, added

.windsurf/rules/code-guide.md
ScottiBR: Nice, I've added in another section

```

### Components Directory

**Rules**: 
- **DO**: treat every component folder as a self-contained unit containing `index.ts`, `[Name].tsx`, and `[Name].test.tsx`.
- **DO**: Every component folder must have an `index.ts`.
- **DO**: External files should **only** import from the `index.ts` of the folder, never the `.tsx` file directly.
- **DO**: extract component-specific helper logic into `utils.ts` or `validate.ts` *inside* the component's own folder.
- **DO NOT** import directly from a `.tsx` file; external files must only import from the folder's `index.ts`.
- **DO NOT** use `export *` in `index.ts` to ensure clear APIs and better tree-shaking.
- **DO NOT** allow sub-component logic (tests, utils, styles) to live outside its specific nested folder.
```

PR #3760, opened 2026-01-13
Remove UPI Collect (UPI ID) 

ScottiBR:
## 📋 Pull Request Checklist

---

## 📝 Summary

UPI Collect deprecated 

changes 
Remove UPI Collect segmented control from Desktop, making the default option qrcode with the redirect button.
image

Remove UPI Collect segmented control from Mobile, making the default option intent app list.
image
---

## 🧪 Tested scenarios

---

[COSDK-889](https://youtrack.is.adyen.com/issue/COSDK-889/Web-Removal-of-UPI-Collect-Flow)

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/UPI/UPI.tsx
+++ b/packages/lib/src/components/UPI/UPI.tsx
@@ -121,7 +121,7 @@ class UPI extends UIElement<UPIConfiguration> {
                         payButton={this.payButton}
                         onChange={this.setState}
                         apps={this.props.apps}
-                        defaultMode={this.mode}
+                        mode={this.mode}
                         showPayButton={this.props.showPayButton}
                         amount={this.props.amount}
                         mandate={this.props.mandate}
--- a/packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.test.tsx
+++ b/packages/lib/src/components/UPI/components/UPIComponent/UPIComponent.test.tsx
@@ -25,7 +25,7 @@ describe('UPIComponent', () => {
         const gpayApp: App = { id: 'gpay', name: 'Google Pay' };
 
         test('should show a list of apps from the given app list', async () => {
-            customRender(<UPIComponent apps={[gpayApp]} defaultMode={UPI_MODE.INTENT} onChange={jest.fn()} showPayButton={false} />);
+            customRender(<UPIComponent apps={[gpayApp]} mode={UPI_MODE.INTENT} onChange={jest.fn()} showPayButton={false} />);
 
             expect(await screen.findByRole('radiogroup')).toBeInTheDocument();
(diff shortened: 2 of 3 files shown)


packages/lib/src/components/UPI/types.ts
```diff
@@ -1,24 +1,22 @@
...
     paymentMethod: {
         type: UpiType;
-        virtualPaymentAddress?: string;
         appId?: string;
     };
 };
 
 export interface UPIConfiguration extends UIElementProps {
-    defaultMode?: UpiMode;
+    defaultMode?: UpiMode; /** @deprecated after upi_collect (vpa) deprecation  */
```
johnayeni: This comment needs to be above this line I think for the deprecation message to show from code intellisense. Maybe you can test that

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -3,65 +3,40 @@ import UIElement from '../internal/UIElement/UIElement';
...
             };
         }
 
-        // Mobile, but no UPI apps
         return {
             ...super.formatProps(props),
-            defaultMode: 'vpa', // Only VPA is possible
-            apps: [],
-            segmentedControlOptions: []
+            defaultMode: UPI_MODE.QR_CODE,
```
johnayeni: Since we no longer allow merchants to set this value, I think we can move it from props to a private property in the UPI class. We could also rename it to `mode` similar to what we have for IRIS.

packages/lib/src/components/UPI/types.ts
ScottiBR: Nice catch. That annotation only flags the prop as deprecated in the IDE (visually) and doesn't trigger a runtime warning in the console. I’ve added a console warning to address this.

packages/lib/src/components/UPI/UPI.tsx
ScottiBR: I agree

packages/lib/src/components/UPI/UPI.tsx
```diff
@@ -143,10 +120,8 @@ class UPI extends UIElement<UPIConfiguration> {
                         }}
                         payButton={this.payButton}
                         onChange={this.setState}
-                        onUpdateMode={this.onUpdateMode}
                         apps={this.props.apps}
-                        segmentedControlOptions={this.props.segmentedControlOptions}
-                        defaultMode={this.props.defaultMode}
+                        defaultMode={this.mode}
```
johnayeni:
```suggestion
                        mode={this.mode}
```

Maybe rename this as well since in the component, we do not have a mechanism to switch the mode so there is technically no initial default

ScottiBR: done

johnayeni (dismissed):

PR #3769, opened 2026-01-21
Improved: restrictions on some special characters on address validation

m1aw:
## 📋 Pull Request Checklist

---
## 📝 Summary

**Motivation**: The previous address validation logic silently stripped invalid characters (emojis, control characters) in the formatter, which provided poor UX as users didn't understand why their input was being modified.

**Changes**:
- Moved invalid character validation from formatters to validators for `street`, `houseNumberOrName`, and `city` fields
- Users now see an error message instead of silent character stripping
- Replaced the blocklist approach (`SPECIAL_CHARS`) with a Unicode property escape regex that blocks emojis (`\p{Extended_Pictographic}`) and control characters (`\p{Cc}`, `\p{Cf}`)
  - Additional we are making sure this Regex is valid, there's small change it might not be fully supported [Can I use](https://caniuse.com/?search=%5Cp)
- Formatters now only handle whitespace normalization (except postal codes which still strip invalid chars)

**Benefits**:
- Better UX: users understand why their input is invalid
- More permissive: allows special characters like `#` (used in Singapore addresses)
- Cleaner separation of concerns: formatters format, validators validate

**Notes for core reviews: (to discuss)**
- We are using a new set of Regex, support for most recent browsers and has fallback
- Unsure if the error message used in the should be this one

---
## 🧪 Tested scenarios

- Address fields with valid text (letters, numbers, international characters, punctuation) - accepted
- Address fields with emojis - validation error shown on blur
- Address fields with control characters - validation error shown on blur
- Optional `houseNumberOrName` field when empty - no validation error
- GB postal code with emojis - emojis stripped by formatter (existing behavior preserved)
- All existing Address and validator-utils tests pass

---

**Closes**:  COSDK-287


packages/lib/src/utils/validator-utils.ts
```diff
@@ -31,24 +40,24 @@ export const getFormattingRegEx = (specChars: string, flags = 'g') => new RegExp
...
-    const hasNoLength = !name.length;
-    // RegEx .test, if run against empty string, will return false
-    return CHARACTER_PATTERNS.noSpecialChars.test(name) || hasNoLength;
+export const hasInvalidChars = (text: string): boolean => {
+    if (!text.length) return true;
+    // Reset lastIndex for global regex before testing
+    INVALID_CHARS_REGEX.lastIndex = 0;
+    // Returns true if no invalid characters found
+    return !INVALID_CHARS_REGEX.test(text);
+};
```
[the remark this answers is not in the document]
m1aw: Indeed, I will make this change.

packages/lib/src/components/internal/Address/validate.ts
```diff
@@ -103,17 +103,48 @@ export const getAddressValidationRules = (specifications): ValidatorRules => {
...
+                errorMessage: ERROR_FIELD_REQUIRED
             },
-            modes: ['blur'],
-            errorMessage: ERROR_FIELD_REQUIRED
-        },
+            {
+                modes: ['blur'],
+                validate: value => hasInvalidChars(value),
+                errorMessage: ERROR_INVALID_CHARACTERS
+            }
```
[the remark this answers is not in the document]
m1aw: This also makes sense.

packages/server/translations/en-US.json
```diff
@@ -173,6 +173,7 @@
     "field.invalid": "Field not valid",
     "field.error.required": "Enter the %{label}",
     "field.error.invalid": "Enter a valid %{label}",
+    "field.error.invalidCharacters": "Remove invalid characters",
```
johnayeni: Should this message contain the group of characters allowed? That can better help the shopper to know what they are typing in wrong.

m1aw: Yeah, this was part of the discussion I had with Arjen. We should but how do convey this?

Right now we have: 
* `\\p{Extended_Pictographic}` - emojis
* `\\p{Cc}` - A control character
* `\\p{Cf}` - A format character

sponglord:
Like the overall approach - but just wondering... since the Unicode control strings are not visible should we not continue to silently strip them?
Otherwise we could be telling the user the input is not valid, because of characters they can't see, and expecting them to correct it, when they can't see what is wrong

packages/lib/src/utils/validator-utils.ts
```diff
@@ -31,24 +40,24 @@ export const getFormattingRegEx = (specChars: string, flags = 'g') => new RegExp
 // Creates a regEx ideal for use in a RegExp.test call for use in a validator
 export const getValidatingRegEx = (specChars: string, exclude: boolean) => new RegExp(`^[${exclude ? '^' : ''}${specChars}]+$`);
```
ribeiroguilherme: `getValidatingRegEx` is now dead-code. It can be removed

packages/lib/src/components/internal/Address/validate.ts
```diff
@@ -93,6 +93,18 @@ export const getPartialAddressValidationRules = (country: string): ValidatorRule
...
+const ruleIsEmpty: ValidatorRule = {
+    modes: ['blur'],
+    validate: value => (isEmpty(value) ? null : true),
+    errorMessage: ERROR_FIELD_REQUIRED
+};
+
+const ruleValidateForSpecialChars: ValidatorRule = {
+    modes: ['blur'],
+    validate: value => validateForSpecialChars(value),
+    errorMessage: ERROR_INVALID_CHARACTERS
```
ribeiroguilherme, line 105: This new `ERROR_INVALID_CHARACTERS` does not have an error code associated to it on the `errorMapping` variable.  Due to this, we send analytic events reporting the error but the error code is empty.

I am afraid we have to align with the API or Data team to define the error code for this error type.

packages/lib/src/utils/validator-utils.ts
```diff
@@ -22,7 +22,16 @@ export const isEmpty = (input: string) => !!(input == null || /^[\s]*$/.test(inp
 export const isString = (input: unknown) => typeof input === 'string' || input instanceof String;
 export const hasText = (input: string) => isString(input) && !isEmpty(input);
 
-export const SPECIAL_CHARS = '?\\+_=!@#$%^&*(){}~<>\\[\\]\\\\'; // N.B. difficulty escaping \ (takes 3 backslashes!)
+// Block emojis and control/format characters, allow everything else
+let INVALID_CHARS_REGEX: RegExp;
```
ribeiroguilherme: Nit: 
You could create an `IIFE` to create the regex. In this way you don't have to use `let`.

On another note, both `stripInvalidChars` and `validateForSpecialChars` methods have to keep mutating the `lastIndex` property on every call. (That seems to be due to how the `g` + `.test()` work on Regex). Honestly I wasn't aware that Regex would keep a state containing the index like that and that we would need to do this.

For better readability might be better to just have two regex and avoid the `lastIndex` reassigment.
```js
const [INVALID_CHARS_TEST, INVALID_CHARS_STRIP] = ((): [RegExp, RegExp] => { 
    ... 
})();
```

then:

```js
export const validateForSpecialChars = (text: string): boolean => !INVALID_CHARS_TEST.test(text ?? '');

export const stripInvalidChars = (text: string): string => (text ?? '').replace(INVALID_CHARS_STRIP, '');
```

ribeiroguilherme (dismissed):
Tested and it is working well. 

It would be nice to follow up on the missing  `errorCode` , so we make sure the analytic events regarding this are complete and not lacking data.

ribeiroguilherme (approved):

PR #3779, opened 2026-01-23
Typescript - Improving type safety in some files

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary
- Enhanced types for couple of utilities functions and few payment methods
- Improved types for ApplePay component in general
- Changed `getProp` function to be typed (although we should get rid of it)
- Updated `browserInfo.ts` to be strictly typed (e.g. not assigning string value to a number prop)
- For Card, we have a function `fieldTypeToSnakeCase` which returns the `target` for the Analytics event.  I didn't refactor it to fully return the type as `UiTarget` because that would require big effort. For now I just refactored the code to be asserted as `fieldTypeToSnakeCase(obj.fieldType) as UiTarget` 

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/.changeset/cold-apes-walk.md
+++ b/.changeset/cold-apes-walk.md
@@ -1,5 +1,0 @@
----
-'@adyen/adyen-web': minor
----
-
-Improved: remove usused css classses
--- a/.changeset/four-drinks-sniff.md
+++ b/.changeset/four-drinks-sniff.md
@@ -1,5 +1,0 @@
----
-'@adyen/adyen-web': patch
----
-
-Fixed: Detect when the containing element for a component that relies on securedFields, is not in the DOM. This means the iframe can never configure since, in order for the window.postMessage to be sent, the iframe has to be in the DOM
--- a/.changeset/full-games-shop.md
+++ b/.changeset/full-games-shop.md
@@ -1,5 +1,0 @@
----
-'@adyen/adyen-web': minor
----
-
-Fix: Removed UPI Collect support to comply with NPCI regulations, updating default flows to QR code for desktop and Intent for mobile.
--- a/.changeset/major-moles-flow.md
+++ b/.changeset/major-moles-flow.md
@@ -1,5 +1,0 @@
----
-'@adyen/adyen-web': patch
----
-
-Fix: Getting checkout attempt ID directly from Analytics module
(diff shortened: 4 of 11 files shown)


packages/lib/src/components/GooglePay/GooglePay.test.ts
```diff
@@ -193,13 +193,13 @@ describe('GooglePay', () => {
...
-            expect(browserInfo.screenWidth).toEqual('');
-            expect(browserInfo.timeZoneOffset).toEqual(expect.any(Number));
-            expect(browserInfo.userAgent).toEqual(expect.any(String));
+            expect(browserInfo.colorDepth).toBeDefined();
+            expect(browserInfo.javaEnabled).toBeDefined();
+            expect(browserInfo.language).toBeDefined();
+            expect(browserInfo.screenHeight).toBeDefined();
+            expect(browserInfo.screenWidth).toBeDefined();
+            expect(browserInfo.timeZoneOffset).toBeDefined();
+            expect(browserInfo.userAgent).toBeDefined();
```
johnayeni, line 202: Would it not be better to leave the previous test which validated the type of the value as well?

packages/lib/src/utils/promiseTimeout.ts
```diff
@@ -6,7 +6,7 @@
  * @param timeOutObject - the object that the promiseTimeout will reject with if the passed promise doesn't settle in time
  */
 const promiseTimeout = (ms: number, promise: Promise<any>, timeOutObject: object) => {
-    let timer;
+    let timer: NodeJS.Timeout;
```
johnayeni:
```suggestion
    let timer: NodeJS.Timeout | null;
```

packages/lib/src/components/GooglePay/GooglePay.test.ts
ribeiroguilherme: There is already test in place for the browserInfo utility on its own test file. 
I changed here to avoid mocking the 'window' in this file, but I can do add the mock here, that is not a problem

johnayeni: Ah okay. since it already has it's own test there is no need here 👍🏿

PR #3794, opened 2026-02-05
Fixing playwright config to enable HTTPS tests to run

ScottiBR:
## 📋 Pull Request Checklist

---

## 📝 Summary

This PR improves the stability of end-to-end tests when running against a `HTTPS` server using the port configuration instead of url.

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/packages/e2e-playwright/tests/automated/visual/visual.spec.ts
+++ b/packages/e2e-playwright/tests/automated/visual/visual.spec.ts
@@ -36,6 +36,9 @@ test.describe('Automated visual testing', () => {
             if (!storyId.includes('await')) {
                 await expect(page.getByTestId('spinner')).toBeHidden();
             }
+            if (storyId.includes('click-to-pay')) {
+                await expect(page.locator('.adyen-checkout-ctp__card-animation')).toBeHidden();
+            }
             await toHaveScreenshot(page.getByTestId('checkout-component'), browserName, `${storyId}.png`, {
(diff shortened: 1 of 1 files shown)


packages/e2e-playwright/playwright.config.ts
```diff
@@ -98,7 +98,7 @@ const config: PlaywrightTestConfig = {
         {
             command: 'npm run build:storybook:e2e && npm run start:prod-storybook',
             cwd: '../..',
-            url: storybookLocalHostUrl,
+            port: storybookPort,
```
sponglord, line 101: But what about if you are running on http? Do the tests still run if you use `port`?

ScottiBR: Yes, I’ve verified this locally and the tests run correctly using only the port configuration.

The only nuance is that while port ensures the server is accepting connections, it doesn't guarantee the DOM or specific assets have fully loaded before the tests kick off. This originally caused a flakiness issue with the Click To Pay animation, but I’ve resolved that by adding an explicit await for that specific element. Everything is now stable.

@johnayeni Can you also double check if your local tests are running with this approach?

image

johnayeni: Yeah, this works for me still. This was what we had before anyways before we switched to the `url` config so should be fine.

sponglord (approved):

PR #3797, opened 2026-02-06
Adding Bonus installment radio button 

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary
Adding support for `bonus` installment type on Installments component:
- If bonus is present, we display a radio button so it can be selected
- Adjusted function `shouldIncludeInstallmentsInPaymentData` to also check that `bonus` plan is a valid one 
---

## 🧪 Tested scenarios
- Added tests for Component rendering
- Checked API requests to see that installments are being added to the payload

---

COSDK-952

**Closes**:  

---

### DIFF
--- a/packages/e2e-playwright/tests/e2e/issuerList/iris.spec.ts
+++ b/packages/e2e-playwright/tests/e2e/issuerList/iris.spec.ts
@@ -26,25 +26,21 @@ test.describe('IRIS Payment Method', () => {
             await expect(iris.page).not.toHaveURL(URL_MAP.iris, { timeout: 5000 });
         }
     );
-    test.skip(
-        'QR Code Flow - should display QR code image and data after generating',
-        { tag: [TAGS.SCREENSHOT] },
-        async ({ page, iris, browserName }) => {
-            await iris.switchToQrCodeMode();
-            expect(await iris.isQrCodeModeSelected()).toBe(true);
-            await expect(iris.generateQrCodeButton).toBeVisible();
+    test('QR Code Flow - should display QR code image and data after generating', { tag: [TAGS.SCREENSHOT] }, async ({ page, iris, browserName }) => {
+        await iris.switchToQrCodeMode();
+        expect(await iris.isQrCodeModeSelected()).toBe(true);
+        await expect(iris.generateQrCodeButton).toBeVisible();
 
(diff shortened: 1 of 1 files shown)


packages/lib/src/components/Card/stories/Card.stories.tsx
```diff
@@ -171,6 +171,7 @@ export const WithMockedFastlane: CardStory = {
 
 export const WithClickToPay: CardStory = {
     render: createCardComponent,
+    tags: ['no-automated-visual-test'],
```
m1aw, line 174: Why was this change?

m1aw (changes requested):
Should we also make the change to fix the CSS for `installments` dropdown? 

I also think we should add a E2E test that takes a screenshot, I'm thinking in `card.installments.spec.ts`. The idea is to type the card until we get installments, then call `toHaveScreenshot` to make an assertion over this like we have in `iris.spec.ts`

packages/lib/src/components/Card/stories/Card.stories.tsx
ribeiroguilherme: CTP is async and unpredictable. Sometimes it loads fast, sometimes it takes quite some time, and sometimes it fails to load and Card UI is displayed (e.g. visa/mc scheme blocks the request due to big number of attempts). It is better to skip it from this workflow IMO

ribeiroguilherme:
> Should we also make the change to fix the CSS for installments dropdown?

That is change not related to this task. Better to be addressed in another PR.

> I also think we should add a E2E test that takes a screenshot, I'm thinking in card.installments.spec.ts. The idea is to type the card until we get installments, then call toHaveScreenshot to make an assertion over this like we have in iris.spec.ts

Yup, I can add that

packages/lib/src/components/Card/stories/Card.stories.tsx
johnayeni: These changes might address this issue with CTP https://github.com/Adyen/adyen-web/pull/3794/changes

ribeiroguilherme: Unfortunately it didn't really work

PR #3802, opened 2026-02-12
Fix/iOS scrolling issue

sponglord:
## 📋 Pull Request Checklist

---

## 📝 Summary

Fixes issue in iOS where programmatically setting focus was not causing the affected field to scroll into view

Also removed implicit `any`s from `/Card/components/CardInput/handlers.ts`

@ScottiBR - following our conversation: looking forward to your thoughts on user-agent detection; and on whether the iOS device is generating any errors when we try to set programmatic focus

---

## 🧪 Tested scenarios

Extensive manually testing in Browserstack, via Netlify deployment, on both iPhone 13 (running iOS 15) & 14 (running iOS 16), in both Safari and Chrome.
Also tested in Safari on an iPhone 14 running iOS 26

---

COSDK-577

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/handlers.ts
+++ b/packages/lib/src/components/Card/components/CardInput/handlers.ts
@@ -21,7 +21,7 @@ export const setFocusOnFirstField = (isValidating: boolean, sfp: SecuredFieldsPr
 
         // If not a cardInput related securedField - find field and set focus on it
         if (!CREDIT_CARD_SF_FIELDS.includes(fieldToFocus)) {
-            setFocusOnNonSF(fieldToFocus, sfp);
+            setFocusOnNonSF(fieldToFocus, sfp, ua.__IS_IOS);
         } else {
             // Is a securedField - so it has its own focus procedures
             sfp.current.setFocusOn(fieldToFocus);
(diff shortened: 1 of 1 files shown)


packages/lib/src/utils/setFocus.ts
```diff
@@ -19,14 +23,16 @@ export const setFocusOnField = (holder: Element | string, fieldToFocus: string,
...
-        field?.focus();
+        field = selectOne(pdHolder, `${focusContextSelector} [name="${fieldToFocus}"]`);
+    }
+
+    // Fix for iOS scrolling issues: can't programmatically set focus on an element on iOS, so we scroll to it instead, so at least it is in view
+    if (ua.__IS_IOS) {
+        windowScrollTo(field);
     }
+
+    field?.focus();
```
ScottiBR: when you call `.focus()` and a manual scroll at the same time, if the focus works it may "double jump" for some IOS edge case.
You may use the [preventScroll](https://caniuse.com/mdn-api_htmlelement_focus_options_preventscroll_parameter) option to prevent focus to scroll for IOS

sponglord: ok. I'll make the change
Out of interest did you see this happening, or are you just being super cautious? :-)

Issue #3799, opened 2026-02-11
paymentMethod.fundingSource always returns 'credit'


fwebdev:
**Summary**
`paymentMethod.fundingSource` from Adyen Web Components (v6.28.0) always returns `"credit"` in `onChange`/state data, regardless of using credit, debit, or prepaid test cards. We need to distinguish credit vs debit vs prepaid for compliance, but cannot rely on `fundingSource`.  

We added splitCardFundingSources: true to the Session POST in the Backend.

Choose different CreditCards from here:
https://docs.adyen.com/development-resources/test-cards-and-credentials/test-card-numbers

**Environment**
- Adyen Web Components: 6.28.0  
- Integration: Sessions flow (web components)  
- Component: Card (`new Card(adyenCheckout, cardConf)`), not `checkout.create`  

onChange of the CardConf
```
onChange: function (state) {
    const fundingSource = state.data.paymentMethod.fundingSource;  
    console.log('Funding Source:', fundingSource); // always returns 'credit'
}
```

m1aw:
Hey @fwebdev this indeed the case at the moment on Components, this configuration is meant to be used together with the Drop-in instead of standalone components. 

For standalone components you can achieve this by passing `fundingSource` directly as a card component configuration. 

I have reached out internally if this is a flow we want to support.

fwebdev:
Thanks, would be good to at least clearify that in the Documentation. There are also several times Code Snippets for older Version pre V6 which makes it more complicated also.

ribeiroguilherme:
Hi @fwebdev ,

We released a fix for this on v6.32.0. Please keep us posted if the issue persists.

Cheers

PR #3828, opened 2026-02-27
Add support for split funding source card components on sessions

m1aw:
## 📋 Pull Request Checklist

---

### Why
Support `splitCardFundingSources` sessions responses where multiple `scheme` payment methods are returned, differentiated by `fundingSource` (credit/debit/prepaid). Without this, standalone Card components can’t reliably load the correct payment method config when more than one `scheme` entry exists.

### How
- Extend payment method lookup to optionally match by `fundingSource` in addition to `type`.
- Plumb `fundingSource` through `UIElement` config resolution so components can select the right payment method config.
- Add unit coverage for `UIElement` config resolution with `fundingSource` (match + fallback).
- Introduce a typed `FundingSourceKeys` union and tighten Storybook `sessionData` typing (sessions-request-focused).

---

## 🧪 Tested scenarios

- Add Storybook coverage to render multiple Card components side-by-side with different `fundingSource` values.

---

**Closes**:  
#3799 

---

### DIFF
--- a/packages/lib/src/components/Card/stories/Card.stories.tsx
+++ b/packages/lib/src/components/Card/stories/Card.stories.tsx
@@ -266,6 +266,7 @@ export const StandaloneStoredCard: CardStory = {
  * - Prepaid: Should NOT render CtP or installments
  */
 export const SplitFundingSourceTest: CardStory = {
+    tags: ['no-automated-visual-test'],
     render: SplitFundingSourceCards,
     args: {
         countryCode: 'BR',


packages/lib/src/core/ProcessResponse/PaymentMethods/PaymentMethods.ts
```diff
@@ -43,7 +43,16 @@ class PaymentMethods {
...
+        // This is because when we set splitCardFundingSources: true, we get multiple card payment methods
+        // with the same type but different fundingSource values
+        if (fundingSource) {
+            const found = this.paymentMethods.find(
+                pm => pm.type === this.mapCreatedComponentType(paymentMethod) && pm.fundingSource === fundingSource
+            );
+            if (found) return found;
+        }
         return this.paymentMethods.find(pm => pm.type === this.mapCreatedComponentType(paymentMethod));
     }
```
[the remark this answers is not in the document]
ribeiroguilherme: I agree with the changes proposed by Gemini. The code comments can also be place on top of the function declaration and there we document the utility of the fundingSource parameter.

packages/lib/src/core/ProcessResponse/PaymentMethods/PaymentMethods.ts
```diff
@@ -43,7 +43,16 @@ class PaymentMethods {
         return Boolean(this.paymentMethods.find(pm => pm.type === this.mapCreatedComponentType(paymentMethod)));
     }
 
-    find(paymentMethod: string): PaymentMethod {
+    find(paymentMethod: string, fundingSource?: string): PaymentMethod {
+        // HACK: For card components, we need to match the fundingSource as well
+        // This is because when we set splitCardFundingSources: true, we get multiple card payment methods
+        // with the same type but different fundingSource values
+        if (fundingSource) {
```
ribeiroguilherme: Another possible approach would be to put this logic in a separated method (e.g. `findByFundingSource`) .  That way each function does its own thing and it is easier to test. It can be easily unit-tested on the `PaymentMethod.test.ts` file

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -132,7 +132,7 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
 
     private getPaymentMethodConfigFromResponse(componentProps: P) {
         if (componentProps?.storedPaymentMethodId) return this.getStoredPaymentMethodDetails(componentProps.storedPaymentMethodId);
-        return this.getPaymentMethodFromPaymentMethodsResponse(componentProps?.type, componentProps?.paymentMethodId);
+        return this.getPaymentMethodFromPaymentMethodsResponse(componentProps?.type, componentProps?.paymentMethodId, componentProps?.fundingSource);
```
ribeiroguilherme: Is `fundingSource` only relevant to Cards? If so, perhaps makes more sense to override the logic on the Card component itself rather than adding it to all payment methods

packages/lib/src/components/Card/stories/Card.stories.tsx
```diff
@@ -253,4 +254,34 @@ export const StandaloneStoredCard: CardStory = {
...
+export const SplitFundingSourceTest: CardStory = {
+    render: SplitFundingSourceCards,
+    args: {
+        countryCode: 'BR',
+        componentConfiguration: {
+            _disableClickToPay: true,
+
+            clickToPayConfiguration: {
+                shopperEmail: 'levelaccess.ctp@adyen.com',
+                merchantDisplayName: 'Adyen Merchant Name'
```
ribeiroguilherme, line 277: Click to pay config should not be needed here

packages/lib/src/components/Card/stories/Card.stories.tsx
```diff
@@ -253,4 +254,34 @@ export const StandaloneStoredCard: CardStory = {
...
+        },
+        sessionData: {
+            splitCardFundingSources: true,
+            shopperEmail: 'levelaccess.ctp@adyen.com',
+            installmentOptions: {
+                card: {
+                    values: [2, 3, 5],
+                    plans: ['regular']
+                }
+            }
```
ribeiroguilherme, line 288: I believe `shopperEmail + installmentOptions` can be also removed

packages/lib/src/components/Card/stories/cardStoryHelpers/SplitFundingSourceCards.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const SplitFundingSourceCards = (args: PaymentMethodStoryProps<CardConfiguration>) => {
+    const { componentConfiguration, ...checkoutConfig } = args;
+
+    return (
+        <Checkout checkoutConfig={checkoutConfig}>
+            {checkout => {
+                const card1 = new Card(checkout, { ...componentConfiguration, fundingSource: 'credit' });
+                const card2 = new Card(checkout, { ...componentConfiguration, fundingSource: 'debit' });
```
ribeiroguilherme, line 28: Minor thing - doing payment  with the Debit component renders the Authorization UI not in the desired place . I didn't test with the `prepaid` one. `credit` works fine

packages/lib/src/components/Card/stories/Card.stories.tsx
m1aw: It's needed as test scenario, for side effects. 
For example: 
- Credit: Should render CtP and installments
- Debit: Should render CtP, but NOT installments
- Prepaid: Should NOT render CtP or installments

packages/lib/src/components/Card/stories/Card.stories.tsx
m1aw: Same as above.

packages/lib/src/components/internal/UIElement/UIElement.tsx
m1aw: I will explore this.

.changeset/red-guests-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+Add support for split funding source card components on sessions
```
ribeiroguilherme: I expect the pipeline to fail here to due the changeset text not containing the prefix

packages/lib/src/core/ProcessResponse/PaymentMethods/PaymentMethods.test.ts
```diff
@@ -47,4 +47,44 @@ describe('PaymentMethodsResponse', () => {
...
+            paymentMethods: [
+                { name: 'Credit Card', type: 'scheme', fundingSource: 'credit' },
+                { name: 'Debit Card', type: 'scheme', fundingSource: 'debit' },
+                { name: 'PayPal', type: 'paypal' }
+            ]
+        };
+
+        test('should return the payment method matching both type and fundingSource', () => {
+            const pmResponse = new PaymentMethods(splitFundingResponse);
+            // eslint-disable-next-line testing-library/await-async-queries
```
ribeiroguilherme, line 62: Do we know why this is flagged as asynchronous method? It does not make sense to me. There is no async code there

packages/lib/src/core/ProcessResponse/PaymentMethods/PaymentMethods.ts
```diff
@@ -47,6 +47,17 @@ class PaymentMethods {
...
 
+    /**
+     * Finds a payment method matching both type and funding source.
+     * Used when `splitCardFundingSources` is enabled, which results in multiple payment methods
+     * of the same type but with different fundingSource values (e.g. credit, debit, prepaid).
+     * Falls back to the first match by type if no funding source match is found.
+     */
+    findByFundingSource(paymentMethod: string, fundingSource: string): PaymentMethod {
+        const mappedType = this.mapCreatedComponentType(paymentMethod);
+        return this.paymentMethods.find(pm => pm.type === mappedType && pm.fundingSource === fundingSource) || this.find(paymentMethod);
```
ribeiroguilherme: Do we need the fallback `this.find(paymentMethod)`  or should this method exclusively look for funding source? 

The `getPaymentMethodFromPaymentMethodsResponse`  method being overridden in Card already guarantee that the `find` will be used in case the funding source search does not work.

packages/lib/src/core/ProcessResponse/PaymentMethods/PaymentMethods.test.ts
m1aw: It took me a while to understand too, that's because it's a testing-library test and it starts with `findBy`.

PR #3840, opened 2026-03-04
Adding new CI/CD flow to prevent typescript any in new code

ScottiBR:
## 📝 Summary
Create a new workflow to inform it the PR is increasing the number of new `implicit any` types:

## Sucess:
image

## Fail:
image

## update comment for every new commit
image

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/.github/workflows/strict-typescript.yml
+++ b/.github/workflows/strict-typescript.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Checkout master
         uses: actions/checkout@v6
         with:
-          ref: master
+          ref: main
           fetch-depth: 1
 
       - name: Install dependencies (master)


.github/workflows/strict-typescript.yml
```diff
@@ -0,0 +1,103 @@
...
+jobs:
+  strict-scan:
+    runs-on: ubuntu-latest
+    # Non-blocking: this job is informational only and will never prevent merging.
+    continue-on-error: true
+
+    env:
+      # Current noImplicitAny error count from ts-errors.log.
+      # Decrease this number each time strict errors are fixed.
+      TSC_STRICT_BASELINE: 2103
```
johnayeni: Instead of hardcoding the baseline here, what you do think of a workflow where we can checkout to the main branch, calculate the number of implicit anys, then compare it to the number in the current branch. We have a similar concent for the [size limit workflow](https://github.com/Adyen/adyen-web/blob/main/.github/workflows/.size-limit.yml) although there is a github action that does this

ScottiBR: Nice suggestion

ScottiBR: I'll need to update the package lock on master adding the script `validate:strict` first in a separated PR

ScottiBR: https://github.com/Adyen/adyen-web/pull/3850

PR #3853, opened 2026-03-06
Fixed: default GooglePay environment to PRODUCTION instead of TEST

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

GooglePay `resolveEnvironment()` defaulted to `TEST` for unrecognized environment strings, while the rest of the SDK defaults to live. This caused merchants using regional environments (e.g. `live-au`) to silently fall back to Google Pay TEST mode in production.

- Changed default from `TEST` to `PRODUCTION` — only explicit `test`/`beta` resolves to TEST
- Made environment matching case-insensitive
- Added integration tests using real `AdyenCheckout` instances validating the full chain: `AdyenCheckout(environment)` → `GooglePay` → [GooglePayService](cci:2://file:///Users/antoniof/workspace/web-dev/adyen-web/packages/lib/src/components/GooglePay/GooglePayService.ts:7:0-68:1) → `resolveEnvironment()`

---

## 🧪 Tested scenarios

- `live` → PRODUCTION
- `test` → TEST
- `live-au` (regional) → PRODUCTION
- `live-uk` (unrecognized) → PRODUCTION
- `TEST` (case-insensitive) → TEST
- Custom `_environmentUrls` with live/test environments → correct resolution

---

**Closes**:
COSDK-818

---

### DIFF
--- a/.changeset/tame-baboons-try.md
+++ b/.changeset/tame-baboons-try.md
@@ -2,4 +2,4 @@
 '@adyen/adyen-web': patch
 ---
 
-Fix: default GooglePay environment to PRODUCTION instead of TEST for unrecognized environment strings
+Fixed: default GooglePay environment to PRODUCTION instead of TEST for unrecognized environment strings


packages/lib/src/components/GooglePay/GooglePayEnvironment.test.ts
```diff
@@ -0,0 +1,106 @@
...
+            }
+        });
+        await checkout.initialize();
+
+        new GooglePay(checkout, {
+            configuration: { merchantId: 'merchant-id', gatewayMerchantId: 'gateway-id' }
+        });
+
+        expect(getGooglePaymentsClientSpy).toHaveBeenCalledWith(expect.objectContaining({ environment: 'TEST' }));
+    });
```
ribeiroguilherme, line 87: Are these test cases relevant at all? We do not change environments based on the `_environmentsUrls` property as far as I know

packages/lib/src/components/GooglePay/utils.ts
```diff
@@ -1,20 +1,16 @@
 import { AddressData } from '../../types/global-types';
 
 /**
- *
+ * Use same logic as Environment.ts where fallback is live (production)
  */
-export function resolveEnvironment(env = 'TEST'): google.payments.api.Environment {
+export function resolveEnvironment(env = 'PRODUCTION'): google.payments.api.Environment {
     const environment = env.toLowerCase();
```
ribeiroguilherme: To avoid having to convert `environment` to lower case in multiple places, it can be nice if it is already converted on a higher level . That could be done [here](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/core/core.ts#L358) I believe

johnayeni:
> It is missing the changeset file

Might be nice to add a workflow that can check for this

packages/lib/src/components/GooglePay/GooglePayEnvironment.test.ts
m1aw: I added them because I remembered we had an issue at some point with PBL and it defaulting to the wrong GooglePay env when we were using custom environments, before we created `_environmentsUrls`.

I understand is just there for documentation, but in my view prevents some side effects in the future. That's why I went the extra length too to make these tests go through core.

packages/lib/src/components/GooglePay/utils.ts
m1aw: I have made those changes, and on env. Can you check?

Issue #3847, opened 2026-03-05
Installments component is broken when missing amount in the global configuration for Advanced Flow


abnersajr:
**Describe the bug**

Installments are throwing an error causing the whole payment selection not work.
After the change from `prop-drilling` to `context-provider`, the origin of the amount probably changed and users without the object at core configuration are broken.

Type definitions and Docs still marking this as optional.

**To Reproduce**
1. Follow the SDK doc: https://docs.adyen.com/online-payments/build-your-integration/advanced-flow?platform=Web&integration=Drop-in&version=6.31.0 **DO NOT ADD THE AMOUNT OBJECT TO THE CONFIG**
2. Try to make a card payment with installments.

**Expected behavior**
Work as used to work getting amount from somewhere, or making a highlight that this is now required and adjusting Type Definitions and Docs.

**Screenshots**

### Doc incorrect:
<img width="1880" height="2216" alt="Image" src="https://github.com/user-attachments/assets/d43cbd8a-6d31-4077-8a33-a02a78c4a027" />

### Console error:
<img width="1736" height="460" alt="Image" src="https://github.com/user-attachments/assets/4a5535ce-4ec1-4cf1-8c8e-ea775576f825" />

**Desktop (please complete the following information):**
 - OS:  Tested on MacOS and Linux
 - Browser Chrome
 - Version Latest

**Additional context**
I assume this error will happens in any browser and OS.

ribeiroguilherme:
Hi @abnersajr ,

Thanks for reaching out and for flagging this issue. Installments are not supposed to be displayed when there is no amount set. 
I will push a fix for this issue and hopefully we can release it soon.

sponglord:
@abnersajr - this is fixed in the latest release [6.32.0](https://github.com/Adyen/adyen-web/releases/tag/v6.32.0)

PR #3858, opened 2026-03-10
Fixing installments when amount is undefined

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary
- Added utility function that contains the proper logic to decide if installments must be displayed or not
- Added unit tests for it
- Also added safe-guard check on Installments component to verify that if the amount is undefined, no UI should be displayed

---

## 🧪 Tested scenarios
- Tested advanced flow, setting amount to null, verifying that there was no exceptions thrown in the console

**Closes**: #3847 

---

### DIFF
--- a/packages/lib/src/components/Card/components/CardInput/CardInput.dualBrandUI.test.tsx
+++ b/packages/lib/src/components/Card/components/CardInput/CardInput.dualBrandUI.test.tsx
@@ -1,7 +1,8 @@
-import { h } from 'preact';
+import { createRef, h } from 'preact';
 import { mount } from 'enzyme';
 import CardInput from './CardInput';
 import { CoreProvider } from '../../../../core/Context/CoreProvider';
+import { AmountProvider } from '../../../../core/Context/AmountProvider';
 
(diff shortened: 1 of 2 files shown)


packages/lib/src/components/Card/components/CardInput/utils.ts
```diff
@@ -220,3 +221,32 @@ export const mustHandleDualBrandingAccordingToEURegulations = (
...
+    installmentOptions?: InstallmentOptions;
+    fundingSource?: string;
+    amount?: PaymentAmount;
+}): boolean => {
+    if (!installmentOptions || Object.keys(installmentOptions).length === 0) return false;
+    if (!amount || amount.value === 0) return false;
+    if (fundingSource && fundingSource !== 'credit') return false;
+
+    return true;
+};
```
ribeiroguilherme, line 252: @m1aw since you recently worked on installments, can you double check that this logic is ok?

m1aw: This makes sense.

PR #3878, opened 2026-03-19
Feature/giving on sessions final

sponglord:
## 📋 Pull Request Checklist

---

## 📝 Summary

A sessions `/payments` response may now mandate showing the Giving (Donation) component by returning `askDonation:true`.

This means that internally, we then need to then go on to check a new endpoint, `/sessions/donationsCampaigns` to see if there are any active campaigns for this merchant and, if so, allow the shopper to donate to this campaign.

The donation is then made against another new endpoint, `/sessions/donations`.

At the same time, the regular payment flow should continue, leading to the expected call to `onPaymentCompleted`.

--------
The approach taken is as follows.
- A merchant defines a `donation` object in their checkout config 
- This has the following properties:
```
donation: {
  autoMount: boolean, // whether the sdk will automatically initiate a Donation component
  delay: number, // the delay in ms, after a successful payment, after which the Donation component wiil be displayed
  onDonationSuccess: (didDonate: boolean) => void, // merchant defined callback for a successful, or a cancelled, donation
  onDonationFailure: (reason: unknown) => void // merchant defined callback for an error in the Donation creation or an error when a donation is made
}
```

_NOTE: The merchant doesn't have to define this `donation` config. It is optional. But if they do define it they are expected to at least add the `onDonationSuccess` and `onDonationFailure` callbacks. This will allow them , at runtime, to know if the shopper donation has been successfully made, or not. Without these callbacks the merchant will have to rely on other avenues e.g. web hooks, or manually checking in the BO/CA, to see what the donation outcome was._

Then in `UIElement` when `handleSuccessResult` is called, if `donation.autoMount = true` (which is the default), a `Donation` component is initialised and passed a **props object\*:**
```
{
  rootNode: HTMLElement, // the default `rootNode` into which to mount 
  commercialTxAmount: number // the *value* of the payment amount (needed for "roundup" donations)
}
``` 
The `Donation` component then uses a `DonationCampaignService` to check the `/sessions/donationsCampaigns` for active campaigns and, if it finds one, passes it back to the `Donation` component to be mounted as a Donation _campaign_. 
By default this will be mounted into the same container as the original Dropin, or component, instance was mounted into.

When the shopper chooses to make donation, or chooses not to - the merchant defined callback, `donation.onDonationSuccess`, is called and passed an object as an argument: `{didDonate:true|false}`
If something goes wrong with the Donation  component creation, or with the `/donations` call, then a new callback, `donation.onDonationFailure`, is called

**For the merchant who wishes to define an alternate place into which to mount the `Donation` component** they can interrupt the process in the `onPaymentCompleted` callback.
Here, the merchant can clear any pre-exisiting UI, for example. in the case of a components implementation - the "payment successful" message.
Then, if they have set `donation.autoMount: false`, they can initialise a new `Donation` component, passing it a **props object\*** in the same form as described above, with the `rootNode` pointing to a container of their choice.

e.g.
```
function onPaymentCompleted(result, element) {
    // If needed clear the UI

    // Check that the /payments result mandates a donation
    if (result.askDonation === true) {
        new Donation(element.core, {
            rootNode: '.my-custom-container`, // add a selector (or reference to an actual DOM node) that points to your custom container
            commercialTxAmount: element.props.amount.value // needed in case the retrieved donation campaign is of type 'roundup'
        });
    }
}
```

The same process will then take place: the `Donation` component uses the `DonationCampaignService` to check for active campaigns and, if one is found, the Donation _campaign_ is mounted after the set delay.

_NOTE: If the merchant doesn't set `donation.autoMount: false` the attempt to  initialise a new `Donation` component will be ignored, and the "automatic" process, of mounting the component into the original container, will proceed._

---
New analytics have been added for:
- when the frontend calls the `/donationsCampaigns` endpoint (`Log` event)
- from within the `Donation` component for when:
    -  "donate" or "cancel" are pressed (both `Log` events) 
    - or, any of the amount buttons are pressed (`Info` event)

---

## 🧪 Tested scenarios
Added new unit & e2e tests
All existing tests pass

---

COSDK-250


packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -3,14 +3,53 @@ import UIElement from '../internal/UIElement/UIElement';
...
+
+        if (checkout.session && isServiceMode) {
+            new DonationCampaignService(checkout, props)
+                .initialise()
+                .then((response: DonationConfiguration) => {
+                    this.props = { ...this.props, ...response };
+                    this.mount(props.rootNode);
+                })
+                .catch((error: unknown) => {
+                    console.debug('Donation::init using DonationCampaignService:: error', error);
```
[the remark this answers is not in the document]
sponglord: ok

packages/lib/src/components/Donation/_DonationCampaignProvider.tsx
```diff
@@ -0,0 +1,198 @@
...
+            return await this.core.session.donations(donationRequestData);
+        } catch (error: unknown) {
+            // TODO - analytics, or will this be picked up as part of the sessions endpoint httpPost handling?
+
+            return Promise.reject(error);
+        }
+    }
+}
+
+export default DonationCampaignProvider;
```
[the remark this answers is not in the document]
sponglord: ok

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,170 @@
...
+        DonationCampaignService.instanceCount++;
+
+        // If the merchant has not explicitly set autoStart to false, and then tries to display the Donation component in a different container,
+        // we need to warn them otherwise, without this check, 2 calls to the /donationCampaigns endpoint will be made - since UIElement
+        // will automatically mount the Donation component in the component's rootNode.
+        if (DonationCampaignService.instanceCount > 1) {
+            console.error(
+                'DonationCampaignService:: You need to set donation.autoStart to false if you wish to display the Donation component in a different container'
+            );
+            return null;
```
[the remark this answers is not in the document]
sponglord: ok

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+    private readonly commercialTxAmount: number = 0;
+    private readonly onDonationCompleted: (didDonate: boolean) => void;
+    private readonly onDonationFailed: (reason: unknown) => void;
+
+    private readonly autoStartTimerMS = 3000;
+    private readonly delayMS: number;
+
+    constructor(checkout: ICore, dcpProps?: DonationCampaignOptions) {
+        DonationCampaignService.instanceCount++;
```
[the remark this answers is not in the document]
sponglord: Incorrect assumption. The payment/session lifecycle means that it will only ever be possible to make one donation per instance of checkout/session

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+    private readonly core: ICore;
+
+    private readonly commercialTxAmount: number = 0;
+    private readonly onDonationCompleted: (result: { didDonate: boolean }) => void;
+    private readonly onDonationFailed: (reason: unknown) => void;
+
+    private readonly autoStartTimerMS = 3000;
+    private readonly delayMS: number;
+
+    constructor(checkout: ICore, dcpProps?: DonationCampaignOptions) {
```
ScottiBR: Mind if we rename `dcpProps` to `donationCampaignProps` or something like? Let's try to avoid abbreviations to keep the codebase searchable (for AI and us)

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+            );
+        }
+
+        this.core = checkout;
+
+        this.onDonationCompleted = checkout.options.donation?.onSuccess;
+
+        this.onDonationFailed = checkout.options.donation?.onError;
+
+        this.commercialTxAmount = dcpProps.commercialTxAmount;
```
ScottiBR: I see the `props` is typed with a `?`. Is this intended to be optional in some cases? If so, we'll need a  optional chaining here `.?commercialTxAmount` or fix the type as a required.

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+            throw new Error(
+                'DonationCampaignService:: You need to set donation.autoStart to false if you wish to display the Donation component in a different container.'
+            );
+        }
+
+        this.core = checkout;
+
+        this.onDonationCompleted = checkout.options.donation?.onSuccess;
+
+        this.onDonationFailed = checkout.options.donation?.onError;
```
ScottiBR: The `donation` object should be required in session flow? 

Should we throw a error during initialization if the merchant hasn't configured properly? or fix the typo on `onDonationCompleted` and `onDonationFailed` to make it optional?

```
if (!checkout.options.donation) {
  throw new Error("Missing donation configuration in checkout options.");
}
```

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+    CheckoutSessionDonationsRequestData,
+    CheckoutSessionDonationsResponse
+} from '../../core/CheckoutSession/types';
+import Donation from './Donation';
+import { AnalyticsLogEvent, LogEventSubtype, LogEventType } from '../../core/Analytics/events/AnalyticsLogEvent';
+
+class DonationCampaignService {
+    public static type = 'donationCampaignService';
+
+    private static instanceCount: number = 0;
```
ScottiBR: This `instanceCount` static property persists in memory even after an donation component is unmounted, this counter will keep increasing every time a merchant triggers a new donation flow (like an upsell after the first donation).
Should this counter be decremented on unmount? Or this is a business rule like a donation is only made once for the same shopper?

packages/lib/src/components/Donation/components/DonationComponent.tsx
```diff
@@ -33,6 +33,8 @@ export default function DonationComponent(props: Readonly<DonationComponentProps
         const value = parseInt(target.value, 10);
         setIsValid(true);
         setAmount((amount: DonationAmount) => ({ ...amount, value }));
+
+        this.props.onAmountSelected({ data: { ...amount, value } });
```
ScottiBR: Should this `onAmountSelected` be together with the other props at the beginning of the component, just to keep the pattern?

```
 const { donation, commercialTxAmount, onAmountSelected, onCancel, onDonate, showCancelButton = true, termsAndConditionsUrl } = props;
```

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+                const donationRequestData: CheckoutSessionDonationsRequestData = {
+                    amount: state.data.amount,
+                    donationCampaignId: id,
+                    donationType: donationType
+                };
+
+                this.callSessionsDonations(donationRequestData, component);
+            },
+            commercialTxAmount: this.commercialTxAmount,
+            ...(restDonationCampaignProps as DonationConfiguration)
```
ScottiBR: There are some typescript new errors are related to this spreading here. the spread is `type` overwriting the explicit `onCancel, onDonate`. If you move {...props} to the beginning of the object, it resolves the typescript conflict and ensures our defined functions handlers take precedence.

ScottiBR (changes requested):
Nice work! A couple of follow-ups:

- TS Issues: CI is flagging a few items here: [Link](https://github.com/Adyen/adyen-web/pull/3878#issuecomment-4090267197).

- E2E Coverage: It’d be great to see tests for the Happy Path, the "Not now" flow, and an API failure (mocking a 5xx on the campaign fetch).

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -466,6 +494,14 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
         }
 
         cleanupFinalResult(result);
+
+        /** If we are using sessions (and haven't moved into a hybrid flow); and the response mandates it - create a Donation instance */
+        const isHybridFlow = this.core.session && (this.props.onSubmit || this.props.onAdditionalDetails);
+
+        if (this.core.session && !isHybridFlow && result.askDonation === true) {
```
ScottiBR: What do you think about making this code a little more declarative by defining another const for "isSessionFlow" or something flow looks like. 
```js
const isHybridFlow = this.core.session && (this.props.onSubmit || this.props.onAdditionalDetails);
const isSessionFLow = this.core.session && !isHybridFlow;

if(isSessionFLow && result.askDonation === true) { ... }

```

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -442,6 +453,23 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
         });
     };
 
+    protected setupSessionsDonation() {
+        const { amount, donation } = this.props;
+
+        // If merchant hasn't disabled autoStart (either explicitly or by not creating a donation object), make the donation component
+        if (!donation || donation.autoStart !== false) {
```
ScottiBR: The logic is a bit of a double-negative, which can be tricky to read quickly. What about something more declarative:
```js
const shouldDonationAutoStart = donation?.autoStart ?? true;
if(shouldDonationAutoStart) {...}
```

packages/lib/src/components/Donation/components/utils.ts
```diff
@@ -15,3 +17,8 @@ const getRoundupAmountLabel = (i18n: Language, { maxRoundupAmount, commercialTxA
     getAmountLabel(i18n, { value: getRoundupAmount(maxRoundupAmount, commercialTxAmount), currency });
 
 export { getAmountLabel, getRoundupAmount, getRoundupAmountLabel };
+
+export function getDonationComponent(txVariant: string, core: ICore) {
+    const DonationClass = core.getComponent(txVariant) as typeof Donation | undefined;
+    return DonationClass ? DonationClass : null;
```
ScottiBR:
```suggestion
    return DonationClass ?? null;
```

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+        this.onDonationCompleted = checkout.options.donation?.onSuccess;
+
+        this.onDonationFailed = checkout.options.donation?.onError;
+
+        this.commercialTxAmount = dcpProps.commercialTxAmount;
+
+        this.delayMS = checkout.options.donation?.delay != null ? checkout.options.donation.delay : this.autoStartTimerMS;
+
+        return this;
```
ScottiBR: The Class constructor already implicitly returns `this`.

packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -3,14 +3,62 @@ import UIElement from '../internal/UIElement/UIElement';
...
+                        // If no campaigns found (response is null), don't mount - silently do nothing
+                    })
+                    .catch((error: unknown) => {
+                        // Call merchant defined callback
+                        checkout.options.donation?.onError?.(error);
+                    });
+            } catch (error: unknown) {
+                // Silently handle duplicate instance errors - the automatic donation flow will proceed
+                console.error('Donation::DonationCampaignService instantiation error', error);
+            }
```
ScottiBR: This will trigger a new sonar issue https://sonarcloud.io/organizations/adyen/rules?open=typescript%3AS7059&rule_key=typescript%3AS7059

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    private static instanceCount: number = 0;
+
+    private readonly core: ICore;
+
+    private readonly commercialTxAmount: number = 0;
+    private readonly onDonationCompleted: (result: { didDonate: boolean }) => void;
+    private readonly onDonationFailed: (reason: unknown) => void;
+
+    private readonly autoStartTimerMS = 3000;
```
ScottiBR: Since this is a default constant, what do you think about renaming it to something like `DEFAULT_DONATION_AUTO_START_DELAY_MS `

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,160 @@
...
+            })
+            .then((donationCampaign: DonationCampaign) => {
+                if (donationCampaign) {
+                    return this.handleDonationCampaign(donationCampaign);
+                }
+                return null;
+            })
+            .catch((error: unknown) => {
+                throw error;
+            });
```
ScottiBR: This chain of promises creates some unnecessary nesting. Since `initialise` is already async/await, we can flatten this to make it more readable and avoid the "callback pyramid."

```ts
    const response = await this.makeSessionsDonationCampaignsCall();
    const [donationCampaign] = response?.donationCampaigns || [];
    return donationCampaign ? this.handleDonationCampaign(donationCampaign) : null;
}
```

This also removes a redundant .catch that was just re-throwing the same error.

ScottiBR (dismissed):

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: It is not required for a merchant to configure a `donation` object. In the happy-happy flow the merchant doesn't need to do _anything_. (They can even get the donation result from a webhook)

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: You're right. The props is not optional. Will fix

packages/lib/src/components/Donation/DonationCampaignService.tsx
ScottiBR: Since donation object is optional, i think the types for `onDonationCompleted | Failed` should reflect that as well:
```ts
private readonly onDonationCompleted?: (didDonate: boolean) => void;
// or
private readonly onDonationCompleted: (didDonate: boolean) => void | undefined;
```

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: There will only ever be one donation from a shopper per session/payment.
But I was wondering if there could be an edge case where the merchant would generate a second session and feed that into the existing instance of checkout to start a second payment flow? I'm just sanity checking that in the chat

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: if you insist. Technically speaking it is still a session flow even if the merchant has then "hybridised" it.

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: I wanted it to be specific that it _has_ to be set to false. Under your logic setting `autoStart: ''` could still be use to prevent starting automatically.
So then the clause has to be built around `donation?.autoStart !== false`

packages/lib/src/components/Donation/Donation.tsx
sponglord: > This will trigger a new sonar issue

Why isn't that error showing up in the PR?

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: Agreed

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: The merchant would really have to be trying to break things for the fact that `instanceCount` isn't reset, to come into play. 
But, for clarity's sake, I could tie resetting this static var to the donation being either completed or cancelled - which are both end states for a donation?

packages/lib/src/components/Donation/Donation.tsx
ScottiBR: Strange, this PR is was not executed in sonar, but the CI/CD say it was. 
https://sonarcloud.io/project/pull_requests_list?id=Adyen_adyen-web

packages/lib/src/components/internal/UIElement/UIElement.tsx
ScottiBR: See your point, I'm okay keeping it the original way

packages/lib/src/components/internal/UIElement/UIElement.tsx
ScottiBR: Maybe it's just me not used to the Hybrid flow, that's why felt strange reading that way. Let's keep the original code.

packages/lib/src/components/Donation/Donation.tsx
ScottiBR: Maybe if you commit something we can double check if the pipe will run, or you can try rename the branch to `(branch|release)-.*` that will force trigger the sonar scan

packages/lib/src/components/Donation/DonationCampaignService.tsx
ScottiBR: I'm ok keeping the way it is, this is totally edge case and I don't think we should put too much effort for that.

packages/lib/src/components/Donation/components/utils.ts
sponglord: Will fix in the code

packages/lib/src/components/Donation/Donation.tsx
sponglord: I've fixed the issue any way, and am no longer running async code in the constructor

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: I get the point about this specific case, and have fixed it.
But I don't understand the error generally - you could have a situation where you deliberately define default handlers, but expect that a set of props could come in and, if defined, overwrite those defaults?

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        this.core = checkout;
+
+        this.onDonationCompleted = (result: { didDonate: boolean }) => {
+            // Reset the count now the process has finished
+            DonationCampaignService.instanceCount = 0;
+
+            // Call the merchant defined handler
+            checkout.options.donation?.onSuccess(result);
```
ScottiBR:
```suggestion
            checkout.options.donation?.onSuccess?.(result);
```

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,145 @@
...
+        };
+        this.onDonationFailed = (reason: unknown) => {
+            DonationCampaignService.instanceCount = 0;
+            checkout.options.donation?.onError(reason);
```
ScottiBR:
```suggestion
            checkout.options.donation?.onError?.(reason);
```

packages/lib/src/components/Donation/DonationCampaignService.test.tsx
```diff
@@ -0,0 +1,575 @@
+import { mock, MockProxy } from 'jest-mock-extended';
+import DonationCampaignService from './DonationCampaignService';
+import type { ICore } from '../../core/types';
+import type { DonationCampaign, DonationCampaignOptions } from './types';
+import type { CheckoutSessionDonationCampaignsResponse, CheckoutSessionDonationsResponse } from '../../core/CheckoutSession/types';
+import CheckoutSession from '../../core/CheckoutSession';
+import type { IAnalytics } from '../../core/Analytics/Analytics';
+
+const createMockCore = (): MockProxy<ICore> => {
```
ScottiBR:
```suggestion
type MockCoreWithSession = MockProxy<ICore> & { session: MockProxy<CheckoutSession> };
const createMockCore = (): MockCoreWithSession => {
 return core:MockCoreWithSession;
```

ScottiBR (dismissed):
Nice, I've just catch a few typescript errors feel free to ignore those.

packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -1,23 +1,74 @@
...
+        super(checkout, initialConfig);
+
         this.donate = this.donate.bind(this);
+        this.cancel = this.cancel.bind(this);
+        this.amountSelected = this.amountSelected.bind(this);
+
+        this.isInServiceMode = isServiceMode;
+
+        if (checkout.session && isServiceMode) {
+            void this.initialiseServiceMode(checkout, props); // NOSONAR - async in constructor is intentional for fire-and-forget service mode initialization
```
ScottiBR, line 30: A suggestion here, feel free to ignore. 
You can use an [queueMicroTask](https://developer.mozilla.org/en-US/docs/Web/API/Window/queueMicrotask) for async logic in constructors as it guarantees the instance is fully constructed before the microtask runs, and it avoids needing to suppress Sonar warnings

sponglord: Interesting. I'll investigate, but not necessarily for this PR

ScottiBR (dismissed):

packages/lib/src/core/CheckoutSession/CheckoutSession.ts
```diff
@@ -110,6 +115,32 @@ class Session {
...
+            }
+
+            return response;
+        });
+    }
+
+    /**
+     * Call donations endpoint
+     */
+    donations(data: CheckoutSessionDonationsRequestData): Promise<CheckoutSessionDonationsResponse> {
```
ribeiroguilherme: Minor thing: 

Function name's in general are verbs, telling what the function is doing. You can check all other function names in this file for reference. This ones could be named such as `fetchDonations()` , `fetchDonationCampaigns()` to keep the standard way

packages/lib/src/components/Donation/types.ts
```diff
@@ -1,9 +1,44 @@
...
+export interface DonationCampaign extends CampaignContentProps {
+    id: string;
+    campaignName: string;
+    donation: Donation;
+    termsAndConditionsUrl?: string;
+}
+
+export interface DonationOptions {
+    autoStart: boolean;
+    delay: number;
```
ribeiroguilherme: There is no intellisense for `autoStart` nor `delay` . It would be nice to have since they are public API.
Can you add a JSDoc description for them?

packages/lib/src/components/Donation/types.ts
```diff
@@ -1,9 +1,44 @@
...
+    /**
+     * Optional callback when the (sessions) donation is completed (or cancelled)
+     * @param didDonate - a boolean stating whether a donation was made (true) or whether the shopper cancelled the donation (false)
+     */
+    onSuccess?: (result: { didDonate: boolean }) => void;
+    /**
+     * Optional callback when the (sessions) donation fails
+     * @param reason - the reason why the donation failed (could be an error message; or a string, stating for example, that the donation payment was refused)
+     */
+    onError?: (reason: unknown) => void;
```
ribeiroguilherme: I understand we can't force merchants to provide callbacks, but are they actually optional?  

The merchant needs to know that the donation was done, so he can move the shopper to the next page. Same applies for when a critical error happens - the merchant needs to react upon that.

I believe we should be more strict and enforce them. This is a broader discussion though that can be applied to all callbacks in general. Using a props validator like John showed us would help a lot with that.

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -442,6 +453,20 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
     };
 
+    protected setupSessionsDonation() {
+        const { amount, donation } = this.props;
+
+        // If merchant hasn't explicitly disabled autoStart by setting it to false
+        if (donation?.autoStart !== false) {
+            const rootNode: HTMLElement = assertIsDropin(this.elementRef) ? this.elementRef._node : this._node;
+
+            const DonationComponentRef = getDonationComponent(TxVariants.donation, this.core);
```
ribeiroguilherme: Why not calling `this.core.getComponent(TxVariants.donation)` directly? 

If that is a Typescript issue, I don't see a problem in doing  ` this.core.getComponent(TxVariants.donation) as typeof Donation` . We are 100% sure it is Donation and also that it must not be null

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -466,6 +491,15 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
         }
 
         cleanupFinalResult(result);
+
+        /** If we are using sessions (and haven't moved into a hybrid flow); and the response mandates it - create a Donation instance */
+        const isHybridFlow = this.core.session && (this.props.onSubmit || this.props.onAdditionalDetails);
+        const isUnalteredSessionFLow = this.core.session && !isHybridFlow;
+
+        if (isUnalteredSessionFLow && result.askDonation === true) {
```
ribeiroguilherme: Correct me if I am wrong, but I don't think these checks are needed:  

`askDonation` is returned only for Sessions flow, meaning that the merchant is not using `onSubmit` nor ` onAdditionalDetails` to finalize the payment. I believe we could just check the `result.askDonation` flag in this case which simplifies the code.

If we end up finding out that there is an issue about this, then we can implement this `hybrid` flow check, but I believe it is not needed here.

packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -1,23 +1,74 @@
...
+     *
+     * The "service" mode is used in the /sessions flow:
+     * - if the /payments response mandates it, we consult the /sessions/donationCampaigns endpoint on behalf of the merchant, and mount any found Donation Campaigns.
+     *
+     * The opposite of this mode is the "direct" mode, used in the Advanced flow:
+     * - the merchant makes their own call to the /donationCampaigns endpoint and provides the Donation Campaign directly to the component.
+     */
+    private static isServiceMode(props: DonationProps): props is DonationCampaignOptions {
+        return 'rootNode' in props && !!props.rootNode;
     }
```
ribeiroguilherme, line 45: Isn't it safe enough to just check if there is a `checkout.session`, then we use the SessionsMode? 
It is more straight forward/clean than checking a component prop

packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -37,20 +88,67 @@ class DonationElement extends UIElement<DonationConfiguration> {
...
+
+        this.props.onCancel({ data, isValid });
+    }
+
+    amountSelected(state: DonationPayload) {
+        // Send analytics
+        const event = new AnalyticsInfoEvent({
+            component: DonationElement.type,
+            type: InfoEventType.selected,
+            selectedValue: JSON.stringify(state.data.amount),
```
ribeiroguilherme, line 136: Have you checked with data team about passing a JSON as string for this value?  Perhaps just the amount is enough, and easier to parse for them.

packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -1,23 +1,74 @@
...
 
-    constructor(checkout: ICore, props?: DonationConfiguration) {
-        super(checkout, props);
+    private readonly isInServiceMode: boolean = false;
+
+    constructor(checkout: ICore, props: DonationProps) {
+        const isServiceMode = DonationElement.isServiceMode(props);
+        const initialConfig = isServiceMode ? undefined : props;
+
+        super(checkout, initialConfig);
```
ribeiroguilherme, line 21: I am not sure if we need this check at the top:

You can initialize the Component as it is, with any props passed to it. Then, if there is a `checkout.session`,  we initialize the DonationService which will eventually override props that were defined previously.

packages/lib/src/components/Donation/components/utils.ts
```diff
@@ -15,3 +17,8 @@ const getRoundupAmountLabel = (i18n: Language, { maxRoundupAmount, commercialTxA
     getAmountLabel(i18n, { value: getRoundupAmount(maxRoundupAmount, commercialTxAmount), currency });
 
 export { getAmountLabel, getRoundupAmount, getRoundupAmountLabel };
+
+export function getDonationComponent(txVariant: string, core: ICore) {
```
ribeiroguilherme: I guess this utility function is not needed - I left a comment where it is being used.

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,145 @@
...
+        };
+
+        this.commercialTxAmount = donationCampaignProps.commercialTxAmount;
+
+        this.delayMS = checkout.options.donation?.delay ?? DEFAULT_DONATION_AUTO_START_DELAY_MS;
+    }
+
+    public async initialise(): Promise<DonationConfiguration | null> {
+        await new Promise(resolve => setTimeout(resolve, this.delayMS));
+        return this.callSessionsDonationCampaigns();
```
ribeiroguilherme: You don't need to wait for the timeout to start fetching the donations campaing, right? 

Best practice here would be to pre-fetch it, and then use it once the timeout is over.

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,145 @@
...
+            throw new Error(
+                'The donation type is "roundup" and the commercialTxAmount is not set.\nIt will not be possible to mount a Donation component.'
+            );
+        }
+
+        return donationComponentProps;
+    }
+
+    private callSessionsDonations(donationRequestData: CheckoutSessionDonationsRequestData, component: Donation) {
+        this.makeSessionDonationsCall(donationRequestData)
```
ribeiroguilherme: Minor thing- I noticed we’re using a mix of Promise.then() and async/await in this file. To keep the logic easy to follow and the style consistent, should we pick one (ideally async/await) and stick with it throughout this module?

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    private readonly delayMS: number;
+
+    constructor(checkout: ICore, donationCampaignProps: DonationCampaignOptions) {
+        DonationCampaignService.instanceCount++;
+
+        // If the merchant has not explicitly set autoStart to false, and then tries to display the Donation component in a different container,
+        // we need to warn them otherwise, without this check, 2 calls to the /donationCampaigns endpoint will be made - since UIElement
+        // will automatically mount the Donation component in the component's rootNode.
+        if (DonationCampaignService.instanceCount > 1) {
```
ribeiroguilherme: Couldn't the request to the `/donationsCampaing` be cached then?
Then all this logic and complexity of counting instances +  throwing errors will go away

ribeiroguilherme (commented):
I do like the solution. It is way better than what we discussed initially! 

The only thing that I am missing is to be able to test it visually. It is pretty hard to review a code without being able to actually run it and see what is happening. Some of my comments might be based on assumptions which I can't confirm without testing it visually.

This can be frustrating, particularly to you, since more comments likely can arise during the visual test from the reviewers. I'd suggest to have stories ready before moving the code to review next time 🙏

packages/lib/src/components/Donation/types.ts
sponglord: Are you saying we need to force a merchant to provide the parent `donation` config object? And then also enforce that they define callbacks within it?
The Impact team are foreseeing a Happy flow where the merchant doesn't have to do anything.
But I see your point

packages/lib/src/components/Donation/types.ts
sponglord: ok

packages/lib/src/core/CheckoutSession/CheckoutSession.ts
sponglord: fair enough

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: you're right

packages/lib/src/components/internal/UIElement/UIElement.tsx
sponglord: yeah, you're right

packages/lib/src/components/Donation/components/utils.ts
sponglord: addressed above

packages/lib/src/components/Donation/Donation.tsx
sponglord: Without this check you have to do this: `super(checkout, props as DonationConfiguration)` in order to get TS to accept the `props` being sent to the `super`.
And this is just fooling the compiler 'cos `props` might be of type `DonationCampaignOptions`.

The way it stands atm you are ensuring that the only props ever set on `super`, or set later as `this.props,` are _actually_ of type `DonationConfiguration`

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: fair enough

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: ok

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: I'm not a big fan of this static var.
The problem it's trying to avoid is the creation, and mounting, of two Donation components. 
When the second one is created (because the merchant forgot to set `autoStart:false`), the second instance needs to know the first once exists (and has already called `/donationsCampaigns` and mounted in the default `rootNode`),  and so the second instance shouldn't proceed anymore with its initialisation.
I'm thinking of trying another approach: storing a `donationInitiated` flag on either the `checkout` instance (or the `checkout.session`).
This will stop it being global and static and instead scope it to the checkout (or session) instance.
What do you think?

packages/lib/src/components/Donation/types.ts
sponglord: I'm changing the type to say that a donation config object doesn't have to be added but, if it is, then it requires callbacks

packages/lib/src/components/Donation/DonationCampaignService.tsx
sponglord: I've gone for the flag stored on `core.session`. This keeps it scoped to the instance and so it automatically resets with the creation of a new session. It cleans up `DonationCampaignService` and means we can stop throwing errors and just have a simple `if-else` type condition in `Donation.tsx`

packages/lib/src/components/Donation/components/DonationComponent.tsx
```diff
@@ -33,6 +33,8 @@ export default function DonationComponent(props: Readonly<DonationComponentProps
         const value = parseInt(target.value, 10);
         setIsValid(true);
         setAmount((amount: DonationAmount) => ({ ...amount, value }));
+
+        onAmountSelected({ data: { amount: { ...amount, value } } });
```
johnayeni, line 37:
```suggestion
        onAmountSelected?.({ data: { amount: { ...amount, value } } });
```

This callback is optional right? If so this would break if the callback does not exist

packages/lib/src/components/Donation/Donation.scss
```diff
@@ -1,7 +1,19 @@
 @use 'styles/index';
 @use 'styles/variable-generator';
 
+@keyframes adyen-checkout-donation-fade-in {
```
johnayeni, line 4: This could be nice to be a shared keyframes animation. Maybe we can create a new `styles/animation` scss sheet and import from there

packages/lib/src/components/Donation/Donation.tsx
```diff
@@ -37,20 +89,67 @@ class DonationElement extends UIElement<DonationConfiguration> {
...
     protected override componentToRender(): h.JSX.Element {
         return (
-            <Fragment>
-                {/*@ts-ignore ref*/}
-                <DonationComponent {...this.props} ref={this.handleRef} onChange={this.setState} onDonate={this.donate} />
-            </Fragment>
+            <DonationComponent
+                {...this.props}
+                /*@ts-ignore ref*/
+                ref={this.handleRef}
```
johnayeni: Can we refactor this to use `this.setComponetRef` instead?

packages/lib/src/components/Donation/components/DonationComponent.tsx
sponglord: The type was wrong. `onAmountSelected` is always defined. I'll fix the type

packages/lib/src/components/Donation/Donation.tsx
sponglord: Strictly speaking not my code. But I'll fix it

packages/lib/src/components/Donation/stories/Donation.stories.tsx
```diff
@@ -14,6 +14,7 @@ const componentConfiguration: DonationConfiguration = {
     nonprofitUrl: 'https://example.org',
     nonprofitDescription:
         'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
+    //@ts-ignore type is correct
     donation: {
         type: 'fixedAmounts',
         currency: 'EUR',
```
johnayeni, line 20: Would it be possible to add new stories as examples for this new donations flow?

packages/lib/src/components/Donation/Donation.scss
sponglord: If a second use case comes up it can be moved into it's own sheet then

packages/lib/src/components/Donation/stories/Donation.stories.tsx
sponglord: There are new stories in the Dropin for this

packages/lib/src/components/Dropin/stories/dropin/Dropin.stories.tsx
```diff
@@ -121,4 +125,88 @@ export const SplitFundingBrazil = {
...
+
+        return (
+            <Checkout checkoutConfig={checkoutConfig}>
+                {checkout => <ComponentContainer element={new DropinComponent(checkout, componentConfiguration)} />}
+            </Checkout>
+        );
+    }
+};
+
+export const SessionsDonationReparented: DropinStory = {
```
johnayeni, line 154:
```suggestion
export const SessionsDonationWithCustomDialog: DropinStory = {
```

This is honestly just a preference so up to you if you think this makes sense

johnayeni (approved):

packages/lib/src/components/Dropin/stories/dropin/Dropin.stories.tsx
sponglord: I think this makes sense with how we're describing it in the docs i.e. "reparenting" the donation component

PR #3895, opened 2026-04-03
Translations - Adding enUS to bundle and Locale CDN request improvements

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

Adding back the v5 mechanism which would perform "smart" assumption to decide a locale in case a valid supported locale is not informed. Examples:
-> `en-GB` fallbacks to `en-US`
-> `us`fallbacks to `en-US`
-> `nl` fallbacks to `nl-NL`
-> `es`fallbacks to `es-ES` and so on..

Besides that, we are adding `en-US` as part of the bundle, so we always have a fallback if request fails. This leads to bundle size increase, which was also adjusted in the size-limit configuration.

The logic is:
- if a locale in the expected format is informed (e.g. `en-GB`), we try to match with our supported locales. If there is no match, we try to match the first part (`en`), which leads to using the `en-US` locale
- if a locale is not in the expected format (e.g. `nl`) , we try to match the first part with our available locales, leading to use the `nl-NL`
- If we don't manage to match any locale OR if locale matches with the English one, we do not make any request to CDN and we use the built-in locale
- We make a request to the CDN only if the parsed locale  is available there. Therefore it is expected to not see more 404's from this release version onwards
- If an unrecognized locale is passed (e.g. `fr-CA`) , and its custom translations are also provided, we do consider it as a valid `locale`. We will internally request the `fr-FR` locale to be used, and the custom translations from `fr-CA` will be applied on top of the French translations
- If the translation request fails, we fallback to using the English translations

---

## 🧪 Tested scenarios
- Added tests for the Language module, Language Service and also the new utilities that helps parsing the locale
- Tested manually setting different locales, checking the outcome and network requests

---


packages/lib/src/language/utils.ts
```diff
@@ -1,54 +1,81 @@
...
+    return Object.keys(customTranslations).reduce<CustomTranslations>((memo, customTranslationLocaleKey) => {
         const locale = formatLocale(customTranslationLocaleKey);
+        if (!locale) return memo;
         memo[locale] = customTranslations[customTranslationLocaleKey];
         return memo;
     }, {});
 }
 
-const replaceTranslationValues = (translation, values) => {
+const replaceTranslationValues = (translation: string, values: { [key: string]: string }) => {
```
johnayeni:
```suggestion
const replaceTranslationValues = (translation: string, values: Record<string, string>) => {
```

packages/lib/src/language/utils.ts
```diff
@@ -1,54 +1,81 @@
...
-        return fullLocale;
-    } else {
-        throw new AdyenCheckoutError('IMPLEMENTATION_ERROR', `Locale '${localeParam}' does not match the expected format`);
-    }
+
+    return fullLocale.length === 5 ? fullLocale : null;
+}
+
+export function parseLocale(locale: string, supportedLocales: readonly string[]): string {
+    if (!locale || locale.length < 1 || locale.length > 5) return DEFAULT_LOCALE;
```
johnayeni:
```suggestion
    if (!locale || locale.length < LOCALE_NAME_MIN_LENGTH || locale.length > LOCALE_NAME_MAX_LENGTH) return DEFAULT_LOCALE;
```

Can we move these values to constants so that we can add descriptions as to why we have them e.g

```ts
/**
   @description Maximum length is 5 because ISO 3166-1 country/region codes can be a max of 5 characters e.g en-US
*/
const LOCALE_NAME_MAX_LENGTH = 5
```

packages/lib/src/language/utils.ts
```diff
@@ -1,54 +1,81 @@
...
+ * @param localeParam - Locale example: 'En_us' or 'en-US'
  *
  * @example
- * formatLocale('En_us'); -> en-US
- * formatLocale('ar') -> ar
+ * formatLocale('En_us');
+ * // returns 'en-US'
  */
-export function formatLocale(localeParam: string): string {
+export function formatLocale(localeParam: string): string | null {
```
johnayeni:
```suggestion
export function formatLocaleToCountryRegionLocale(localeParam: string): string | null {
```

packages/lib/src/language/utils.ts
```diff
@@ -1,54 +1,81 @@
 import { h } from 'preact';
 import { CustomTranslations } from './types';
-import AdyenCheckoutError from '../core/Errors/AdyenCheckoutError';
+import { DEFAULT_LOCALE } from './constants';
 
 /**
- * Returns a locale with the proper format
+ * Convert to ISO 639-1
```
johnayeni:
```suggestion
 * @description Convert to ISO 639-1
```

packages/lib/src/language/LanguageService.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        try {
+            return await httpGet<Translations>({
+                loadingContext: this.cdnUrl,
+                errorLevel: 'fatal',
+                errorMessage: `Translations: Failed to fetch translations for locale "${cdnLocale}"`,
+                path: `sdk/${this.sdkVersion}/translations/${cdnLocale}.json`
+            });
+        } catch (error) {
+            console.log(`LanguageService - fetchTranslationsFromCdn(): Failed to fetch locale "${cdnLocale}."`);
```
johnayeni:
```suggestion
            console.warn(`LanguageService - fetchTranslationsFromCdn(): Failed to fetch locale "${cdnLocale}."`);
```

So the merchants error reporting tool can catch this

packages/lib/src/language/Language.ts
```diff
@@ -1,44 +1,79 @@
...
-            ...(!!this.customTranslations[this.locale] && this.customTranslations[this.locale])
-        };
+    public async requestTranslations(): Promise<void> {
+        try {
+            const translations = await this.service.fetchTranslationsFromCdn(this.locale);
+
+            this._translations = {
+                ...enUS,
+                ...translations,
+                ...(!!this.customTranslations[this.locale] && this.customTranslations[this.locale])
```
johnayeni:
```suggestion
                ...(Boolean(this.customTranslations[this.locale]) && this.customTranslations[this.locale])
```

packages/lib/src/language/constants.ts
```diff
@@ -1 +1,36 @@
+export const CDN_SUPPORTED_LOCALES = [
```
sponglord, line 1: Just thinking out loud: this list, which needs to be manually maintained, could be a point of failure if we add a new translations .json, and forget to reference it here.

Is there any way we could have some test in the CI that matches the translations files in the `server/translations` directory to this list and throws an error if they don't match?
Or do we think Gemini will pick up on this sort of thing? (Or can we guide Gemini to look for this?)

packages/lib/src/language/Language.ts
```diff
@@ -1,44 +1,79 @@
...
-        const { locale, translations, customTranslations } = props;
+        const { locale, customTranslations, service } = props;
 
-        if (!locale) {
-            throw new AdyenCheckoutError('IMPLEMENTATION_ERROR', 'Language: "locale" property is not defined');
-        }
+        this.service = service;
+
+        this.customTranslations = formatCustomTranslations(customTranslations);
+        this.supportedLocales = this.createSupportedLocalesList(customTranslations);
```
sponglord: Found an edge-case scenario that doesn't work as expected. Are you ready... :-p

If the merchant specifies an unsupported `locale`, in the wrong case, e.g. `nl-ar` this gets mapped to `nl-NL`, as expected.

But if the merchant also specifies some _custom translations_ 
- if they write them in the correct case:
```
 'nl-AR': {
    'creditCard.cardNumber.label': 'number label',
    'creditCard.expiryDate.label': 'date label'
}
```
then they do get picked up and applied on top of the Dutch translations - as expected.

- But, if they write them in the wrong case:
```
 'nl-ar': {
    'creditCard.cardNumber.label': 'number label',
    'creditCard.expiryDate.label': 'date label'
}
```
they don't get applied.

I'm not quite sure what's going wrong - I can see that in `language/utils.ts`, in the `formatCustomTranslations` function, the `customTranslationLocaleKey` does get formatted to the right case (`nl-AR`), but the actual custom translations are not picked up.

UPDATE:
I think this line needs to be:
`this.supportedLocales = this.createSupportedLocalesList(this.customTranslations);`

i.e. you need to apply what is returned from the call to `formatCustomTranslations()`, on the previous live, to this next step

(I've tested locally and it does fix it)

packages/lib/src/core/core.ts
```diff
@@ -433,13 +414,35 @@ class Core implements ICore {
...
+     * Retrieve the parsed locale from the i18n module and assigns it to the core options.
+     */
+    private assignLocaleToCore(): void {
+        this.options.locale = this.modules.i18n.locale;
+    }
+
+    /**
+     * Request translations to the CDN. If an error occurs, it will be logged to the console in development mode.
+     */
+    private async requestTranslations(): Promise<void> {
```
ScottiBR, line 438: Currently, we have three layers of `try/catch` blocks (here, inside modules and inside `fetchTranslationsFromCdn` ), each logging the same error in different ways. This makes debugging more confusing. 

Since these intermediate functions don't have any unique recovery logic, letting the error bubble up naturally (or just trusting the bottom-layer log) keeps the code much cleaner and our logs more accurate.

Suggestion: only Keep the try/catch and logging inside `fetchTranslationsFromCdn`. This is the "source of truth" for the network request, and error logs.

ScottiBR (changes requested):

packages/lib/src/language/Language.ts
```diff
@@ -1,44 +1,79 @@
-import { getTranslation } from './utils';
+import { formatCustomTranslations, getTranslation, parseLocale } from './utils';
```
johnayeni, line 1: Should we not have this `language` folder inside `core`? I would consider this as a module as well like the rest of the other modules like Analytics and Risk

The PIE team are doing the same https://github.com/Adyen/adyen-platform-experience-web/tree/develop/src/core/Localization

This might not necessarily be in the scope of this PR.

packages/lib/src/language/utils.ts
```diff
@@ -1,54 +1,81 @@
...
+    return matchLocale(formattedLocale || locale, supportedLocales) || DEFAULT_LOCALE;
 }
 
 /**
  * Makes sure that if custom translation is defined using not properly formatted locale keys, then it gets formatted correctly
- * Ex: Custom translation defined as { en_US: { ... }} will be adjusted to { 'en-US': { ... }}
+
+ * Custom translation defined as { en_US: { ... }} will be adjusted to { 'en-US': { ... }}
  */
 export function formatCustomTranslations(customTranslations: CustomTranslations = {}): CustomTranslations {
```
johnayeni, line 77: There is a bug where custom translations for two letter code locales only don't work

```ts
locale: 'ar',
        translations: {
            ar: {
                'creditCard.cardNumber.label': 'number label',
                'creditCard.expiryDate.label': 'date label'
            }
        }
```

The bug is in `formatCustomTranslations()` . It calls formatLocale() on every custom translation key, but formatLocale() returns null for two-letter codes like "ar" (line 45: no country code → null). So the entire `ar` translations block is silently dropped.

Meanwhile, CDN_SUPPORTED_LOCALES does include 'ar' as a valid two-letter locale. And parseLocale('ar') resolves this.locale to 'ar' correctly.

johnayeni: A fix I tried is to update the `formatCustomTranslations` method to this

```ts
export function formatCustomTranslations(customTranslations: CustomTranslations = {}): CustomTranslations {
    return Object.keys(customTranslations).reduce<CustomTranslations>((memo, customTranslationLocaleKey) => {
        const locale = formatLocale(customTranslationLocaleKey) || toTwoLetterCode(customTranslationLocaleKey);
        memo[locale] = customTranslations[customTranslationLocaleKey];
        return memo;
    }, {});
}
```

packages/lib/src/language/utils.ts
ribeiroguilherme: Is there any benefit in using this here?

packages/lib/src/language/LanguageService.ts
ribeiroguilherme: What do you mean? Couldn't it also report from normal logs or are there filters for this?

packages/lib/src/language/constants.ts
ribeiroguilherme: Oh yeah, good point. I will add a check for this

packages/lib/src/language/Language.ts
ribeiroguilherme: Good catch. I checked on v5 and it indeed uses the parsed `customTranslations` rather then the raw object. I will add a test case for it too

packages/lib/src/language/Language.ts
ribeiroguilherme: I thought about moving it there indeed, but then it would be out of the scope of this PR

packages/lib/src/language/constants.ts
ribeiroguilherme: I added a CI job for that

packages/lib/src/language/utils.ts
ribeiroguilherme: Good catch. I pushed a fix for it with your suggestion and also added a test case for it

packages/lib/src/core/core.ts
ribeiroguilherme: My logic here was that:
- if the request fails on `LanguageService`, then we would handle it accordingly and fallback to English
- if the spread/assignment logic fails somehow in the `requestTranslations`, we are covered and we have at least a fallback working
- if there is any unforeseen issue that bubbles up to the core, at least the error is caught and not let an unhandled error being thrown at the merchant environment (as it happened with PBL)

Perhaps the try/catch block at the `requestTranslations()` can be removed as it seems that the spread operation is defensive and safe enough, but I'd rather keep the try/catch on the top level to not let any possible unhandled error flood into merchants environment.

packages/lib/src/language/LanguageService.ts
johnayeni: This is from Gemini

---

By default, most modern error-tracking tools like Sentry, LogRocket, or Datadog are designed to capture "breadcrumbs" rather than just reporting every log as a standalone issue. Here is how the hierarchy usually works:

1. Console Warns & Errors (The "Breadcrumbs")
Most SDKs automatically "monkey patch" the global console object. When your code executes a console.warn() or console.error(), the tool records it as a breadcrumb.

Breadcrumbs are a timeline of events leading up to a crash.

They are stored locally in the browser/app memory and are only sent to the server if an actual crash or unhandled exception occurs.

This prevents your dashboard from being flooded with thousands of minor warnings while still providing context for why a "real" error happened.

2. Console Logs (The "Verbose" Data)
Standard console.log() statements are often ignored by default or filtered out in production environments to save on bandwidth and noise.

Performance: Sending every console.log to a remote server would drastically slow down your application.

Privacy: Logs often contain sensitive debug info that shouldn't be sent to a third-party processor.

Configuration: You can usually toggle a setting (e.g., captureConsole: true) to include them as breadcrumbs, but they rarely trigger an "Alert" on their own.

packages/lib/src/language/utils.ts
johnayeni: Actually, here this is an internal method so no real benefit

johnayeni (dismissed):

packages/lib/src/core/core.ts
ribeiroguilherme: > Perhaps the try/catch block at the requestTranslations() can be removed as it seems that the spread operation is defensive and safe enough, but I'd rather keep the try/catch on the top level to not let any possible unhandled error flood into merchants environment.

I updated the code according to this change

johnayeni (approved):

ScottiBR (approved):
LGTM

PR #3900, opened 2026-04-10
Update SDKData

dimitra-adyen:

---

## 📝 Summary

- Added the below four new fields to sdkData:

    - channel (value: web)
    - platform (value: web)
    - sdkVersion 
    - paymentMethodBehavior (value: `nativeComponent` for payment methods with dedicated component or `genericComponent` for the rest)
 
- For analytics, updated the values for channel and platform from 'Web' to 'web' for consistency. 

---

## 🧪 Tested scenarios

- Tested for UnionPay:
   sdkData contains: "channel":"web","platform":"web","sdkVersion":"6.33.0","paymentMethodBehavior":"genericComponent"
- Tested for Twint:
  sdkData contains:  "channel":"web","platform":"web","sdkVersion":"6.33.0","paymentMethodBehavior":"nativeComponent" 

---

COSDK-1150

**Closes**:  

---


packages/lib/src/core/config.ts
```diff
@@ -46,3 +46,21 @@ export default {
...
+ * @internal
+ */
+export const PLATFORM = { WEB: 'web' } as const;
+export type PLATFORM = (typeof PLATFORM)[keyof typeof PLATFORM];
+
+/**
+ * @internal
+ */
+export const PAYMENT_METHOD_BEHAVIOR = { NATIVE: 'nativeComponent', GENERIC: 'genericComponent' } as const;
+export type PAYMENT_METHOD_BEHAVIOR = (typeof PAYMENT_METHOD_BEHAVIOR)[keyof typeof PAYMENT_METHOD_BEHAVIOR];
```
johnayeni:
```suggestion
export type PaymentMethodBehavior = (typeof PAYMENT_METHOD_BEHAVIOR)[keyof typeof PAYMENT_METHOD_BEHAVIOR];
```

To keep the case type we use for types consistent

packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -160,8 +160,8 @@ class Analytics implements IAnalytics {
         }
 
         const payload: AnalyticsEventPayload = {
-            channel: 'Web',
-            platform: 'Web',
+            channel: CHANNEL.WEB,
+            platform: PLATFORM.WEB,
```
johnayeni: Did data confirm this one yet that its fine to change the case of this fields being sent for analytics?

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -89,12 +91,12 @@ abstract class BaseElement<P extends BaseElementProps> implements IBaseElement {
     public get data(): PaymentData {
         const order = this.state.order || this.props.order;
         const componentData = this.formatData();
-
+        const isGeneric = !getComponentNameOfPaymentType(componentData.paymentMethod?.type);
```
johnayeni:
```suggestion
        const doesPaymentMethodHaveNativeComponent = Boolean(getComponentNameOfPaymentType(componentData.paymentMethod?.type));
```

packages/lib/src/components/internal/BaseElement/BaseElement.ts
```diff
@@ -89,12 +91,12 @@ abstract class BaseElement<P extends BaseElementProps> implements IBaseElement {
...
         const clientData = this.core.modules.risk.data;
         const checkoutAttemptId = this.core.modules.analytics.checkoutAttemptId ?? NO_CHECKOUT_ATTEMPT_ID;
+        const paymentMethodBehavior = isGeneric ? PAYMENT_METHOD_BEHAVIOR.GENERIC : PAYMENT_METHOD_BEHAVIOR.NATIVE;
```
johnayeni:
```suggestion
        const paymentMethodBehavior = doesPaymentMethodHaveNativeComponent ? PAYMENT_METHOD_BEHAVIOR.NATIVE : PAYMENT_METHOD_BEHAVIOR.GENERIC 
```

.changeset/soft-houses-post.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+New: The payload returned in the `onSubmit` callback has increased in size, because of changes to the included `sdkData` field.
```
johnayeni, line 5: This changelog message should describe the change being made not necessarily the indication e.g `Include data about natively supported payment methods in SDK data`. SDK data is also an encoded json so nothing really changes for the merchant in terms of how to handle `onSubmit`

packages/lib/src/core/Analytics/Analytics.ts
dimitra-adyen: Yes, we verified with Olamide, along with @jreij. Turns out that the backend already passes the above as lowercase, so nothing will change for data.

packages/lib/src/core/config.ts
dimitra-adyen: Updated. PLATFORM and CHANNEL types are also in Pascal Case now

packages/lib/src/components/internal/BaseElement/BaseElement.ts
dimitra-adyen: Updated.

packages/lib/src/components/internal/BaseElement/BaseElement.ts
dimitra-adyen: Updated.

packages/lib/src/core/config.ts
```diff
@@ -46,3 +46,21 @@ export default {
...
+ * @internal
+ */
+export const CHANNEL = { WEB: 'web' } as const;
+export type Channel = (typeof CHANNEL)[keyof typeof CHANNEL];
+
+/**
+ * @internal
+ */
+export const PLATFORM = { WEB: 'web' } as const;
+export type Platform = (typeof PLATFORM)[keyof typeof PLATFORM];
```
ribeiroguilherme: Is there any benefit in having this as a dictionary? 

I don't think we will end up adding more values to this at some point. Perhaps it can be just a const that is exported.

.changeset/soft-houses-post.md
dimitra-adyen: Changelog message added per the [ticket description](https://youtrack.is.adyen.com/issue/COSDK-1150/Web-Enhance-SDK-data-to-support-hosted-components). I checked if additional details should be provided, however it will remain as-is since its purpose is to notify merchants about possible impact (e.g. those with backend settings enforcing max payload length) due to the increased payload size.

packages/lib/src/core/config.ts
```diff
@@ -46,3 +46,21 @@ export default {
...
+
+/**
+ * @internal
+ */
+export const PAYMENT_METHOD_BEHAVIOR = { NATIVE: 'nativeComponent', GENERIC: 'genericComponent' } as const;
+export type PaymentMethodBehavior = (typeof PAYMENT_METHOD_BEHAVIOR)[keyof typeof PAYMENT_METHOD_BEHAVIOR];
```
ribeiroguilherme: Something to try: you can leverage an `enum` type/object here. 

```ts
enum PaymentMethodBehavior {
  Native: 'nativeComponent',
  Generic: 'genericComponent'
}
```

I believe this approach avoids the need for such a specific exported type and reduces cognitive load, as the current type feels a bit overengineered for this use case.

dimitra-adyen: I had used enums initially but I figured that they are not tree shaken, and they were causing a slightly higher bundle size that's why I switched to the const. If you don't consider it an issue, I'm also in favor of that approach so happy to update my implementation.

dimitra-adyen: Replaced them with enums.

packages/lib/src/core/config.ts
dimitra-adyen: CHANNEL is an enum now, since in the future we are planning on having additional values and there, and PLATFORM is a const with a string value.

packages/lib/src/core/config.ts
```diff
@@ -69,3 +69,24 @@ export default {
...
+    WEB = 'web'
+}
+
+/**
+ * @internal
+ */
+export const PLATFORM = 'web';
+
+/**
+ * @internal
```
sponglord, line 89: Can you add an explanation to the comment saying what these actually mean?
A `nativeComponent` means what? That there is a 3rd party app available on the device that can take over the payment flow?
So `genericComponent` means we continue to handle the payment in the web sdk?

dimitra-adyen: Added explanation of PAYMENT_METHOD_BEHAVIOR values

packages/lib/src/core/config.ts
```diff
@@ -69,3 +69,26 @@ export default {
...
+}
+
+/**
+ * @internal
+ */
+export const PLATFORM = 'web';
+
+/**
+ * nativeComponent: payment method explicitly supported with a dedicated component by the SDK
+ * genericComponent: payment method not explicitly supported by the SDK and handled in default fallback component
```
sponglord: Can we just be a bit more specific & mention that for web the "default fallback component" is the Redirect component e.g. 
`genericComponent: payment method not explicitly supported by the SDK and handled in default fallback component i.e. the Redirect component`
or
`genericComponent: payment method not explicitly supported by the SDK and handled in default fallback component (Redirect)`

dimitra-adyen: I updated the comment to mention Redirect as the fallback component. That said, I recommend that we avoid referencing specific component names in comments in the future since it isn't a good practice and requires extra maintenance effort, otherwise they can become misleading.

johnayeni (dismissed):

johnayeni (dismissed):

johnayeni (approved):

PR #3906, opened 2026-04-14
Fix/explicit anys in card

sponglord:
## 📋 Pull Request Checklist

---

## 📝 Summary

Fix explicit `any` in `Card.tsx` and it's children

---

## 🧪 Tested scenarios

---

COSDK-1065


packages/lib/src/components/Card/components/CardInput/components/types.ts
```diff
@@ -1,6 +1,9 @@
-import { BrandConfiguration, CardBrandsConfiguration } from '../../../types';
+import { BrandConfiguration, CardBrandsConfiguration, DualBrandSelectElement } from '../../../types';
 import { ComponentChildren } from 'preact';
 import { CVCPolicyType, DatePolicyType } from '../../../../internal/SecuredFields/lib/types';
+import { SFPErrorMap, SFPValid } from '../../../../internal/SecuredFields/SFP/types';
+
+export type DualBrandingChangeHandler = (brandValue: string) => void;
```
m1aw: For keep things consistent should this be moved to '../../../types' so it sits next to DualBrandSelectElement?

packages/lib/src/components/Card/components/CardInput/defaultProps.ts
```diff
@@ -42,14 +42,14 @@ export default {
...
-    onConfigSuccess: (): any => {},
-    onAllValid: (): any => {},
-    onFieldValid: (): any => {},
-    onBrand: (): any => {},
-    onError: (): any => {},
-    onBinValue: (): any => {},
-    onBlur: (): any => {},
-    onFocus: (): any => {},
-    onChange: (): any => {}
+    onLoad: (): void => {},
```
m1aw: Some of these functions receive some arguments, but I guess this is besides the scope of explicit `anys`.

packages/lib/src/components/Card/components/CardInput/validate.ts
```diff
@@ -1,4 +1,4 @@
-import { ValidatorRules } from '../../../../utils/Validator/types';
+import { ValidateFunction, ValidatorMode, ValidatorRules } from '../../../../utils/Validator/types';
```
m1aw: I was playing around with Typescript because I didn't like the way we were returning undefined here for what should be very much a defined type.

So I reached something like this:

```suggestion
import { ValidateFunction, ValidatorMode, ValidatorRule, ValidatorRules } from '../../../../utils/Validator/types';
import { formatCPFCNPJ } from '../../../internal/SocialSecurityNumberBrazil/utils';
import validateSSN from '../../../internal/SocialSecurityNumberBrazil/validate';
import { isEmpty } from '../../../../utils/validator-utils';
import {
    BOLETO_SOCIAL_SECURITY_NUMBER_INVALID,
    CREDITCARD_HOLDER_NAME_INVALID,
    CREDITCARD_TAX_NUMBER_INVALID
} from '../../../../core/Errors/constants';

export const cardInputFormatters = {
    socialSecurityNumber: formatCPFCNPJ
};

export const cardInputValidationRules = {
    socialSecurityNumber: [
        {
            modes: ['blur'],
            validate: value => {
                if (isEmpty(value)) return null;
                return validateSSN(value);
            },
            errorMessage: BOLETO_SOCIAL_SECURITY_NUMBER_INVALID
        }
    ],
    taxNumber: [
        {
            modes: ['blur'],
            validate: value => (isEmpty(value) ? null : value?.length === 6 || value?.length === 10),
            errorMessage: CREDITCARD_TAX_NUMBER_INVALID
        }
    ],
    holderName: [
        {
            // Will fire at startup and when triggerValidation is called and also applies as text is input
            modes: ['blur'],
            validate: value => (isEmpty(value) ? null : true), // true, if there are chars other than spaces
            errorMessage: CREDITCARD_HOLDER_NAME_INVALID
        }
    ],
    default: [
        {
            modes: ['blur'],
            // ensuring we don't try to run this against objects e.g. billingAddress
            validate: value => !!value && typeof value === 'string' && value.trim().length > 0
        }
    ]
} as const satisfies ValidatorRules;

export const getRuleByNameAndMode = (name: keyof typeof cardInputValidationRules, mode: ValidatorMode): ValidateFunction => {
    const ruleOrRules = cardInputValidationRules[name];
    const ruleArr = (Array.isArray(ruleOrRules) ? ruleOrRules : [ruleOrRules]) as ValidatorRule[];
    const rule = ruleArr.reduce((acc, elem) => {
        if (!acc.length) {
            if (elem.modes.includes(mode)) {
                acc.push(elem.validate);
            }
        }
        return acc;
    }, [] as ValidateFunction[]);
    return rule[0];
};
```

This should make it possible to make sure that name is one of the keys of  the object `cardInputValidationRules`.

packages/lib/src/components/Card/components/CardInput/validate.test.ts
```diff
@@ -5,20 +5,20 @@ const taxNumValidationFn = getRuleByNameAndMode('taxNumber', 'blur');
 
 describe('validate holder name', () => {
     test('validates a card holder name', () => {
-        expect(holderNameValidationFn('')).toBe(null);
-        expect(holderNameValidationFn('John Smith')).toBe(true);
-        expect(holderNameValidationFn('   ')).toBe(null);
+        expect(holderNameValidationFn?.('', null)).toBe(null);
```
m1aw, line 8: With the TS changes bellow this is probably not needed.

packages/lib/src/components/Card/components/CardInput/validate.ts
sponglord: Aaargh! (not at you :-D) 
Gemini & Windsurf have had me jumping back and forth with conflicting opinions on how this should all look.
AI bureaucrats!
I'll try your suggestion and see if it satisfies both of them

packages/lib/src/components/Card/components/CardInput/defaultProps.ts
sponglord: It is, but I've added arguments anyway. It makes it clearer what these callbacks are supposed to receive

packages/lib/src/components/Card/components/CardInput/validate.test.ts
sponglord: I'm just gonna leave them in now. It is only a test

packages/lib/src/components/Card/components/CardInput/validate.ts
sponglord: However your change throws up a load of TS errors in the pipeline! (According to AI, anyway)

m1aw (approved):

PR #3912, opened 2026-04-15
Improve types of amazonpay component and add unit tests

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

Replace explicit and implicit any types throughout the AmazonPay component with proper type definitions, and add unit tests.

Type changes:
   •  Define AmazonWindowObject interface with typed Amazon Pay SDK methods (renderButton, initCheckout, signout, bindChangeAction) replacing generic object / Record<string, Function>
   •  Type onSubmit state parameter as SubmitData, onClick/onSignOut callbacks with proper signatures, and onError as (error: AdyenCheckoutError, component: UIElement) => void
   •  Add return types to service functions (getAmazonSignature, getCheckoutDetails, updateAmazonCheckoutSession)
   •  Add AmazonWindowObject to the global Window declaration in custom.d.ts
   •  Remove AmazonPay from the eslint ignore list

Refactoring:
   •  Replace implicit this ref pattern with useRef + setComponentRef in AmazonPayComponent
   •  Replace {...this.props} spread with explicit prop passing in AmazonPayComponent

Tests:
   •  Add new test files: AmazonPayButton.test.tsx, AmazonPayComponent.test.tsx, SignOutButton.test.tsx, ChangePaymentDetailsButton.test.tsx, OrderButton.test.tsx, defaultProps.test.ts
   •  Expand existing AmazonPay.test.tsx with tests for formatProps, formatData, submit, browserInfo, and handleDeclineFlow edge cases
   •  Increase default jest timeout to 10s in setup file

---

## 🧪 Tested scenarios

---

- COSDK-1068

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
+++ b/packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
@@ -15,7 +15,7 @@ export default function AmazonPayButton(props: Readonly<AmazonPayButtonProps>) {
     const settings = getAmazonPaySettings(props, amount);
 
     const handleOnClick = () => {
-        return new Promise(() => void props.onClick()).then(this.initCheckout).catch(error => {
+        return new Promise((resolve, reject) => void props.onClick(resolve, reject)).then(this.initCheckout).catch(error => {
             if (props.onError) props.onError(error, this.componentRef);
         });
     };
--- a/packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
+++ b/packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
@@ -54,5 +54,5 @@ export default function AmazonPayButton(props: Readonly<AmazonPayButtonProps>) {
     }, []);
 
     if (!props.showPayButton) return null;
-    return <div className="adyen-checkout__amazonpay__button" id="amazonPayButton" />;
+    return <div className="adyen-checkout__amazonpay__button" id="amazonPayButton" data-testid="amazon-pay-button" />;
 }
--- a/packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
+++ b/packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
@@ -15,15 +15,14 @@ export default function AmazonPayButton(props: Readonly<AmazonPayButtonProps>) {
     const settings = getAmazonPaySettings(props, amount);
 
     const handleOnClick = () => {
-        // eslint-disable-next-line @typescript-eslint/no-misused-promises
-        new Promise(props.onClick).then(this.initCheckout).catch(error => {
+        return new Promise(() => void props.onClick()).then(this.initCheckout).catch(error => {
             if (props.onError) props.onError(error, this.componentRef);
         });
     };
 
     const renderAmazonPayButton = (): void => {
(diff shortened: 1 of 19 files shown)


packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
```diff
@@ -15,15 +15,14 @@ export default function AmazonPayButton(props: Readonly<AmazonPayButtonProps>) {
     const settings = getAmazonPaySettings(props, amount);
 
     const handleOnClick = () => {
-        // eslint-disable-next-line @typescript-eslint/no-misused-promises
-        new Promise(props.onClick).then(this.initCheckout).catch(error => {
+        return new Promise((resolve, reject) => void props.onClick(resolve, reject)).then(this.initCheckout).catch(error => {
```
m1aw, line 18: I was going through the changes, maybe I'm missing something but I think `onClick` has been badly typed in the first place. It shouldn't have been `(*) => Promise<void>`. I don't think there's nothing in the code that expects this to be a Promise but maybe I'm wrong.

I went back in the file history and it was like this when it was implemented, but I don't see how this makes sense.

packages/lib/src/components/AmazonPay/components/SignOutButton.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+const mockAmazonRef: AmazonWindowObject = {
+    Pay: {
+        renderButton: jest.fn().mockReturnValue({ onClick: jest.fn() }),
+        initCheckout: jest.fn(),
+        signout: jest.fn(),
+        bindChangeAction: jest.fn()
+    }
+};
+
+const renderSignOutButton = (props = {}) => {
```
m1aw, line 20: I don't think I agree with these types of tests.
There's not a lot of logic here, I feel like we would be better to do an actual integration test and test this behavior in the AmazonPay Element. 

Same goes for ChangePaymentDetailsButton, OrderButton, and even to some extent AmazonPayButton.

packages/lib/src/components/AmazonPay/components/AmazonPayButton.tsx
johnayeni: For this PR, I am trying mostly to keep the code functionality exactly the same as possible since this is mostly about typing the current state of the component correctly. Since we have not made any changes to this payment method in a long time, I would leave making these kind of changes for when next the component is iterated upon.

packages/lib/src/components/AmazonPay/components/SignOutButton.test.tsx
johnayeni: I agree here that integration tests would be more effective since for this unit tests, we are mocking most of the amazon pay SDK anyway. However I needed to add this tests because of our current unit test threshold, we didn't have any unit tests for this component before 🥲 .

Even though these tests might not be the most effective, I think for now its better than having nothing anyways. If anything, it can help spot out if someone makes a code change to this component that has an unintended side effect.

ribeiroguilherme (approved):

johnayeni:
I was unable to test a full payment flow with Amazonpay because of an error with the sandbox environment, but since most of the changes in this PR is related to types, these changes should be fine.

PR #3922, opened 2026-04-17
Fix/explicit anys in secured fields

sponglord:
## 📋 Pull Request Checklist

---

## 📝 Summary

Fix/explicit `any` in `internal/SecuredFields` and its sub-directories

---

## 🧪 Tested scenarios

---

**Closes**:  

---


packages/lib/src/components/internal/SecuredFields/SFP/SecuredFieldsProvider.ts
```diff
@@ -287,7 +288,7 @@ class SecuredFieldsProvider extends Component<SFPProps, SFPState> {
     }
 
     public sfIsOptionalOrHidden(fieldType: string): boolean {
```
johnayeni, line 290:
```suggestion
    public sfIsOptionalOrHidden(fieldType: SFFieldType): boolean {
```

Would it be possible to do this instead? So we don't need to cast the type explicitly

johnayeni (approved):

packages/lib/src/components/internal/SecuredFields/SFP/SecuredFieldsProvider.ts
sponglord: I'll make a note and look into it for a future PR. Atm everything is passing - so I want to merge now :-)

PR #3924, opened 2026-04-20
Analytics - Requesting attempt ID without blocking initialization

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary
Attempt ID request was being awaited in the core initialization, potentially adding some delay for the UI to be rendered. This PR fixes it by not awaiting for it anymore. 

Changes:
- The request is done on `core.ts`  but not awaited anymore. It is fire-and-forget call.
- In the meantime, while the request is being executed, the analytics module might receive the flavor and events to be sent. The analytics module was adjusted to send this data once the attempt ID is available.

---

## 🧪 Tested scenarios
- Added tests to Core to make sure UI does not block if setUp is not finalized
- Added tests to Analytics module to make sure flavor and events are sent once attempt ID is fetched
- Manually added a delay to setup call, and checked that UI is not blocked and flavor + events are sent after ID is retrieved
- Tested that flavor is reported accordingly for dropin and components integration

### DIFF
--- a/packages/lib/src/core/Analytics/Analytics.ts
+++ b/packages/lib/src/core/Analytics/Analytics.ts
@@ -71,7 +71,6 @@ class Analytics implements IAnalytics {
 
     private set checkoutAttemptId(checkoutAttemptId: string) {
         this._checkoutAttemptId = checkoutAttemptId;
-        void this.onCheckoutAttemptIdReceived();
     }
 
     public async setUp({
@@ -103,6 +102,8 @@ class Analytics implements IAnalytics {
             };
 
             this.checkoutAttemptId = await this.service.requestCheckoutAttemptId(payload);
+
+            this.dispatchPendingAnalytics();
 
             this.storage.set({
                 id: this.checkoutAttemptId,
@@ -183,7 +184,7 @@ class Analytics implements IAnalytics {
             return;
         }
 
-        if (!this.eventsQueue.hasEventsInQueue) {
+        if (!this.eventsQueue.hasEvents) {
             return;
         }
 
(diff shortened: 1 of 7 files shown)


packages/lib/src/core/Analytics/Analytics.ts
```diff
@@ -96,10 +102,10 @@ class Analytics implements IAnalytics {
                 ...(availableCheckoutAttemptId && { checkoutAttemptId: availableCheckoutAttemptId })
             };
 
-            this.capturedCheckoutAttemptId = await this.service.requestCheckoutAttemptId(payload);
+            this.checkoutAttemptId = await this.service.requestCheckoutAttemptId(payload);
```
ScottiBR, line 104: I’d suggest removing the setter for checkoutAttemptId to avoid the hidden side effect of triggering a fire-and-forget network call during a simple assignment. Following the Principle of Least Astonishment, code is much easier to maintain and test when property assignments don't kick off background I/O; instead, making the call to onCheckoutAttemptIdResolved() explicit in your setUp logic keeps the control flow linear and predictable.

```js
const id = await this.service.requestCheckoutAttemptId(payload);
this._checkoutAttemptId = id;
this.onCheckoutAttemptIdResolved();
```

packages/lib/src/core/core.test.ts
```diff
@@ -346,7 +385,7 @@ describe('Core', () => {
             const flushPromises = () => new Promise(process.nextTick);
             await flushPromises();
 
-            const ach = dropin.dropinRef.state.elements[0];
+            const ach = (dropin.dropinRef as unknown as DropinComponent).state.elements[0];
```
ScottiBR: Double cast is not a good practice (almost an `any`), should it be `DropinComponent | null`?

packages/lib/src/core/Analytics/AnalyticsEventQueue.ts
```diff
@@ -25,6 +25,10 @@ class AnalyticsEventQueue {
         return this.logs;
     }
 
+    public get hasEventsInQueue(): boolean {
```
ScottiBR: Maybe use just `hasEvents` since queue is already in the name of the class `this.eventsQueue.hasEvents`

ScottiBR (changes requested):

ScottiBR (approved):
LGTM

PR #3957, opened 2026-05-01
GooglePay stories update

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary
- Moved stories to `GooglePay/stories` folder
- Added all stories on `GooglePay.stories.tsx` file, and created files for the individual demos
- Added demo for Express with session patching
- Created utils to reuse the code between express stories
- Adjusted stories to use countryCode as `BR` so we can test them straightforward

---

## 🧪 Tested scenarios

---

**Closes**:  

---


packages/lib/src/components/GooglePay/stories/GooglePayExpressDemo.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { useEffect, useState } from 'preact/hooks';
+import { ComponentContainer } from '../../../../storybook/components/ComponentContainer';
+import getCurrency from '../../../../storybook/utils/get-currency';
+import { makePayment } from '../../../../storybook/helpers/checkout-api-calls';
+import GooglePay from '../GooglePay';
+import { createGooglePayAmountHelper, getShippingOptions, getTransactionInfo, EXPRESS_DEMO_SETTINGS } from './googlePayExpressUtils';
+
+import type { ICore } from '../../../types';
+
+const GooglePayAmountHelper = createGooglePayAmountHelper(EXPRESS_DEMO_SETTINGS.INITIAL_AMOUNT);
```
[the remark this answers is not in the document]
ribeiroguilherme: Each page is a full page refresh/reset. There is no need for this.

packages/lib/src/components/GooglePay/stories/GooglePayExpressDemo.tsx
```diff
@@ -0,0 +1,160 @@
...
+                            if (shippingAddress?.countryCode && shippingOptionData?.id) {
+                                paymentDataRequestUpdate.newTransactionInfo = GooglePayAmountHelper.calculateNewTransactionInfo(
+                                    shippingAddress.countryCode,
+                                    shippingOptionData.id
+                                );
+                            }
+                        }
+
+                        resolve(paymentDataRequestUpdate);
+                    });
```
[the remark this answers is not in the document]
ribeiroguilherme: We are using same approach as good docs:
https://developers.google.com/pay/api/web/guides/tutorial#implement

packages/lib/src/components/GooglePay/stories/GooglePayExpressSessionsDemo.tsx
```diff
@@ -0,0 +1,221 @@
...
+                            if (shippingAddress?.countryCode && shippingOptionData?.id) {
+                                paymentDataRequestUpdate.newTransactionInfo = GooglePayAmountHelper.calculateNewTransactionInfo(
+                                    shippingAddress.countryCode,
+                                    shippingOptionData.id
+                                );
+                            }
+                        }
+
+                        resolve(paymentDataRequestUpdate);
+                    });
```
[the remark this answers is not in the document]
ribeiroguilherme: We are using same approach as good docs:
https://developers.google.com/pay/api/web/guides/tutorial#implement

packages/lib/src/components/GooglePay/stories/GooglePayExpressSessionsDemo.tsx
```diff
@@ -0,0 +1,221 @@
...
+import getCurrency from '../../../../storybook/utils/get-currency';
+import { createSession, patchCheckoutSession } from '../../../../storybook/helpers/checkout-api-calls';
+import { RETURN_URL, STORYBOOK_ENVIRONMENT_URLS } from '../../../../storybook/config/commonConfig';
+import GooglePay from '../GooglePay';
+import { AdyenCheckout } from '../../../core/AdyenCheckout';
+import { createGooglePayAmountHelper, getShippingOptions, getTransactionInfo, EXPRESS_DEMO_SETTINGS } from './googlePayExpressUtils';
+import type { PaymentAmount } from '../../../types/global-types';
+import type { CheckoutSession, CoreConfiguration } from '../../../types';
+
+const GooglePayAmountHelper = createGooglePayAmountHelper(EXPRESS_DEMO_SETTINGS.INITIAL_AMOUNT);
```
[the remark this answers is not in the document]
ribeiroguilherme: Each page is a full page refresh/reset. There is no need for this.

packages/lib/src/components/GooglePay/stories/GooglePay.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+export const Default: GooglePayStory = {
+    render: ({ componentConfiguration, ...checkoutConfig }) => (
+        <Checkout checkoutConfig={checkoutConfig}>
+            {checkout => <ComponentContainer element={new GooglePay(checkout, componentConfiguration)} />}
+        </Checkout>
+    ),
+
+    args: {
+        countryCode: 'BR',
+        componentConfiguration: {
```
m1aw, line 28: Would a suggestion be to create storybook controls for this? For example a dropdown for button color, button size mode, button type?

packages/lib/src/components/GooglePay/stories/GooglePayExpressDemo.tsx
```diff
@@ -0,0 +1,160 @@
...
+                }
+            },
+            shippingAddressRequired: true,
```
m1aw: Should we bring all of these props to the top so they don't get missed by people using this code as documentation?

packages/lib/src/components/GooglePay/stories/GooglePayExpressDemo.tsx
```diff
@@ -0,0 +1,160 @@
...
+            },
+
+            transactionInfo: getTransactionInfo(),
+
+            callbackIntents: ['SHIPPING_ADDRESS', 'SHIPPING_OPTION'],
+
+            paymentDataCallbacks: {
+                onPaymentDataChanged(intermediatePaymentData) {
+                    // eslint-disable-next-line no-async-promise-executor, @typescript-eslint/no-misused-promises
+                    return new Promise(async resolve => {
```
m1aw: Is it not possible to make `onPaymentDataChanged` async?
Then we wouldn't need:
```js
// eslint-disable-next-line no-async-promise-executor, @typescript-eslint/no-misused-promises
return new Promise(async resolve => {
```

m1aw (dismissed):

packages/lib/src/components/GooglePay/stories/GooglePay.stories.tsx
ribeiroguilherme: I see what you mean. It is a good idea, but I don't think that would work with our current _Storybook args architecture_, because the `componentConfiguration` argument holds all the Google Pay configuration properties in it as a JSON object.

That is different than the other arguments (e.g. `shopperLocale`, `countryCode`), since they do not hold a JSON structure by holding a single value assigned to it. So we can actually configure these arguments to be drop-down style for example.

If the args of the component itself could be lifted one level up (putting them outside the `componentConfiguration`), than it would work though, but that is not really how things are set today.

packages/lib/src/components/GooglePay/stories/GooglePayExpressDemo.tsx
ribeiroguilherme: Done

packages/lib/src/components/GooglePay/stories/GooglePayExpressDemo.tsx
ribeiroguilherme: Done

packages/lib/src/components/GooglePay/stories/GooglePay.stories.tsx
m1aw: I think on story book you can configure these by passing the object path as a string like `srConfig.showPanel`, but anyway this is out of scope of this change.

packages/lib/src/components/GooglePay/stories/GooglePayExpressSessionsDemo.tsx
```diff
@@ -7,37 +7,73 @@ import { RETURN_URL, STORYBOOK_ENVIRONMENT_URLS } from '../../../../storybook/co
...
-                    Country code is <code>BR</code> and currency is <code>BRL</code>.
-                </li>
-                <li>Initial amount is 100.00; shipping costs are added on top.</li>
-                <li>Ships only to Brazil and the United States.</li>
-                <li>Shipping options differ based on the selected country.</li>
-            </ul>
-        </div>
-    );
-}
+const GOOGLE_PAY_EXPRESS_PROPS: GooglePayConfiguration = {
```
m1aw: Oh, this wasn't what I meant. 
I was talking about having a split between props and callbacks so we can immediately see all the props that get configured on top of the object and the callbacks at the bottom of the object.

```
// properties
isExpress: true,

transactionInfo: getTransactionInfo(),
callbackIntents: ['SHIPPING_ADDRESS', 'SHIPPING_OPTION'],

shippingAddressRequired: true,
shippingAddressParameters: {
    phoneNumberRequired: true
},
shippingOptionRequired: true,

// callbacks
onAuthorized: (data, actions) => {
    console.log('Authorized data', data);
    actions.resolve();
},

onSubmit: () => (),

paymentDataCallbacks: {
        async onPaymentDataChanged(intermediatePaymentData) {
...
```

packages/lib/src/components/GooglePay/stories/GooglePay.stories.tsx
ribeiroguilherme: Feel free to share an example if you have one, I can do that. 

I think the `srConfig.showPanel` seems like an object but in reality is just a string that gets mapped to a boolean value.  I didn't find in the code it being treated as object.

packages/lib/src/components/GooglePay/stories/GooglePayExpressSessionsDemo.tsx
ribeiroguilherme: Ah, ok. I updated the code

packages/lib/src/components/GooglePay/stories/GooglePay.stories.tsx
m1aw: Indeed, it's working however for `srConfig.showPanel` but it doesn't for anything in `componentConfiguration`. Actually I don't fully see what the difference is but should not matter.

m1aw (approved):

PR #3980, opened 2026-05-07
Fix: Select add additional aria attributes

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

**Problem:** The non-filterable `<select>` button was missing ARIA attributes required for the listbox disclosure pattern, making it non-compliant for screen readers. Additionally, the button's accessible name did not include the selected value.
 
**Changes:**
- Added `aria-haspopup="listbox"` — announces to screen readers that the button opens a listbox (not a menu or dialog)
- Added `aria-controls={selectListId}` — programmatically links the button to the listbox it controls
- Fixed `aria-expanded` — was already present but `showList` was not being passed down; now correctly reflects open/closed state
- Added `aria-labelledby` combining label id + selected value span id — ensures the button's accessible name includes both the field label and the currently selected option (e.g., "Identifier, Mobile, button")
- Added `id` to the label element in Field component to enable the `aria-labelledby` reference
 
**Note:** The combobox (filterable) path already had all required ARIA attributes — no changes needed there.

---

## 🧪 Tested scenarios

1. Button announcement on focus

- Tab to the select button
- Expected: SR announces "[selected value or placeholder], collapsed, button" (or similar per SR)
- Verifies: aria-haspopup="listbox" and aria-expanded="false" are read correctly

2. Open/close state

- Click or press Space/Enter to open
- Expected: SR announces "expanded" on open, "collapsed" on close
- Verifies: aria-expanded toggling

3. Listbox association

- Open the dropdown
- Expected: SR moves focus into the listbox and announces it as a listbox (not an orphan element)
- Verifies: aria-controls linking button → listbox

4. Keyboard-only navigation

- Tab to button → Space to open → Arrow keys to navigate → Enter to select → Escape to close without selecting
- Expected: no focus traps, list closes and focus returns to button on Escape/Tab
- Verifies: full keyboard flow

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
+++ b/packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
@@ -12,7 +12,19 @@ function SelectButtonElement({ filterable, toggleButtonRef, ...props }) {
         return <div {...strippedProps} ref={toggleButtonRef} />;
     }
 
-    return <button id={props.id} aria-expanded={props.showList} aria-disabled={props.readonly} aria-describedby={props.ariaDescribedBy} type={'button'} {...props} ref={toggleButtonRef} />;
+    return (
+        <button
+            id={props.id}
+            aria-haspopup="listbox"
+            aria-expanded={props.showList}
+            aria-controls={props.selectListId}
+            aria-disabled={props.readonly}
+            aria-describedby={props.ariaDescribedBy}
+            type={'button'}
+            {...props}
+            ref={toggleButtonRef}
+        />
+    );
 }
 
 function SelectButton(props: Readonly<SelectButtonProps>) {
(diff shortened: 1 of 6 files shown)


packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
```diff
@@ -65,11 +77,14 @@ function SelectButton(props: Readonly<SelectButtonProps>) {
             onKeyDown={!readonly ? props.onButtonKeyDown : null}
             toggleButtonRef={props.toggleButtonRef}
             id={props.id}
+            showList={showList}
+            selectListId={props.selectListId}
```
ribeiroguilherme, line 81: We are leaking non-standard attributes into the `<button>` created by the `SelectButtonElement`. 
You can add the new props to the other destructured values, so that would not happen. Ex:

```js
function SelectButtonElement({ filterable, toggleButtonRef, showList, selectListId, ...props }) {
```

ribeiroguilherme (dismissed):

packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
m1aw: Nvm, that didn't even make sense. Anyway, it's fixed. Can you recheck?

ribeiroguilherme (approved):

packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
ribeiroguilherme: Done

PR #3991, opened 2026-05-15
Improved: New implementation of useFormWithA11y with fields aria-live for CtP and PayTo

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

Introduces useFormWithA11y, a wrapper around useForm that adds two accessibility mechanisms without requiring the SRPanel infrastructure. 

There's 2 main changes with this:

1. For Submit-time focus management, we use the useFormWithA11y. It works after triggerValidation() is called (e.g. clicking Continue/Submit), focus automatically moves to the first invalid field in schema order. This satisfies the WCAG 4.1.3 / E47 / E102 requirement that keyboard and screen reader users are directed to errors.

2. Inline `aria-live` announcements, a new opt-in `errorLive` prop on Field adds aria-live="polite" to the error span. When an error string appears or changes (blur type errors), screen readers announce it without the user needing to navigate there. This is off by default so existing components are unaffected.

**`useFormWithA11y.ts` is also a replacement to useForm and could be used on other components besides just PayTo, and should be the default going forward.**

---

### ClickToPay — submit with empty / invalid email

 1. Open a page that renders the ClickToPay login form
 2. Without entering anything, press Enter or click Continue
 3. Expected: focus moves to the email input; VoiceOver reads the field label and aria-invalid state
 4. Type a valid email and press Enter
 5. Simulate a "not enrolled" API response
 6. Expected: focus returns to the email input; the API error message is announced via aria-live ("Email not found" or equivalent)
 7. Tab away and Shift+Tab back
 8. Expected: error is re-read via aria-describedby

### ClickToPay — blur-time validation (no unwanted focus steal)

1. Click into the email field, then Tab out without typing
2. Expected: the field shows an error state but focus does not move — it continues to the next focusable element naturally

### PayTo (PayID) — submit with multiple invalid fields

 1. Open the PayTo PayID form
 2. Switch identifier to Email and leave the email field empty; press Pay / submit
 3. Expected: focus moves to the email field; error is announced
 4. Tab through remaining fields
 5. Expected: each error span is read as focus reaches it (via aria-describedby)
 6. Repeat with ABN and Org ID identifiers

---

**Closes**:  

---


packages/lib/src/components/internal/FormFields/Field/Field.tsx
```diff
@@ -115,6 +116,7 @@ const Field: FunctionalComponent<Readonly<FieldProps>> = props => {
                 className={classNames({ 'adyen-checkout-contextual-text--error': true, 'adyen-checkout-contextual-text--hidden': !showError })}
                 {...(contextVisibleToSR && { id: `${uniqueId.current}${ARIA_ERROR_SUFFIX}` })}
                 aria-hidden={contextVisibleToSR ? undefined : 'true'}
+                {...(contextVisibleToSR && errorLive && { 'aria-live': 'polite' })}
```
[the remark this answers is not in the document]
m1aw: This is not true, we don't use `display: none` here so it works.

packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLoginInput.tsx
```diff
@@ -22,31 +23,41 @@ interface CtPLoginInputDataState {
...
     const [isLoginInputDirty, setIsLoginInputDirty] = useState<boolean>(false);
 
     const validateInput = useCallback(() => {
         setIsLoginInputDirty(true);
         triggerValidation();
     }, [triggerValidation]);
 
+    const focusInput = useCallback(() => {
+        if (containerRef.current) {
+            setFocusOnField(containerRef.current, 'shopperLogin');
```
dimitra-adyen, line 48: I was testing this solution and I noticed that after entering an invalid email (NOT_FOUND Error), the focus doesn't move back to the input field, it rather stays on the continue button. I believe it is because we are trying to set the focus on the input while it is still disabled.

packages/lib/src/utils/useForm/useErrorFocus.ts
```diff
@@ -0,0 +1,30 @@
...
+    return holder as Element | string;
+}
+
+/**
+ * Hook that provides a utility to focus the first invalid field in a form.
+ * Uses the existing setFocusOnField utility which handles dropdowns, iOS scroll quirks, etc.
+ *
+ * @param holder - DOM Element, CSS selector string, or a RefObject pointing to the form root element
+ */
+export function useErrorFocus(holder: Element | string | RefObject<Element>) {
```
dimitra-adyen, line 16: 👌

packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLoginInput.tsx
m1aw: Thank you for catching this, I missed this while testing, I think I paid more attention to PayTo.

Also the change was not made to the OtP part of ClickToPay so I have also made that change. Also is cleaner overall, let me know what you think of the last commit.

packages/lib/src/utils/useForm/useFormWithA11y.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+beforeEach(() => {
+    mockSetFocusOnField.mockClear();
+});
+
+// ---------------------------------------------------------------------------
+// useErrorFocus
+// ---------------------------------------------------------------------------
+
+describe('useErrorFocus', () => {
```
ScottiBR: Let's move these into a dedicated useErrorFocus.test.ts file right next to the hook. Also, could we add coverage for the CSS string and RefObject branches? We especially need to test that unmounted null guard so focus doesn't get silently swallowed.

ScottiBR (commented):
Just a quick comment

dimitra-adyen (dismissed):

packages/lib/src/utils/useForm/useFormWithA11y.test.tsx
m1aw: @ScottiBR can you review?

ScottiBR (dismissed):
LGTM (you still need to fix some typescript regression before merge)

sponglord (approved):

PR #4002, opened 2026-05-27
 Fixed: Await component now announces its status text to screen readers

ScottiBR:
## 📋 Pull Request Checklist

---

## 📝 Summary

After form submission, the PayTo await screen renders the "Awaiting your approval" banner silently — NVDA Speech Viewer shows nothing for that region. This violates **WCAG 4.1.3 (Status Messages)** and **EN 301 549 clause 9.4.1.3**.

This PR adds `useA11yReporter(props.awaitText)` to the shared `Await` component so the await text is announced once through the global `SRPanel` (`role="log"`, `aria-live="polite"`) when the screen mounts. The visible banner remains as plain text — no second live region is added, preventing double announcements.

**Changes:**

- `Await.tsx` — +1 import, +1 `useA11yReporter` hook call (before early returns, per hooks rules)
- `Await.test.tsx` — Added `Accessibility` describe block verifying `srPanel.setMessages` is called with the await text on mount; extended `renderAwait` helper to accept an optional `srPanel` for spying

This fix is payment-agnostic and benefits all payment methods using the `Await` component (PayTo, MBWay, etc.), following the same pattern already used in `QRFinalState`.

---

## 🧪 Tested scenarios

1. **Unit test (new):** Verified `srPanel.setMessages` is called with the `awaitText` value on component mount.
2. **All existing Await tests pass** (14/14).
3. **Manual test with PayToAwaitScreen story:** Confirmed NVDA Speech Viewer announces "Awaiting your approval" once on mount; banner is not focusable; no double announcements.
4. **TypeScript strict scan:** Zero new errors.
5. **Lint + Style lint:** Clean.

---

**Closes**:

### DIFF
--- a/packages/lib/src/components/internal/Await/Await.test.tsx
+++ b/packages/lib/src/components/internal/Await/Await.test.tsx
@@ -13,12 +13,14 @@ jest.mock('../../../core/Services/payment-status');
 
 const renderAwait = ({
     awaitProps,
-    amountProviderProps
+    amountProviderProps,
+    srPanel: externalSrPanel
 }: {
     awaitProps: AwaitComponentProps;
     amountProviderProps?: Partial<AmountProviderProps>;
+    srPanel?: SRPanel;
 }) => {
-    const srPanel = new SRPanel(global.core);
+    const srPanel = externalSrPanel ?? new SRPanel(global.core);
 
     return render(
         <CoreProvider i18n={global.i18n} loadingContext="test" resources={global.resources}>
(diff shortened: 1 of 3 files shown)


packages/lib/src/components/internal/Await/Await.test.tsx
```diff
@@ -13,12 +13,14 @@ jest.mock('../../../core/Services/payment-status');
...
-    amountProviderProps
+    amountProviderProps,
+    srPanel: externalSrPanel
 }: {
     awaitProps: AwaitComponentProps;
     amountProviderProps?: Partial<AmountProviderProps>;
+    srPanel?: SRPanel;
 }) => {
-    const srPanel = new SRPanel(global.core);
+    const srPanel = externalSrPanel ?? new SRPanel(global.core);
```
johnayeni: Small request but can we already take the opportunity to replace `global.core` here with `core` from `setupCoreMock` ?

ScottiBR: Nice catch!

johnayeni (approved):

PR #4004, opened 2026-05-27
Fix: use semantic list markup for brand icon components

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

Brand icon lists were rendered as `<div>` containers with `<span>` items, which is not semantically correct and causes accessibility issues (implicit list role announced by screen readers without proper list structure).

### Changes

- **`BrandIcons`**: Changed container element from `<div>` to `<ul>` and each brand item from `<span>` to `<li>`, giving the list proper semantic structure. Added `list-style: none; padding: 0` to reset default `<ul>` styles. Added `ariaHidden` prop to allow callers to suppress SR announcement when the brand list is decorative.
- **`BrandImage`**: Added polymorphic `as` prop (defaults to `'span'`) so the wrapper element can be swapped to `<li>` when rendered inside a list context.
- **`AvailableBrands`** (Card): Passes `ariaHidden` to `BrandIcons` when the list is decorative; updated tests to reflect the new list markup.
- **`PaymentMethodIcon`**: Renamed `altDescription` prop to `alt` for consistency; updated all call sites (`PaymentMethodItem`, `UPIIntentAppItem`, `PaymentMethodBrands`, `OrderPaymentMethods`).

### Testing

Existing `AvailableBrands` tests updated to assert against `<ul>`/`<li>` elements. No breaking changes to the public merchant-facing API.

---

## 🧪 Tested scenarios

---

- COSDK-1186

**Closes**:  

---

### DIFF
--- a/.changeset/four-rivers-shave.md
+++ b/.changeset/four-rivers-shave.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed: Brand icon lists in Card and Drop-in now use correct list semantics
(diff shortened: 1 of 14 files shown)


m1aw (approved):
Looks good, it might be worth adding a description to the list.

johnayeni:
> Looks good, it might be worth adding a description to the list.

This would mostly likely be picked up as a follow up

PR #4005, opened 2026-05-27
PoC - Google Accelerated Checkout 

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

**Google Accelerated Checkout** support has been added to the Adyen Web Drop-in. This introduces a generic **showDropinHeaderWhenSelected** mechanism that lets a payment method render its own full UI in place of the standard Drop-in item header, and adds logic to promote Google Pay to the top of the Stored payment methods section whenever the shopper is eligible for accelerated checkout (independent of whether the experiment is enabled). When both the shopper is eligible and the experiment is enabled, Google Pay renders its accelerated checkout UI (an inline iframe) instead of the standard button, with automatic fallback to the standard button on load failure. 

###  1. Generic mechanism to hide Drop-in item header when selected
A payment method can now render its own full UI when selected, hiding the standard Drop-in PaymentMethodItem header/background.

- **UIElement** — new overridable getter **showDropinHeaderWhenSelected** (defaults to true). 
- **PaymentMethodItem.tsx** — computes hideHeader = !paymentMethod.showDropinHeaderWhenSelected && isSelected; conditionally omits the entire header block and adds the --headerless modifier class.
- **PaymentMethodItem.scss** — styling for --selected.--headerless (primary background + tighter details padding).
- **GooglePay.tsx** — overrides the getter to return false only in ACCELERATED_CHECKOUT mode.

### 2. Promote GooglePay into the Stored Payment Methods section
- **Dropin/elements/promoteGooglePay.ts** (new, explicitly marked temporary/experiment-only) — finds a promotable -GooglePay element (from regular or instant payment lists), removes it from its list, prepends it as the first stored payment element, and flags it oneClick so it opens at the top.
- Promotion depends only on shopper eligibility (isShopperEligibleForAcceleratedCheckout), not on whether the experiment is enabled.
- **DropinComponent.tsx** — wires promoteGooglePayIfNeeded into the Promise.all(...).then(...) chain when elements resolve.

### 3. Accelerated Checkout UI (eligibility + experiment gated)
- **GooglePay.tsx** — isAvailable() rewritten to run eligibility check + isReadyToPay in parallel (Promise.allSettled). Sets isShopperEligibleForAcceleratedCheckout and only switches mode to ACCELERATED_CHECKOUT when eligible AND acceleratedCheckoutExperiment === 'enabled'. Renamed internal service field googlePay → googleButtonClient.
- **GooglePayComponent.tsx (new)** — chooses between GoogleAcceleratedCheckout and the standard GooglePayButton; falls back to the button if the accelerated iframe fails to load.
- **GoogleAcceleratedCheckout.tsx (new)** — renders the iframe container, loads the Google client, and handles dynamic resize.
- **GoogleAcceleratedCheckoutClient.ts (new)** — wraps google.payments.api.AcceleratedCheckoutClient, lazy-loads the accelerated script, exposes isAvailable()/load()/resize subscription.
- **PaymentDataRequest.ts (new)** — class building the Google Pay payment-data request from config.

---

## 🧪 Tested scenarios

This feature cannot be tested locally — it must be verified on the deployed Netlify branch. The Accelerated Checkout UI only appears when the experiment is enabled AND the shopper is eligible; GooglePay promotion to the stored section happens whenever the shopper is eligible (regardless of experiment state).


packages/lib/src/components/GooglePay/GooglePay.test.ts
```diff
@@ -473,38 +489,121 @@ describe('GooglePay', () => {
...
+            expect(gpay.mode).toBe(GooglePaymentMode.STANDARD_BUTTON);
+            expect(gpay.isShopperEligibleForAcceleratedCheckout).toBe(true);
+            expect(core.modules.analytics.sendAnalytics).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    type: InfoEventType.eligibilityPassed,
+                    component: 'googlepay_accelerated_checkout_experiment'
+                })
+            );
+        });
+
```
ScottiBR, line 568: Add another fallback test to ensure the header is visible when you fallback, even when load failed:

Image

```suggestion

        test('should restore the Drop-in header when accelerated checkout load falls back to the standard button', async () => {
            const core = setupCoreMock();
            const gpay = new GooglePay(core, {
                configuration: { merchantId: 'merchant-id', gatewayMerchantId: 'gateway-id', acceleratedCheckoutExperiment: 'enabled' }
            });
            const acceleratedCheckoutClient = getAcceleratedCheckoutClientInstance();
            const buttonClient = getButtonClientInstance();
            acceleratedCheckoutClient.isAvailable.mockResolvedValue({ status: 'SUCCESS' });
            acceleratedCheckoutClient.load.mockResolvedValue({ status: 'ERROR' });
            acceleratedCheckoutClient.onPaymentSheetResize.mockReturnValue(() => {});
            buttonClient.createButton.mockResolvedValue(document.createElement('button'));

            await gpay.isAvailable();
            expect(gpay.showDropinHeaderWhenSelected).toBe(false);

            render(gpay.render());
            await act(() => new Promise(process.nextTick));

            expect(screen.getByTestId('googlepay-button-container')).toBeInTheDocument();
            expect(gpay.showDropinHeaderWhenSelected).toBe(true);
        });
```

packages/lib/src/components/GooglePay/components/GooglePayComponent.tsx
```diff
@@ -0,0 +1,50 @@
...
+const GooglePayComponent = ({
+    defaultMode,
+    googleAcceleratedCheckoutClient,
+    googleButtonClient,
+    showPayButton,
+    googleButtonProps
+}: Readonly<Props>) => {
+    const [activeMode, setPaymentMode] = useState<GooglePaymentMode>(defaultMode);
+
+    const onAcceleratedCheckoutLoadFailed = useCallback(() => {
```
ScottiBR, line 37: This works, however when load fails with `ERROR` status the `GooglePay.mode` remains `ACCELERATED_CHECKOUT`, so Drop-in keeps the payment-method header hidden even though the standard button is now rendered.

Image

ScottiBR (changes requested):

packages/lib/src/components/GooglePay/GooglePay.test.ts
ribeiroguilherme: Added test case

packages/lib/src/components/GooglePay/components/GooglePayComponent.tsx
ribeiroguilherme: Header should be visible now once the `load()` fails

ScottiBR (dismissed):
LGTM!

packages/lib/src/components/GooglePay/utils/map-adyen-brands-to-googlepay-brands.ts
```diff
@@ -11,7 +11,13 @@ const brandMapping: Record<string, google.payments.api.CardNetwork> = Object.fre
     maestro: 'MAESTRO'
 });
 
-export function mapGooglePayBrands(brands: string[]): google.payments.api.CardNetwork[] {
-    const mappedBrands = brands.map(brand => brandMapping[brand]).filter((brand): brand is google.payments.api.CardNetwork => !!brand);
+export function mapGooglePayBrands(brands: string[], countryCode: string): google.payments.api.CardNetwork[] {
```
ScottiBR: What if we keep the original function, just adding a new condition to the filter?
```js
const mappedBrands = brands
  .map(brand => brandMapping[brand])
  .filter((brand): brand is google.payments.api.CardNetwork => 
    !!brand && !(brand === google.payments.api.CardNetwork.MAESTRO && countryCode !== 'BR')
  );
```

ScottiBR (changes requested):

packages/lib/src/components/GooglePay/utils/map-adyen-brands-to-googlepay-brands.ts
ribeiroguilherme: It works, but I didn't want to mix with the original function since it must be removed later on.  Keeping it separate will be easier to spot what has to be removed.

ScottiBR (dismissed):

ScottiBR (dismissed):

ScottiBR (dismissed):

PR #4006, opened 2026-05-28
Add new message for click to pay invalid email error

dimitra-adyen:
## 📋 Pull Request Checklist

---

#### Issue
When a user entered an invalid email on the Click to Pay login screen, INVALID_PARAMETER error was returned. This was surfaced to the user with a generic fallback message: "Something went wrong, try again or use the manual card entry", which is not helpful to them.

#### Changes
- CtPLogin.tsx:  In the login error handler, INVALID_PARAMETER errors are now remapped to the new INVALID_EMAIL error code, so the UI can show a specific, contextual message instead of the generic fallback. Since currently we are only supporting logging in through email, this is the only parameter entered by the user that could be invalid.
- en-US.json — Added the new translation key ctp.errors.INVALID_EMAIL with the value:
"Enter a valid email address (e.g. name@example.com)"which  gives users a concrete suggestion on the expected format.

---

## 🧪 Tested scenarios

---

**Closes**:  
COSDK-1185
---

### DIFF
--- a/packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLogin.tsx
+++ b/packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLogin.tsx
@@ -55,8 +55,9 @@ const CtPLogin = (): h.JSX.Element => {
         } catch (error: unknown) {
             if (error instanceof SrciError) console.warn(`CtP - Login error: ${error.toString()}`);
             if (error instanceof TimeoutError) console.warn(error.toString());
-            if (isSrciError(error)) setErrorCode(error?.reason);
-            else console.error(error);
+            if (isSrciError(error)) {
+                setErrorCode(error?.reason == 'INVALID_PARAMETER' ? 'INVALID_EMAIL' : error?.reason);
+            } else console.error(error);
 
             setIsLoggingIn(false);
         }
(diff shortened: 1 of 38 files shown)


packages/lib/src/components/internal/ClickToPay/components/CtPLogin/CtPLogin.tsx
```diff
@@ -55,8 +55,9 @@ const CtPLogin = (): h.JSX.Element => {
         } catch (error: unknown) {
             if (error instanceof SrciError) console.warn(`CtP - Login error: ${error.toString()}`);
             if (error instanceof TimeoutError) console.warn(error.toString());
-            if (isSrciError(error)) setErrorCode(error?.reason);
-            else console.error(error);
+            if (isSrciError(error)) {
```
m1aw, line 58: Can you also make sure we get the same message when we don't type anything?
Right now it shows an empty error message. I believe the change is to add `ctp.errors.INVALID_EMAIL` message to validation schema on Login Input `useForm`.

dimitra-adyen: Updated the useForm validation schema for ctp login with the translation key, and updated the implementation to display it.

m1aw (dismissed):

m1aw (approved):

PR #4023, opened 2026-06-09
Fixed: PayTo accessibility enhancements for form instruction and account holder errors

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

- **Type fixes for Select component**: Added \`SelectButtonElementProps\` interface extending \`HTMLAttributes<HTMLElement>\`, typed previously untyped refs (\`filterInputRef\`, \`toggleButtonRef\`, \`selectListRef\`), and replaced deprecated \`h.JSX\` namespace imports with direct preact imports (\`HTMLAttributes\`, \`TargetedKeyboardEvent\`).
- **Fix aria-describedby association in PayTo BSBInput**: The Fieldset's \`aria-describedby\` was on the fieldset root, not reaching inner input fields. Moved the association to pass \`descriptionId\` down to the inner \`Field\` components so screen readers correctly associate the instruction text with the input.
 
---
 
## 🧪 Tested scenarios
 
- Manually verified PayTo BSB input fields have correct \`aria-describedby\` association with the instruction text.
- Verified Select component renders and functions correctly with typed props.
- TypeScript compilation and ESLint pass clean.
 
---
 
**Closes**: Closes: COSDK-1184


packages/lib/src/components/PayTo/components/BSBInput.tsx
```diff
@@ -74,6 +77,7 @@ export default function BSBInput({ setComponentRef, defaultData, placeholders, o
                     onInput={handleChangeFor('bankAccountNumber', 'input')}
                     onBlur={handleChangeFor('bankAccountNumber', 'blur')}
                     placeholder={placeholders?.bankAccountNumber}
+                    aria-describedby={instructionId}
                     required={true}
                 />
             </Field>
```
dimitra-adyen: While testing, I noticed that for BSBInput the instruction gets announced twice when focusing on the first element. Both the `fieldset` element (through Fieldset's aria-describedby) and the `input` element (through InputBase's aria-describedby) reference the same instructionId, so the same text is announced twice on focus. Do we need `aria-describedby` inside the `fieldset` element since the child input already handles the announcement correctly?

m1aw: Good catch. My guess what's happening here that fieldset by default "labels" the first input, this doesn't happen on PayID because it has a selector first.

dimitra-adyen: For PayID I think it is not happening, because ariaDescribedBy is not getting passed down to the `SelectButtonElement`. When testing PayID, the instruction gets announced only if you tab into the select element, not if you click on it. Once I added it, it gets announced twice as well. What I tried locally, was keep `aria-describedby` on the `fieldset`, only if there is a description and no externalDescriptionId.

dimitra-adyen (dismissed):

dimitra-adyen (approved):

PR #4027, opened 2026-06-10
fix: add autocomplete tokens to personal details and address fields

m1aw:
## 📋 Pull Request Checklist

- [x] If adding new analytics events: N/A

---

## 📝 Summary

Adds WCAG 2.2 compliant `autocomplete` tokens to all form fields that collect personal data, to support browser autofill and meet accessibility requirements (LevelAccess task CW-77).

**Changes:**
- New `AutocompleteToken` / `AutocompleteValue` types in `FormFields/types.ts`
- `InputBase` accepts an optional `autocomplete` prop (renders `autocomplete=undefined` when `null`)
- `PersonalDetails`: `given-name`, `family-name`, `email`, `tel`, `bday`
- `Address / FieldContainer`: `billing` or `shipping` prefixed tokens derived from the `label` prop (`billingAddress` / `deliveryAddress`)
- Same tokens applied to `PayToNameFields`, `PayIDInput`, `BacsInput`, `BankTransferInput`, `DragonpayInput`, `SendCopyToEmail`, `BrazilPersonalDetail`

---

## 🧪 Tested scenarios

- Unit tests added to `OpenInvoice.test.tsx` covering personal details, billing address, and delivery address autocomplete tokens
- Unit tests added to `FieldContainer.test.tsx` for all text fields × billing/delivery address types
- Existing `PersonalDetails.test.tsx` autocomplete test already on branch

---

**Closes**: COSDK-1167


packages/lib/src/components/Ach/components/AchComponent.tsx
```diff
@@ -162,6 +163,7 @@ function AchComponent({
                         onBlur={handleChangeFor('routingNumber', 'blur')}
                         required={true}
                         maxlength={9}
+                        autocomplete={undefined}
```
sponglord, line 166: Should this not be `null` or `"off"`? (As per the accompanying WCAG reference doc: `a11y-autocomplete.md`)
Alternatively can you add to the doc explaining when the `undefined` use case is appropriate?

packages/lib/src/components/internal/CompanyDetails/CompanyDetails.tsx
```diff
@@ -80,6 +80,7 @@ export default function CompanyDetails(props: Readonly<CompanyDetailsProps>) {
                         onInput={inputEventHandler('input')}
                         onBlur={eventHandler('blur')}
                         spellcheck={false}
+                        autocomplete={'organization'}
```
sponglord, line 83: The `a11y-autocomplete.md` doesn't mention this - what does this cause to be auto-completed?

packages/lib/src/components/internal/FormFields/types.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * WCAG 2.2 Input Purposes autocomplete tokens.
+ * @see https://www.w3.org/TR/WCAG22/#input-purposes
+ */
+export type AutocompleteToken =
```
sponglord, line 5: Is this the list of all possible values from w3.org?
Do they all come into play in our sdk? If not should we just list the types we actually use?

packages/lib/src/components/Ach/components/AchComponent.tsx
m1aw: Let me update the doc. Indeed I designed to use `null` instead of `undefined ` but I changed because in preact `undefined` is the default "hide this property" value.
"off" is not needed in this case as it's better suited to let the browser handle it.

Undefined is nice since we force developers to think about the autocomplete value.

packages/lib/src/components/internal/CompanyDetails/CompanyDetails.tsx
m1aw: It's just another value available in: https://www.w3.org/TR/WCAG22/#input-purposes

packages/lib/src/components/internal/FormFields/types.ts
m1aw: They currently don't all come into play in the SDK, but since these are internal types I thought it' easier just to have them.

sponglord (dismissed):

sponglord (dismissed):

sponglord (approved):

PR #4029, opened 2026-06-11
fix: improve Select keyboard navigation and ARIA compliance

dimitra-adyen:
## 📋 Pull Request Checklist

---

## 📝 Summary

Ensure keyboard navigation in the Select component in non filterable mode

- Add role="combobox" to the non-filterable <button>, so SRs understand that the button controls a listbox and pass the arrow keys directly to the component rather than use them for virtual cursor navigation. 
- Add aria-activedescendant conditionally, when the dropdown is open, so screen readers don't announce a stale active option when the list is collapsed.
- Fix a bug in keyboard navigation (setNextActive / setPreviousActive) where items were compared by object reference instead of the `id`, causing navigation to silently break when the item reference wasn't identical.
---

## 🧪 Tested scenarios

PayID: Tab into the identifier Select -> Enter the list -> Navigate using the keyboard arrow keys
ClickToPay: Tab into the Cards Select -> Enter the list -> Navigate using the keyboard arrow keys
Ach: Tab into the Account Type Select -> Enter the list -> Navigate using the keyboard arrow keys
Card with Installments: Enter the card details -> Tab into the Installments Select -> Enter the list -> Navigate using the keyboard arrow keys

- Attached screen recordings to the ticket
---

**Closes**:  
COSDK-744
---

### DIFF
--- a/.changeset/stale-icons-talk.md
+++ b/.changeset/stale-icons-talk.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed: a11y keyboard navigation inside Select list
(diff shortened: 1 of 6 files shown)


m1aw (approved):
Tested on VoiceOver and NVDA

Issue #3315, opened 2025-05-09
Setup `GooglePay` component with 3DS


jureperak:
**Describe the bug**
I'm trying to implement Adyen web package using Advanced flow for GooglePay component. The issue I'm having is implementing 3DS correctly. 
Not sure how I should trigger in the onSubmit method to close the GooglePay sheet and mount the result of the /payment (action object) into 3DS element. 

In `onSubmit` method of `GooglePay`, after i call `/payment` i receive `IdentifyShopper` with `action` object. Then I'm passing `action` object into instance of `AdyenCheckout` into `createFromAction` and mounting 3DS, like this: `adyenCheckoutInstance.createFromAction(action).mount(threeDSecureRef.current);`

Immediately after that I'm calling `actions.resolve({ resultCode, action });`

But for some reason, i get two `/submitThreeDS2Fingerprint` API calls (where one fails and second returns something meaningful, `action` object with `type: threeDS2` and `subtype: "challenge"`)

**To Reproduce**
I prepared a simple code that can serve as a reference so we are on the same page. 
1. Go to [stackblitz](https://stackblitz.com/edit/vitejs-vite-dw73guk5)

**Expected behavior**
After clicking on GooglePay, selecting test card with 3DS (Adyens test card 3ds_visa), clicking Continue, should close GooglePay sheet and show 3DS in my container.

jureperak:
@sponglord any chance you can help me out 🙂

jureperak:
> Alternatively, if you want to use createFromAction to mount the 3DS2 component in a different container - then you can call actions.resolve passing it an empty object. This will prevent the SDK from trying to handle the action itself & creating it's own 3DS2 component

This part almost work, the only issue I'm having is that when approving payment in GooglePay sheet and mounting 3DS2 into my container, i got event `OnPaymentCompleted` called. But that should not happen as the 3DS2 was not resolved.  https://stackblitz.com/edit/vitejs-vite-dw73guk5

So the question is, when using GooglePay component, do i need to set `onSubmit` and `onAdditionalDetails` events from AdyenCheckout or GooglePay instance?  @sponglord

ribeiroguilherme:
@jureperak  about the code sample you shared: you don't need to call `createFromAction` within the `onSubmit`. You can pass the `resultCode` and `action` to the `actions.resolve()` and the Component will render the 3DS component for you.

[...]

PR #4044, opened 2026-06-26
Feature/enable review

dimitra-adyen:
## 📋 Pull Request Checklist

---

#### Overview
Introduces an `onReview` callback and a `processPayment` method that together enable merchants to present a review page to shoppers before the final payment is submitted. This feature targets the /sessions flow (processPayment is sessions only and will return an error if called in the advanced flow).

#### New Public API
`onReview(state: PaymentData, component: UIElement,  reviewDetails: ReviewDetails)`
A new optional callback in CoreConfiguration. When configured, it is called instead of immediately submitting the payment when the shopper clicks the pay button. `reviewDetails` contains additional data relevant to the review page. It currently holds the 'orderStatus', when a partial payment order is in progress. The merchant receives the payment data and is responsible for rendering their own review page, then calling checkout.processPayment(state) to proceed.

Note: Payment methods that manage their own native submission flow, like PayPal, Klarna, Apple Pay, Google Pay, AmazonPay, ANCV and PayByBankPix, bypass this callback to avoid interrupting their native experiences.

`checkout.processPayment(data: PaymentData) (Sessions flow only)`
A new method on core for use from the review page. Calls session.submitPayment, then handles the full response lifecycle:
- onPaymentCompleted / onPaymentFailed 
- onAction when an action (redirect, 3DS2, QR code) is required
- core.update + onOrderUpdated when there's remaining amount

`onAction(actionElement: UIElement)`
A new callback in CoreConfiguration that fires when an action (redirect, 3DS2, QR code, etc.) needs to be presented to the shopper. The merchant is responsible for mounting the received actionElement to their page. After calling processPayment, if the response contains an action, onAction is invoked with the component to render.

#### Internal Changes

- When `onReview` is set and the component is valid, submit():
  - Fires a LogEventType.review analytics event
  - Calls onReview(this.data, this.elementRef) and returns — no payment call is made
  
- `UIElement.executePaymentsCall()`
  - Extracted from submit() so payment methods that bypass the review flow (e.g. PayByBankPix's payWithStoredPayment) can call the payment API directly without going through the onReview gate.

- PayButton "Continue" label + icon: When onReview is active, showReview is passed as an explicit prop to each `<PayButton>` instance. The PayButton then:
  - Renders "Continue" instead of any custom button label
  - Hides the payment method icon

- **Klarna widget bypass**: When `useKlarnaWidget: true`, `KlarnaPayments.submit()` skips `onReview` and calls `executePaymentsCall()` directly to avoid interrupting Klarna's native SDK flow. `PayButton` receives `showReview={false}` explicitly so the label shows the normal pay amount instead of "Continue".

- **`PayButton.showReview` prop**: A new optional `showReview` prop on PayButton allows the rendering component to override the context value. This is used by `UIElement.payButton` (and all class overrides) to pass `showReview={!!this.props.onReview}`, ensuring the button label reflects the *component-level* `onReview` rather than the outer Dropin context. Without this, the Dropin's `CoreProvider` singleton would cause all nested components to inherit `showReview: true` even when a specific component has had `onReview` null.

#### Supporting Changes
- `PaymentData` type extended with additional optional fields: socialSecurityNumber, installments, shopperEmail, shopperName, telephoneNumber, dateOfBirth, bankAccount, beneficiaryId
- `DonationCampaignOptions.commercialTxAmount` made optional: falls back to checkout.options.amount.value, supporting the review page scenario where Donation is instantiated outside the Dropin context.
---

## 🧪 Tested scenarios

---

**Closes**:  

COSDK-1144
Respective ADRs:
- ADR-2482
- ADR-2623
- ADR-2652
---


packages/lib/src/components/Card/Card.tsx
```diff
@@ -336,7 +336,14 @@ export class CardElement extends UIElement<CardConfiguration> {
     protected override payButton = (props: PayButtonProps): h.JSX.Element => {
         const isZeroAuth = this.props.amount?.value === 0;
         const isStoredCard = this.props.storedPaymentMethodId?.length > 0;
-        return <PayButton {...props} label={isZeroAuth && !isStoredCard ? this.props.i18n.get('payButton.saveDetails') : ''} onClick={this.submit} />;
+        return (
+            <PayButton
+                {...props}
+                label={isZeroAuth && !isStoredCard ? this.props.i18n.get('payButton.saveDetails') : ''}
+                onClick={this.submit}
+                showReview={!!this.props.onReview}
```
johnayeni, line 361:
```suggestion
                showReview={Boolean(this.props.onReview)}
```

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -606,7 +616,7 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
      * Get the payButton component for the current element
      */
     protected payButton = (props: PayButtonProps) => {
-        return <PayButton {...props} onClick={this.submit} />;
+        return <PayButton {...props} onClick={this.submit} showReview={!!this.props.onReview} />;
```
johnayeni, line 634:
```suggestion
        return <PayButton {...props} onClick={this.submit} showReview={Boolean(this.props.onReview)} />;
```

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -662,7 +672,13 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
 
     render() {
         return (
-            <CoreProvider i18n={this.props.i18n} loadingContext={this.props.loadingContext} resources={this.resources} analytics={this.analytics}>
+            <CoreProvider
+                i18n={this.props.i18n}
+                loadingContext={this.props.loadingContext}
+                resources={this.resources}
+                analytics={this.analytics}
+                showReview={!!this.props.onReview}
```
johnayeni:
```suggestion
                showReview={Boolean(this.props.onReview)}
```

packages/lib/src/components/Card/Card.test.tsx
```diff
@@ -9,8 +9,10 @@ import { CardFocusData } from '../internal/SecuredFields/lib/types';
 import { mock } from 'jest-mock-extended';
 import { AmountProvider } from '../../core/Context/AmountProvider';
 import { ICore } from '../../types';
+import { _MockProxy } from 'jest-mock-extended/lib/Mock';
 
 describe('Card', () => {
+    const core = global.core as _MockProxy<ICore> & ICore;
```
johnayeni:
```suggestion
    const core =  setupCoreMock();
```

We can just use setup core mock here

packages/lib/src/components/Card/Card.test.tsx
```diff
@@ -57,19 +59,46 @@ describe('Card', () => {
...
                 expect(await screen.findByRole('button', { name: 'Confirm preauthorization' })).toBeTruthy();
             });
         });
+
+        describe('onReview', () => {
+            const amount = { value: 1000, currency: 'USD' };
+            const customRender = (ui: h.JSX.Element, contextShowReview = false) =>
+                render(
+                    // @ts-ignore ignore
+                    <CoreProvider i18n={global.i18n} loadingContext="test" resources={global.resources} showReview={contextShowReview}>
```
johnayeni:
```suggestion
                    <CoreProvider i18n={core.modules.i18n} loadingContext="test" resources={core.modules.resources} showReview={contextShowReview}>
```

packages/lib/src/components/Card/Card.test.tsx
```diff
@@ -57,19 +59,46 @@ describe('Card', () => {
...
                 customRender(card.payButton());
                 expect(await screen.findByRole('button', { name: 'Confirm preauthorization' })).toBeTruthy();
             });
         });
+
+        describe('onReview', () => {
+            const amount = { value: 1000, currency: 'USD' };
+            const customRender = (ui: h.JSX.Element, contextShowReview = false) =>
+                render(
+                    // @ts-ignore ignore
```
johnayeni: Why was this ts ignore needed?

packages/lib/src/components/Donation/DonationCampaignService.tsx
```diff
@@ -31,7 +31,7 @@ class DonationCampaignService {
         this.onDonationCompleted = checkout.options.donation?.onDonationSuccess;
         this.onDonationFailed = checkout.options.donation?.onDonationFailure;
 
-        this.commercialTxAmount = donationCampaignProps.commercialTxAmount;
+        this.commercialTxAmount = donationCampaignProps.commercialTxAmount ?? checkout.options.amount?.value ?? 0;
```
johnayeni, line 34: What actually ends up being the behaviour of the donation component if the amount is zero? Can the shopper still make a donation action?

packages/lib/src/components/internal/PayButton/PayButton.test.tsx
```diff
@@ -13,14 +13,16 @@ import { mock } from 'jest-mock-extended';
 const renderPayButton = ({
     payButtonProps = {},
     amountProviderProps = {},
-    i18n = global.i18n
+    i18n = global.i18n,
```
johnayeni:
```suggestion
    i18n = core.modules.i18n,
```

We can take the opportunity to use `setupCoreMock` in this file now and replace these global modules

packages/lib/src/components/internal/PayButton/utils.ts
```diff
@@ -43,8 +43,13 @@ function createButtonLabel(
     amount: PaymentAmount,
     isZeroAuth: boolean,
     customAmount?: PaymentAmount,
-    secondaryAmount?: PaymentAmount
+    secondaryAmount?: PaymentAmount,
+    showReview?: boolean
```
johnayeni:
```suggestion
function createButtonLabel(i18n, { customLabel: string,
amount: PaymentAmount,
isZeroAuth: boolean,
customAmount?: PaymentAmount,
secondaryAmount?: PaymentAmount,
showReview?: boolean
}
```

I think we should refactor this createButtonLabel parameters to accept an object so that it is clear what the values are for in the places were the the function is being called. Since it has multiple boolean values now, it would become harder to read.

packages/lib/src/components/internal/UIElement/UIElement.test.tsx
```diff
@@ -580,6 +588,75 @@ describe('UIElement', () => {
...
+            element.submit();
+            expect(onReview).not.toHaveBeenCalled();
+        });
+    });
+
+    describe('payButton()', () => {
+        const amount = { value: 1000, currency: 'USD' };
+        const renderPayButton = (ui: h.JSX.Element, contextShowReview = false) =>
+            render(
+                <CoreProvider i18n={global.i18n} loadingContext="test" resources={global.resources} showReview={contextShowReview}>
```
johnayeni:
```suggestion
                <CoreProvider i18n={core.modules.i18n} loadingContext="test" resources={core.modules.resources} showReview={contextShowReview}>
```

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -257,6 +257,16 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
             return;
         }
 
+        if (this.props.onReview) {
+            this.submitAnalytics(new AnalyticsLogEvent({ component: this.type, type: LogEventType.review, message: 'Review page displayed' }));
```
johnayeni: Just thinking about this event message here 'Review page displayed', technically we do not know if the merchant ended up displaying the review page but that rather the review flow was triggered. The page being displayed is completely up to the merchant but our SDK only sees that the merchants wants the data to display this page.

Should this message reflect that then to be more accurate?

packages/lib/src/components/Klarna/KlarnaPayments.tsx
```diff
@@ -38,8 +38,16 @@ class KlarnaPayments extends UIElement<KlarnaConfiguration> {
...
+        if (this.props.useKlarnaWidget) {
+            this.executePaymentsCall();
+            return;
+        }
+        super.submit();
+    }
+
     public payButton = (props: PayButtonProps) => {
-        return <PayButton onClick={this.submit} {...props} />;
+        return <PayButton onClick={this.submit} {...props} showReview={this.props.useKlarnaWidget ? false : !!this.props.onReview} />;
```
johnayeni, line 50:
```suggestion
        return <PayButton onClick={this.submit} {...props} showReview={this.props.useKlarnaWidget ? false : Boolean(this.props.onReview)} />;
```

packages/lib/src/components/PayByBankPix/PayByBankPix.test.ts
```diff
@@ -324,6 +324,32 @@ describe('PayByBankPix', () => {
...
+                payByBankPixDetails: {
+                    receiver: 'mock-receiver',
+                    ispb: 'mock-issuer',
+                    deviceId: 'mock-device'
+                },
+                amount: { value: 100, currency: 'BRL' }
+            });
+
+            render(element.render());
+            await user.click(screen.getByRole('button', { name: /Continue/i }));
```
johnayeni: Should we check for the whole string here instead of just `continue`?

packages/lib/src/components/Twint/Twint.tsx
```diff
@@ -31,7 +31,7 @@ class TwintElement extends RedirectElement {
      * @param props - props
      */
     protected override payButton = (props: PayButtonProps): h.JSX.Element => {
-        return <PayButton {...props} label={this.payButtonLabel()} onClick={this.submit} />;
+        return <PayButton {...props} label={this.payButtonLabel()} onClick={this.submit} showReview={!!this.props.onReview} />;
```
johnayeni, line 34:
```suggestion
        return <PayButton {...props} label={this.payButtonLabel()} onClick={this.submit} showReview={Boolean(this.props.onReview)} />;
```
You can choose to ignore all the comments that are similar to this by the way but it just a personal preference here as I think this is more readable.

packages/lib/src/core/Context/CoreProvider.tsx
```diff
@@ -44,7 +46,8 @@ const CoreProvider = ({ i18n, loadingContext, resources, analytics, children }:
                 i18n,
                 loadingContext,
                 resources,
-                analytics
+                analytics,
+                showReview
```
johnayeni: Why do we also need `showReview` to be part of the CoreProvider? 

We are just checking the `onReview` prop in most places and it seems this is only used for the PayButton.

sponglord:
This flow is just for /sessions right? 
For the Advanced flow merchants still proceed as they always did with [these steps](https://docs.adyen.com/online-payments/two-step-checkout)?

If so, I think the description should be clearer to emphasise this (/sessions only)

packages/lib/src/components/Twint/Twint.tsx
dimitra-adyen: I see we are using both in the codebase, so I would prefer that we agree on one. We can discuss it with the team and agree on either `Boolean` or `!!`, so I make the change in all instances for consistency.

packages/lib/src/components/internal/UIElement/UIElement.tsx
dimitra-adyen: Good one. Replaced it with `Review flow triggered`

packages/lib/src/components/internal/UIElement/UIElement.test.tsx
dimitra-adyen: fixed

packages/lib/src/components/internal/PayButton/utils.ts
dimitra-adyen: updated

packages/lib/src/components/internal/PayButton/PayButton.test.tsx
dimitra-adyen: fixed

packages/lib/src/components/Card/Card.test.tsx
dimitra-adyen: fixed

packages/lib/src/components/Card/Card.test.tsx
dimitra-adyen: Good catch, missed this while removing some checks inside the test. Fixed

packages/lib/src/components/Card/Card.test.tsx
dimitra-adyen: Fixed

packages/lib/src/components/Donation/DonationCampaignService.tsx
dimitra-adyen: The behavior stays the same. If the amount is 0, we throw an error that an amount is not set.

packages/lib/src/components/PayByBankPix/PayByBankPix.test.ts
dimitra-adyen: fixed

packages/lib/src/core/Context/CoreProvider.tsx
dimitra-adyen: We are indeed using it only inside the `PayButton`. I passed the `onReview` prop from the remaining components/payment methods, for which the review flow can be enabled, to the `PayButton` and removed it from here. Thanks!

johnayeni (dismissed):
There are still a few more `!!this.props.onReview` but asides that, mostly good from my end!

packages/lib/src/core/core.ts
```diff
@@ -368,6 +368,51 @@ class Core implements ICore {
...
+                            new AdyenCheckoutError('IMPLEMENTATION_ERROR', 'onAction callback is required to handle payment actions')
+                        );
+                    }
+                    return;
+                }
+                const order = response.order;
+                if (order && (order.remainingAmount?.value ?? 0) > 0) {
+                    return this.update({ order }).then(() => {
+                        this.options.onOrderUpdated?.({ order });
+                    });
```
ribeiroguilherme: Do we need to care about partial payments here? How does it work within a review page context? (e.g. there is no Drop-in in that page)

packages/lib/src/core/config.ts
```diff
@@ -51,6 +51,7 @@ export const GENERIC_OPTIONS = [
     'beforeRedirect',
     'beforeSubmit',
     'onSubmit',
```
ribeiroguilherme, line 53: Isn't `onAction` supposed to be here?

packages/lib/src/components/Dropin/stories/dropin/ReviewPage.stories.tsx
```diff
@@ -0,0 +1,405 @@
...
+
+export const KlarnaWidget: DropinStory = {
+    args: { countryCode: 'NL', useSessions: true },
+    render: renderKlarnaStory(true)
+};
+
+export const KlarnaRedirect_InIsolationMode: DropinStory = {
+    args: { countryCode: 'NL', useSessions: true },
+    render: renderKlarnaStory(false)
+};
```
ribeiroguilherme: The Storybook is mainly to showcase the flows we have. It is like a demo tool, that can be used by us or external stakeholders like product, team leads, etc.

It feels like some of these new stories are mainly to perform a 'manual test' and see that the feature works as expected, although they do not add much value to be part of the Storybook demos IMO.

Besides that, these stories (e.g. klarna widget, klarna redirect in isolation mode, dropin with cards bypassing review) already have tests covering these use-cases in the codebase so should be fine to not add them as part of our show-case examples.

That is just my opinion - if you wanna leave like that is also fine. I just want to avoid us adding/covering all sort of edge-cases for components/flows to Storybook, otherwise it will become too complex and with repetitive stories that will be hard to maintain if we move towards that direction.

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -257,6 +257,16 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
...
+        if (this.props.onReview) {
+            this.submitAnalytics(new AnalyticsLogEvent({ component: this.type, type: LogEventType.review, message: 'Review flow triggered' }));
+            this.props.onReview(this.data, this.elementRef);
+            return;
+        }
+
+        this.executePaymentsCall();
+    }
+
+    public executePaymentsCall(): void {
```
ribeiroguilherme, line 280: One of the ideas of the `onAction` callback was that it would also enable us to let merchants render any action somewhere else in the UI (as requested #3315 for example ) . Would that still be possible with this feature? If so, is that part of the scope of the changes?

packages/lib/src/core/types.ts
```diff
@@ -284,6 +293,13 @@ export interface CoreConfiguration {
      */
     afterAdditionalDetails?(component: UIElement): void;
 
+    /**
+     * Callback called when an action (for example a QR code or 3D Secure 2 authentication screen) is shown to the shopper.
```
ribeiroguilherme: This comment is not accurate. This callback is called when an action have to be created; not when it is displayed to the shopper.

packages/lib/src/core/core.ts
```diff
@@ -368,6 +368,51 @@ class Core implements ICore {
...
+                }
+                cleanupFinalResult(response);
+                this.options.onPaymentCompleted?.(response);
```
ribeiroguilherme, line 407: The logic of handling the response is kinda of duplicated here but also in the `UIElement.handleResponse()` method.

Perhaps (and if it makes sense) we could try to use the  _strategy pattern_ and have a shared logic that could be reused in both places if that makes sense. Then we would have a single place which would hold the logic for handling the response. (There is also a tech debt ticket that mentions this)

packages/lib/src/components/Dropin/stories/dropin/ReviewPage.stories.tsx
```diff
@@ -0,0 +1,405 @@
...
+            amount={amount}
+            shopperLocale={shopperLocale}
+            onReview={(data, sessionId) => setReviewState({ data, sessionId })}
+        />
+    );
+};
+
+export const Dropin: DropinStory = {
+    args: { countryCode: 'NL', useSessions: true },
+    render: renderDropin
```
ribeiroguilherme: The Donation keeps getting rendered when payment is authorized. Shall we disable it for this flow?

packages/lib/src/core/core.ts
```diff
@@ -368,6 +368,51 @@ class Core implements ICore {
...
+    public processPayment(data: PaymentData): void {
+        if (!this.session) {
+            this.options.onError?.(new AdyenCheckoutError('IMPLEMENTATION_ERROR', 'processPayment requires a session to be configured'));
+            return;
+        }
+
+        const event = new AnalyticsLogEvent({
+            type: LogEventType.submit,
+            message: 'Shopper clicked pay',
+            component: LogEventType.review
```
ribeiroguilherme: Are we supposed to send 'review' as a component value? Was this validated by the data team?
Obviously there is no component in the screen. I am just wondering if this is the right value to have it here.

dimitra-adyen: I validated the events with the Data team, but not specifically the component for this case. Since `component` is required and as you mentioned there's no component in that screen I followed what we do for  `actions` that's why I passed `review` as `component`. For actions we're passing as component the action itself (e.g. threeDS2, qrCode, voucher, redirect, etc). If you think it should be different I can check specifically for this with Data.

packages/lib/src/components/Dropin/stories/dropin/ReviewPage.stories.tsx
dimitra-adyen: removed this variation completely

packages/lib/src/core/types.ts
dimitra-adyen: accidentally copied from `onActionHandled`. fixed!

packages/lib/src/components/internal/UIElement/UIElement.tsx
dimitra-adyen: Very good point! I updated `handleResponse` to check `onAction` and allow merchants render the action wherever they want. If I understand the issue you tagged correctly, this should solve it. The only part that we should highlight, is that once the merchants pass `onAction` they are responsible for unmounting the payment component. Take a look and let me know

packages/lib/src/components/Dropin/stories/dropin/ReviewPage.stories.tsx
dimitra-adyen: Fair point! I had kept them indeed for everyone to verify the edge cases, but I totally agree with your point on maintainability. I removed the `Review Page` stories, and kept two stories to showcase the solution. One under dropin and one under the cards component.

packages/lib/src/core/config.ts
dimitra-adyen: fixed

packages/lib/src/core/core.ts
dimitra-adyen: I discussed this with Product, and I shared my idea about checking `onReview` only in the scenario that the Giftcard balance covers the amount, and in any other case we skip it and `onReview` gets checked only after another payment method has been selected to cover the remaining amount.
It was decided to not implement it and keep it as is, meaning `onReview` gets checked when a Giftcard is entered, triggering `onOrderUpdated` when there's a remaining amount, needing merchants to create a new Checkout instance there with the existing session and order.
I removed `this.update` which was redundant since as you mentioned there's not dropin in the Review page.
If you have any doubts about it, i'm happy to discuss it again.

m1aw: This is an interesting decision, indeed I would vouch for your suggestion, I would feel better on stating that Partial Payments are not supported on this flow, because that's indeed what happens.

packages/lib/src/core/core.ts
dimitra-adyen: Created tech debt ticket for it (COSDK-1385)

packages/lib/src/core/core.ts
dimitra-adyen: Updated the events as below, after aligning with the data team: 
```category: log
type: "review"
component: <txVariant, e.g. "scheme">
message:  "Review flow triggered"
```
```
category: log
type:     "submit"
subType:  "review"
component: <data.paymentMethod.type>
message:  "Shopper clicked pay"
```

packages/lib/src/core/core.ts
dimitra-adyen: Updated the solution for partial payments, so that if the total amount can't be covered by the giftcard, we are making an `/orderStatus` call and pass the response to `onReview`, so merchants can use that info for display purposes in their Review page. If the giftcard can cover the total amount, there's no order, so no such call is made and the only available information is through the paymentMethod.

packages/lib/docs/adr/ADR-0003-review-page-flow.md
```diff
@@ -0,0 +1,123 @@
...
+
+Chosen option: **"Option 3 — Configuration property + explicit callback"**
+
+**Justification:** It is the only option that scores positively on platform alignment, developer experience, support cost reduction, and future-proofing simultaneously. It works with both the built-in and custom pay buttons, supplies extra information (e.g. card last four, up-to-date session details) through a dedicated callback, and requires no hidden-button or pre-fabricated-state juggling. Implementation effort is the sole driver where it does not lead, and the other benefits clearly outweigh it.
+
+Option 1 was rejected as unintuitive, treating Advanced and Sessions differently, and allowing SDK initialisation with pre-fabricated state. Option 2 was rejected because it cannot use the built-in pay button and still requires exposing component state on web with no clear mechanism for additional information.
+
+### Positive Consequences
+
+- Advanced and Sessions flows are aligned
```
ribeiroguilherme: Are they aligned, though? 😅

packages/lib/docs/adr/ADR-0003-review-page-flow.md
```diff
@@ -0,0 +1,123 @@
...
+- Merchant is notified only when component state is valid for payment
+- SDK owns the button label change ("Pay" → "Continue") and can emit a reliable `review` analytics event
+- Additional data (card last four, session details) is easy to surface
+
+### Negative Consequences
+
+- New public callback (`onReview`) and new configuration surface
+- On web, breaks the existing documented Advanced flow review page pattern
```
ribeiroguilherme: We will keep the `documented Advanced flow` it as it is, no?

packages/lib/docs/adr/ADR-0003-review-page-flow.md
```diff
@@ -0,0 +1,123 @@
...
+
+When `onReview` is configured:
+
+- The shopper clicks Continue with a valid state, `submit()` calls `onReview(state, component, reviewDetails)` and **returns immediately — no payment call is made**. The merchant renders its review page and must call `checkout.processPayment(state)` to resume.
+- When a partial payment order is in progress, the SDK fetches the order status first and passes it as `reviewDetails.orderStatus`. If that call fails, `onReview` is still invoked with an empty `reviewDetails` object so the flow is never blocked.
+- The SDK automatically relabels the Pay button to **"Continue"** and hides the payment method icon; no merchant configuration needed.
+- A `review` analytics event is fired, enabling tracking of review page usage.
+
+```typescript
+onReview?(state: PaymentData, component: UIElement): void;
```
ribeiroguilherme: It is missing the `reviewDetails` parameter

packages/lib/docs/adr/ADR-0003-review-page-flow.md
```diff
@@ -0,0 +1,123 @@
...
+
+- In the Advanced flow the SDK never owns the `/payments` call, so there is nothing for `processPayment` to invoke. It would have to call back into `onSubmit`.
+
+Advanced flow merchants are therefore unaffected and keep the existing pattern: perform the `/payments` call from the review page, and mount any returned action with `checkout.createFromAction(action)`. `onReview` still fires, so the review page itself is identical in both flows.
+
+## Advice and Concerns
+
+- **Unsupported payment methods.** The following bypass `onReview` to avoid interrupting their native flows: **Apple Pay, Google Pay, AmazonPay, PayPal, Klarna (widget), ANCV, PayByBankPix**.
+    - Apple Pay / Google Pay / AmazonPay: the payment must complete while their overlay is open, so data cannot be carried to a review page.
+    - PayPal: possible in **Advanced flow only**, not Sessions ([explanation](https://hub.is.adyen.com/engineering/platform/payments/checkout/web/a-developers-guide-to/paypal#qa)).
```
ribeiroguilherme: We do not reference internal docs in the public docs

ribeiroguilherme (dismissed):

ribeiroguilherme (approved):

PR #4066, opened 2026-07-07
Replace deprecated keypress event with keydown

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

The [`keypress`](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) event is deprecated, so this swaps it out for [`keydown` ](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) everywhere. Same behaviour, just no longer using a dead API. Also renamed the related handlers/props/types to match, and did a bit of cleanup along the way.

## What changed

- **Inputs** (`InputBase`): `onKeyPress`/`handleKeyPress` → `onKeyDown`/`handleKeyDown`
- - **Inputs** (`InputBase`): Remove `onKeyUp` handler as it is not being used and `onKeyDown` now covers the usecase for `onKeyUp`
- **Button**: dropped the redundant `onKeyPress` prop, kept `onKeyDown`
- **UIElement / Dropin**: `handleKeyPress` → `handleKeyDown` (Enter-to-submit still works the same)
- **SecuredFields**: renamed `onKeyPressed`/`handleKeyPressed`/`SFKeyPressObj` → `onKeyDown`/`handleKeyDown`/`SFKeyDownObj`
- Tidied up unused props and marked a few `SecuredFieldsProvider` fields `readonly`
- Added a `minor` changeset

## Testing

- [x] Unit tests pass
- [x] Enter still submits in Card / Custom Card / Giftcard
- [x] Enter on a Drop-in payment-method header doesn't submit

---

## 🧪 Tested scenarios

---

- COSDK-520

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/internal/FormFields/InputBase.tsx
+++ b/packages/lib/src/components/internal/FormFields/InputBase.tsx
@@ -65,23 +65,20 @@ export default function InputBase({ setRef, ...props }: Readonly<InputBaseProps>
 
     /**
      *  Event is fired when a key that produces a character value is pressed down.
-     *  ENTER keypress also triggers this event.
-     *
-     *  TODO: 'keypress' event is deprecated
-     *  https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event
+     *  ENTER keydown also triggers this event.
      */
-    const handleKeyPress = useCallback(
+    const handleKeyDown = useCallback(
         (event: TargetedKeyboardEvent<HTMLInputElement>) => {
-            if (props?.onKeyPress) props.onKeyPress(event);
+            if (props?.onKeyDown) props.onKeyDown(event);
         },
-        [props?.onKeyPress]
+        [props?.onKeyDown]
     );
 
     /**
      * Event is fired when certain keys are pressed (keys that do not output characters):
      * Backspace, Arrow keys, Shift, Ctrl, Command, Option, Esc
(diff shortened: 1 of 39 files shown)


packages/lib/src/components/internal/FormFields/InputBase.tsx
```diff
@@ -63,31 +63,11 @@ export default function InputBase({ setRef, ...props }: Readonly<InputBaseProps>
...
-        [props?.onKeyPress]
-    );
-
-    /**
-     * Event is fired when certain keys are pressed (keys that do not output characters):
-     * Backspace, Arrow keys, Shift, Ctrl, Command, Option, Esc
-     *
-     * Exception: ENTER keypress triggers 'onKeyPress' AND 'onKeyUp'
-     */
-    const handleKeyUp = useCallback(
```
m1aw, line 86: This probably should be in the PR description to highlight this was intentional. Just to confirm this is removed because it's not used anywhere? Looking at the commits just made this more confusing.

m1aw (approved):

packages/lib/src/components/internal/FormFields/InputBase.tsx
johnayeni: I have updated the PR description for this 👍🏿

PR #4077, opened 2026-07-10
Detect multiple instances of AdyenCheckout and add suppressible warning

sponglord:
## 📋 Pull Request Checklist

---

## 📝 Summary

When integrating adyen-web, merchants can make the mistake of initialising AdyenCheckout multiple times. 

In 99% of cases this is not necessary (see below for known use cases requiring more than one instance of AdyenCheckout). Merchant use cases such as:
- separating "branded card" components e.g. one for Visa, one for MC
- having the card PM and then having the Dropin for non-card PMs
- separating instant payment methods from the rest

can all be handled against one instance of Checkout.

There are known issues caused by initialising AdyenCheckout multiple times.  One such example is: 
- **One Session ID can be linked to multiple `checkoutAttemptId`'s**
    _Real world case_: In a `/sessions` integration, a merchant initialised four standalone Components, and for each Component one AdyenCheckout was created. For each Checkout instance a request for a `checkoutAttemptId` was made. The request includes a reference to the `sessionId` which led to four `checkoutAttemptId`'s being linked to one Session ID.

**Why this is bad?**
Apart from not being "best practice", and being indicative of a poor integration health - this will also affect a merchant's conversion rates. Of the four `checkoutAttemptId`'s only one will be associated with the actual payment. This will give an impression that the merchant's conversion rate is only 25% (4 instances of Checkout only led to 1 payment), when in fact it is 100% (the shopper only had 1 checkout experience in the UI, which led to a successful payment).

---
Multiple instance of checkout also lead to initialising and loading the `RiskModule` multiple times (with it's subsequent https calls), and loading securedFields an excessive amount of times. All of which represent poor integration health and may trigger alerts in the backend as we see an excess of calls to the risk and securedField urls

---
This PR tracks how many times AdyenCheckout is instantiated and puts a warning in the console if this more than once.
This will alert merchants, at dev time, that they are doing something wrong

However, because there are known use cases requiring more than one instance of AdyenCheckout ,this console warning can be suppressed by a new Checkout level config property, which at the moment, is called `suppressConfigWarnings`. The message in the console warning refers to this property:
> Multiple instances of AdyenCheckout detected on the same page. This is not recommended and may lead to unexpected behaviour. If this is intentional you can suppress this warning by setting "suppressConfigWarnings: true" in your Checkout config.

I've kept the name, `suppressConfigWarnings`, generic because I want to discuss if this may be something we might want to use to cover other, dev-time, warnings. For example, we've discussed in the past about generating warnings when we see that some Checkout config properties have been incorrectly, or sub-optimally, set.

---

**Known use cases requiring more than one instance of AdyenCheckout**
There are more details of the following in the referenced ticket, COSDK-1032, but a high level overview of these use case is:
- Multi-Transaction Checkouts: such as, Marketplace Baskets with Split Orders, or Baskets with Concurrent, Separate Charges (like a subscription + a one-time gift card purchase)
- Multi-Tenant or Multi-Entity Routing on a Single Page: products from different merchants handled at the same time, merchants that are perhaps legally, or geographically, separate
- Micro-Frontend Architecture and Component Isolation: Express Checkout may be separated from Standard Checkout when different teams are responsible for each set of components 
- Cross-Border / Multi-Currency Split Orders: If a shopper is checking out with a complex cart where certain items are fulfilled from a local entity (e.g. in EUR) and others from an international entity (USD)

Also, the flow for the forthcoming review page for /sessions will require a new instance of AdyenCheckout. 
And there have been cases in the past, for example with Plugins, where a new instance has been used to handle an action.
In both cases though this activity takes place on a new page, where the reference to the original instance of AdyenCheckout has been lost. These scenarios will not be affected by this warning (unless they spin up multiple instance of AdyenCheckout on the new page!).

---

## 🧪 Tested scenarios

- Multiple instances of Checkout trigger the console warning
- This warning can be suppressed by adding `suppressConfigWarnings: true` to the top level Checkout config
- Validated this approach with Product
---

COSDK-1032

---


packages/lib/src/core/core.ts
```diff
@@ -51,6 +52,10 @@ class Core implements ICore {
         Core.metadata.bundleType = type;
     }
 
+    public static increaseInitialisationCount(): void {
+        Core.metadata.numberOfInitialisedCheckouts += 1;
+    }
```
ScottiBR, line 57: Do we need to worry about the case where a merchant creates a checkout, tears it down, then creates another one, like a SPA re-initing on navigation? Since the counter only ever increments and there's no decrement on teardown, i think we'd warn 'multiple instances on the same page' even though only one is alive at a time. Maybe we decrement on remove/unmount?

ScottiBR (commented):
Should we add a few unit tests for this? Like `should warn when a second instance is created on the same page`, `should not warn for the first instance`, and `should suppress the warning when suppressConfigWarnings is true`, plus one that `suppressConfigWarnings` isn't flagged as an invalid config prop.

packages/lib/src/core/core.ts
```diff
@@ -100,6 +105,15 @@ class Core implements ICore {
             );
         }
 
+        Core.increaseInitialisationCount();
+        if (Core.metadata.numberOfInitialisedCheckouts > 1) {
+            if (!this.options.suppressConfigWarnings) {
```
ScottiBR: small thing, feel free to ignore, we can flatten these two ifs into one

sponglord:
> Should we add a few unit tests for this? Like `should warn when a second instance is created on the same page`, `should not warn for the first instance`, and `should suppress the warning when suppressConfigWarnings is true`, plus one that `suppressConfigWarnings` isn't flagged as an invalid config prop.

Yeah! I wrote these tests first time round (when I tackled this issue in March) but forgot to migrate them to this new PR

packages/lib/src/core/core.ts
sponglord: thinking...

sponglord: I don't know quite how we would do this.
We don't have an official tear down function for AdyenCheckout (perhaps we need one?)
Doing it on a PM/UIElement unmount won't work - since that can be triggered several times e.g. in the 3DS2 flow the `Card` gets unmounted, and both the `3DS2Fingerprint` & `Challenge` get unmounted. I've tested, and this leads the `Core.metadata.numberOfInitialisedCheckouts` to equal `-2`.
Thoughts?

ScottiBR: yeah, fair point. i was thinking of the merchant unmounting/removing the component or page and dropping the checkout reference, rather than us having a real destroy hook for the SDK. 

My main concern is just that the counter is really tracking “checkouts created since page load”, rather than “checkout instances currently alive”. But I do agree that we can't safely decrement this from a UIElement unmount. 

Maybe that's an acceptable limitation for now and we think this better in v7?

PR #4080, opened 2026-07-10
Fixed: Remove explicit any types from Dropin, ANCV, Giftcard and ThreeDS2 components

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

Part of the ongoing task to eliminate `@typescript-eslint/no-explicit-any` (COSDK-1070). This PR addresses 18 issues across four component folders:

- **Giftcard** (6 issues), of note:
  - Moved `CardBrandsConfiguration` to shared folder types file 
  - New `onOrderUpdatedCallbackType`
- **ThreeDS2** (5 issues), of note:
  - added explicit return type to `promiseTimeout`
- **ANCV** (4 issues), of note:
  -  replaced the invalid `ref: Ref<typeof ANCVInput>` typing (which pointed at the component function itself, leaving `setStatus`/`showValidation` silently broken since `this` doesn't work in a function component) with the `setComponentRef`/`ComponentMethodsRef` pattern already used by Ach, Econtext, etc.; typed `paymentData` as `string`, and callback props with new `onOrderRequestCallbackType` / `onOrderUpdatedCallbackType`.
  -  Added test coverage for beneficiaryId data propagation, validation, and the submit flow
- **Dropin** (3 issues), of note:
  - typed `onCreateElements` return as a 4-tuple of `Promise<UIElement[]>` which is strange but....

---

## 🧪 Tested scenarios

Mostly type-only changes; no runtime behaviour altered — with one exception in ANCV, where the `setComponentRef` fix restores `setStatus`/`showValidation`, which were previously silently broken.
- Removed the four folder ignores from `eslint.config.js` and ran `yarn lint` — no remaining explicit-any violations in the covered folders.
- Added unit tests in `ANCV.test.tsx` covering beneficiaryId propagation to state/`onChange`, validation error display, and the `onSubmit` payment flow.

---

**Closes**: COSDK-1070

### DIFF
--- a/packages/lib/src/components/ANCV/components/ANCVInput.tsx
+++ b/packages/lib/src/components/ANCV/components/ANCVInput.tsx
@@ -1,5 +1,5 @@
-import { h, Ref } from 'preact';
-import { useEffect, useState } from 'preact/hooks';
+import { h } from 'preact';
+import { useEffect, useRef, useState } from 'preact/hooks';
 import { useCoreContext } from '../../../core/Context/CoreProvider';
 import LoadingWrapper from '../../internal/LoadingWrapper';
 import InputText from '../../internal/FormFields/InputText';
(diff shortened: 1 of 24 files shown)


packages/lib/src/components/ANCV/components/ANCVInput.tsx
```diff
@@ -10,7 +10,7 @@ import { ANCVDataState } from '../types';
 import { UIElementProps } from '../../internal/UIElement/types';
 
 export interface ANCVInputProps extends UIElementProps {
-    ref?: any;
+    ref?: Ref<typeof ANCVInput>;
```
johnayeni: Can we refactor this and use `setComponentRef` instead?

m1aw: That's what I'm working on atm, also cleaning up the `this` from the onChange.

packages/lib/src/components/Dropin/types.ts
```diff
@@ -26,6 +26,7 @@ type PaymentActionTypesMap = {
  * Type must be loose, otherwise it will take priority over the rest
  */
 type NonMappedPaymentMethodsMap = {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- we allow "unknown" payment methods defaulting to redirect
     [key: string]: any;
```
johnayeni, line 30:
```suggestion
    [key: string]: string;
```

Even those these payment methods are unknown, should it not then still be a string type?

m1aw: Yeah, but they are not string are they? It's the PM object.

johnayeni (approved):

PR #4081, opened 2026-07-10
feat: emit `paymentListDisplayed` analytic event from drop-in when payment methods are displayed

FabioRomeiro:
## 📋 Pull Request Checklist

---

## 📝 Summary

- **Motivation:** It is crucial for our data quality around the Drop-in component to be aware of exactly what **is** shown to the shopper, what **is not** being shown, and most importantly, **how** it is being rendered. The question we want to answer is _"How were the payment methods exactly displayed on the shopper's Drop-in in a specific checkout session?"_
- **Problem:** Our current analytical events, like the `rendered` event, don't tell the full story. Currently the front-end can not only remove a payment method from the drop-in, but it can also affect the original order coming from `/paymentMethods` response, this breaks visibility on the data analysis.
- **Solution:** A new Drop-in level event triggered when all components are ready to be used by the user. This new event then sends 2 new properties:
    - `paymentMethods`: A list of payment methods ordered by how it was displayed to the user (currently the layout order is always `fastlane` -> `instant` -> `stored` -> `regular`).
    - `unavailablePaymentMethods`: If a payment method came in the `/paymentMethods` response and was not displayed for ANY reason, it will be placed here. Here are some cases in which we add the payment method to this list:
        - `isAvailable` returned false
        - `showPaymentMethods` is set to false
        - `showStoredPaymentMethods` is set to false
        - `filterStoredPaymentMethods` filtered it out
        - The payment method is part of `UNSUPPORTED_PAYMENT_METHODS`

---

## 🧪 Tested scenarios

- PII fields coming from store payment methods are not passed to the event
- New `paymentListDisplayed` event ignore properties injected via `filterStoredPaymentMethods`
- Event sends payment methods in display order, not API response order
- New `paymentListDisplayed` event emitted only once per mount.
- Payment methods filtered-out by `filterStoredPaymentMethods` should be included in `unavailablePaymentMethods`
- If all payment methods are unavailable, all of them should be on `unavailablePaymentMethods`
- Excludes UPI sub-variants from the `unavailablePaymentMethods` when the UPI parent is available.

---

**Closes**:  

---


packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
```diff
@@ -184,6 +199,19 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
...
+     */
+    private createAnalyticsPaymentMethods(paymentMethods: Record<string, any>[]): AnalyticsPaymentMethod[] {
+        return paymentMethods.map(({ type, brand, brands, fundingSource, displayMode }) => ({
+            type,
+            ...(brand !== undefined && { brand }),
+            ...(brands !== undefined && { brands }),
+            ...(fundingSource !== undefined && { fundingSource }),
+            displayMode
+        }));
+    }
```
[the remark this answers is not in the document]
FabioRomeiro: `createAnalyticsPaymentMethods` only receives the output of `getReadyPaymentMethods`/`getUnavailablePaymentMethods`, which never emit null/undefined entries. The filter would guard an unreachable state

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -32,24 +40,55 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
+            this.props.onElementsCreated([...instantElements, ...storedElements, ...elements, ...fastlaneElements]);
+
+            const dropinReadyEvent = new AnalyticsInfoEvent({
+                type: InfoEventType.Ready,
+                component: 'dropin',
+                paymentMethods,
+                unavailablePaymentMethods
+            });
+            this.props.core.modules.analytics.sendAnalytics(dropinReadyEvent);
+        });
```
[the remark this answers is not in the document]
FabioRomeiro: That is a good point. The catch wasn't there before, but it does make sense to add it. I'll just drop the raw error?.message and let Status.Error fall back to the localized error.message.unknown, so shoppers don't see internal error text.

.changeset/dropin-ready-payment-methods-event.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': minor
+---
+
+New: Drop-in emits a new `ready` analytics event reporting the payment methods it rendered (each with `displayMode`) in display order, plus the `/paymentMethods` items it did not render (`unavailablePaymentMethods`).
```
ribeiroguilherme: Changeset description needs to be updated to reflect the correct event name

packages/lib/src/components/Dropin/types.ts
```diff
@@ -150,9 +153,22 @@ export type onOrderCancelType = (
...
+export interface PaymentMethodDisplayModeEntry {
+    displayMode: PaymentMethodDisplayMode;
+    paymentMethods: Array<PaymentMethod | StoredPaymentMethod>;
+}
+
+export interface BuildPaymentMethodsResult {
+    storedPaymentMethodElements: Promise<UIElement[]> | UIElement[];
+    paymentMethodElements: Promise<UIElement[]> | UIElement[];
+    instantPaymentMethodElements: Promise<UIElement[]>;
+    fastlanePaymentMethodElement: Promise<UIElement[]> | UIElement[];
```
ribeiroguilherme: Aren't those objects always a `Promise<UIElement[]>` ? Can they be an UIElement not wrapped in a Promise?

packages/lib/src/components/Dropin/Dropin.tsx
```diff
@@ -166,21 +172,34 @@ class DropinElement extends UIElement<DropinConfiguration> implements IDropin {
...
             ? createElements([fastlanePaymentMethod], paymentMethodsConfiguration, dropinProps, this.core)
             : [];
 
-        return [storedElements, elements, instantPaymentElements, fastlanePaymentElement];
+        const paymentMethodDisplayModes: PaymentMethodDisplayModeEntry[] = [
+            { displayMode: 'fastlane', paymentMethods: fastlanePaymentMethod ? [fastlanePaymentMethod] : [] },
+            { displayMode: 'instant', paymentMethods: instantPaymentMethods },
+            { displayMode: 'stored', paymentMethods: storedPaymentMethods },
+            { displayMode: 'regular', paymentMethods: paymentMethods }
+        ];
```
ribeiroguilherme: This `paymentMethodDisplayModes` can also be created directly in the `DropinComponent` `prepareDropinData()` method, no? 

In this way there is no need to touch this file nor change its logic. We can keep the Analytics event build up in the other file.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -32,24 +40,59 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
+                    stored: storedElements,
+                    regular: elements
+                };
+
+                const availablePaymentMethods = getReadyPaymentMethods(
+                    paymentMethodDisplayModes.map(e => e.displayMode),
+                    elementsByDisplayMode,
+                    this.props.core
+                );
+                const unavailablePaymentMethods = getUnavailablePaymentMethods(paymentMethodDisplayModes, elementsByDisplayMode);
```
ribeiroguilherme: Can't we return the `availablePaymentMethods` and `unavailablePaymentMethods` in the expected shape that the analytic event wants? 

By doing that, seems like there is no need to create a new type as well (`PaymentMethodWithDisplayMode`) nor adding logic to the `AnalyticsInfoEvent` 

I imagine you could do some sort of:

```js
const dropinPaymentListDisplayedEvent = new AnalyticsInfoEvent({
      type: InfoEventType.PaymentListDisplayed,
      component: 'dropin',
      availablePaymentMethods: createAvailablePaymentsList(...),
      unavailablePaymentMethods: createUnavailablePaymentsList(...)
});
```

The `createAvailablePaymentsList()` / `createUnavailablePaymentsList()` could return `AnalyticsPaymentMethod[]`

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
```diff
@@ -16,6 +23,8 @@ type AnalyticsInfoEventProps = {
     cdnUrl?: string;
     selectedValue?: string;
     presentedValues?: Array<string>;
+    availablePaymentMethods?: Record<string, any>[];
+    unavailablePaymentMethods?: Record<string, any>[];
```
ribeiroguilherme: Did you forget to type the  `any`?

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -32,24 +40,59 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
-            ([storedPaymentElements, elements, instantPaymentElements, fastlanePaymentElement, orderStatus]) => {
+        void Promise.all([
+            storedPaymentMethodElements,
+            paymentMethodElements,
+            instantPaymentMethodElements,
+            fastlanePaymentMethodElement,
+            orderStatusPromise
+        ])
+            .then(([storedElements, elements, instantElements, fastlaneElements, orderStatus]) => {
+                const elementsByDisplayMode: Record<PaymentMethodDisplayMode, UIElement[]> = {
```
ribeiroguilherme: It took quite some time to wrap my head around this change 😅  

If I understood it right, you are basically passing from `Dropin.tsx` to here an object describing the _expected_ payment method display modes (`paymentMethodDisplayModes`) . 

Then in here, you compare that to the `elementsByDisplayMode` which contains  only the _available_ payment methods, and from that you determine the analytics event.

Maybe it is because the naming of the variables are very similar. I don't have a good suggestion here, I'm just sharing that it is a bit confusing IMO.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -32,24 +40,59 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
-        );
+                this.props.onElementsCreated([...instantElements, ...storedElements, ...elements, ...fastlaneElements]);
+
+                const dropinPaymentListDisplayedEvent = new AnalyticsInfoEvent({
+                    type: InfoEventType.PaymentListDisplayed,
+                    component: 'dropin',
+                    availablePaymentMethods,
+                    unavailablePaymentMethods
+                });
+                this.props.core.modules.analytics.sendAnalytics(dropinPaymentListDisplayedEvent);
```
ribeiroguilherme: I would suggest to move the code of _build+send_ the analytics event to a separated method, which runs after the Drop-in is created.

In the way it is right now, if something goes wrong with the analytics code, the whole Drop-in would not render.

Perhaps create its own method (e.g. `reportPaymentMethodList()`) and encapsulate all the logic there within a try-catch . If something goes wrong there, it would not affect the Drop-in rendering.

packages/lib/src/components/Dropin/types.ts
FabioRomeiro: They used to, but a recent change on main just updated that :)

packages/lib/src/components/Dropin/Dropin.tsx
FabioRomeiro: Those are 2 different data types, one is "raw payment methods" and the other "UIElements".

This is to prevent duplicate calls to `splitPaymentMethods`, which might cause a desync in the logic of how elements are built, thus making the event data unreliable.

But the chances of that happening are low, for the sake of readability it might be worth duplicating it for now

packages/lib/src/components/Dropin/components/DropinComponent.tsx
FabioRomeiro: Sure, that way we can focus things related to that event in one place.

I chose to go with that first approach based on the existing `createAnalyticsConfigData` that cleans up the event payload in the `AnalyticsInfoEvent` class.

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
FabioRomeiro: Nice catch. I was following `configData` at this point, that was before implementing the rest of the logic on the page.

packages/lib/src/components/Dropin/components/DropinComponent.tsx
FabioRomeiro: That is actually good feedback.

The changes were made focusing on avoiding duplicating logic, keeping the codebase standard, and keeping the elements creation closely related to the event list representation.

But like I mention in the other comment, your understanding of the codebase is as important as the data reliability, so let me know if with the latest changes everything is more clear 😄

packages/lib/src/components/Dropin/components/DropinComponent.tsx
FabioRomeiro: Fair point!

FabioRomeiro: Btw, on the catch block I added a console.warn, like it is for the `sendAnalytics` method itself. Lmk if you think we should take another approach here

packages/lib/src/components/Dropin/components/DropinComponent.tsx
```diff
@@ -55,12 +58,58 @@ export class DropinComponent extends Component<DropinComponentProps, DropinCompo
...
+                this.setStatus('error');
+            });
 
         this.onOrderCancel = this.getOnOrderCancel();
     };
 
+    /**
+     * Builds and sends the payment list displayed analytics event.
+     */
+    private reportPaymentMethodList = (elementsByDisplayMode: Record<PaymentMethodDisplayMode, UIElement[]>): void => {
```
ribeiroguilherme: Minor thing and also matter of preference, but I would rename the `elementsByDisplayMode` to `availableElementsByDisplayMode`. 

This way it would be more clear that they are the **available** ones , otherwise the name conflicts a bit with `paymentMethodDisplayModes`

ribeiroguilherme (dismissed):
LGTM

packages/lib/src/components/Dropin/components/DropinComponent.tsx
ribeiroguilherme: I think that is alright for now. It is not relevant to let merchants know that the analytics failed at this step

ribeiroguilherme (dismissed):

ribeiroguilherme (approved):

PR #4082, opened 2026-07-13
Analytics - Reporting Core properties

ribeiroguilherme:
## 📋 Pull Request Checklist

---

## 📝 Summary

Adds a one-time analytics event that captures the global checkout configuration the merchant passes when initializing the SDK. Previously we only tracked configuration for individual payment method components, not the checkout-level setup (environment, locale, amount, callbacks, etc.).

This enables data-driven insights into how merchants configure the SDK globally, helps detect misconfigurations, and improves SDK defaults/documentation — while obfuscating sensitive data before transmission.

### What changed
- New "initialized" checkout event: On AdyenCheckout.initialize(), an Info event is sent with component: 'checkout' and type: 'initialized', carrying the merchant configuration as configData.
- Sent once per instance: Guarded by a flag so it is not re-sent when update() re-runs initialization. The event is queued and dispatched once the Analytics module has a checkoutAttemptId; it is skipped entirely when analytics is disabled.
- Reports the raw merchant props, not the internal this.options (which has defaults applied and is normalized/merged with session data). This reflects what the merchant actually set. Captured as a shallow copy at construction time.
- Field obfuscation in the shared AnalyticsInfoEvent config transformer:
  - Masked with `<masked>`: clientKey, session, paymentMethodsResponse, translations
  - Functions → `<function>`
  - Arrays → comma-joined string, objects → JSON stringified (both capped at 128 chars)
  - Primitives kept as-is
  - Excluded entirely: internal fields (`_environmentUrls`, `loadingContext`), Drop-in-specific fields (`paymentMethodsConfiguration`, `paymentMethodComponents`, `order`), injected props, and PII fields
  - Type-safety: excluded/masked field lists now use satisfies Array<keyof CoreConfiguration | keyof DropinConfiguration> so typos/renames fail compilation.
---

## 🧪 Tested scenarios
- Manual test (advanced flow and sessions flow) inspecting the new Analytics event being sent
- Added unit tests

### DIFF
--- a/packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
+++ b/packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
@@ -1,6 +1,9 @@
 import { AbstractAnalyticsEvent, AnalyticsEventCategory } from './AbstractAnalyticsEvent';
 import { mapErrorCodesForAnalytics } from '../utils';
 
+import type { CoreConfiguration } from '../../types';
+import type { DropinConfiguration } from '../../../components/Dropin/types';
+
 type AnalyticsInfoEventProps = {
     type: InfoEventType;
     component: string;
@@ -80,10 +83,10 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
     private readonly type: InfoEventType;
 
     /**
-     * Component config data set by the merchant. Sent only in 'rendered' events
+     * Component config data set by the merchant. Sent on 'rendered' events or 'initialized' events
      * @private
      */
-    private readonly configData?: Record<string, string | boolean>;
+    private readonly configData?: Record<string, string | boolean | number | null>;
     private readonly target?: UiTarget;
     private readonly issuer?: string;
     private readonly isExpress?: boolean;
@@ -116,7 +119,8 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
         if (props.validationErrorCode) this.validationErrorCode = props.validationErrorCode;
         if (props.validationErrorMessage) this.validationErrorMessage = props.validationErrorMessage;
         if (props.presentedValues) this.presentedValues = props.presentedValues;
-        if (this.type === InfoEventType.rendered) {
+
+        if (this.type === InfoEventType.rendered || (this.type === InfoEventType.Initialized && props.configData)) {
             this.configData = this.createAnalyticsConfigData(props?.configData);
         }
 
@@ -130,8 +134,11 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
      * Set of properties that must not be included when creating the configData for Analytics
      * @private
      */
-    private get configDataExcludedFields() {
-        const DROPIN_FIELDS = ['paymentMethodsConfiguration'];
+    private get configDataExcludedFields(): string[] {
+        const DROPIN_FIELDS = ['paymentMethodsConfiguration', 'paymentMethodComponents'] satisfies Array<keyof DropinConfiguration>;
+
+        const CORE_INTERNAL_FIELDS = ['_environmentUrls', 'loadingContext'] satisfies Array<keyof CoreConfiguration>;
+
         const FIELDS_INJECTED_BY_DROPIN = [
             'elementRef',
             'isDropin',
@@ -140,8 +147,7 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
             'paymentMethodId',
             'isInstantPayment',
             'type'
-        ];
-        const PII_FIELDS = ['data', 'holderName', 'shopperEmail', 'email', 'telephoneNumber', 'clickToPayConfiguration'];
+        ] satisfies Array<keyof DropinConfiguration>;
 
         /**
          * TODO: Many unit tests are passing 'modules' as props, which leads to circular structure issue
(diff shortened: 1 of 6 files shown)


packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
```diff
@@ -140,40 +147,57 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
...
+                }
+
+                if (this.configDataMaskedFields.includes(key)) {
+                    result[key] = '<masked>';
+                } else if (typeof value === 'function') {
+                    result[key] = '<function>';
+                } else if (Array.isArray(value)) {
+                    result[key] = value.join(', ').substring(0, MAX_STRING_LENGTH);
+                } else if (typeof value === 'object' && value !== null) {
+                    result[key] = JSON.stringify(value).substring(0, MAX_STRING_LENGTH);
```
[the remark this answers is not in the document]
johnayeni: Can we simplify this block by having some early returns instead of multiple else if statements?

packages/lib/src/core/core.test.ts
```diff
@@ -190,6 +190,80 @@ describe('Core', () => {
...
+
+        beforeEach(() => {
+            sendAnalyticsSpy = jest.spyOn(Analytics.prototype, 'sendAnalytics').mockImplementation(() => {});
+        });
+
+        afterEach(() => {
+            sendAnalyticsSpy.mockRestore();
+        });
+
+        test('should send an "initialized" info event for the checkout with obfuscated config', async () => {
```
m1aw, line 204: Can we add more tests here? 
For example PII, data, etc?

m1aw:
Also I couldn't find the ticket associated with this change?

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
```diff
@@ -116,7 +119,8 @@ export class AnalyticsInfoEvent extends AbstractAnalyticsEvent {
         if (props.validationErrorCode) this.validationErrorCode = props.validationErrorCode;
         if (props.validationErrorMessage) this.validationErrorMessage = props.validationErrorMessage;
         if (props.presentedValues) this.presentedValues = props.presentedValues;
-        if (this.type === InfoEventType.rendered) {
+
+        if (this.type === InfoEventType.rendered || (this.type === InfoEventType.Initialized && props.configData)) {
             this.configData = this.createAnalyticsConfigData(props?.configData);
         }
```
FabioRomeiro, line 125: `createAnalyticsConfigData` is a fragile guardrail. I understand it was already like that before you, but this is a good opportunity to discuss a safer approach on how we handle the data we pass to events.
The safer approach would be mapping the allowed props we wanna surface. It could be through a compile-time classification of the core config, in case we don't wanna solely rely on Typescript.
Personally, I believe the cost of securing this is worth it to protect against the risk of including a new sensitive property in the props and forgetting to update this method.
What do you think?

packages/lib/src/core/core.test.ts
ribeiroguilherme: I added tests on the `AnalyticsInfoEvent.test.ts` file

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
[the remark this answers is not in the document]
ribeiroguilherme: I updated the code to have early returns

ribeiroguilherme:
> Also I couldn't find the ticket associated with this change?

@m1aw the ticket is COSDK-1131

packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts
ribeiroguilherme: Mapping allowed props would be a good approach but that is high effort one, specially because we need to combine all props from all component interfaces and also make sure that the list is always up-to-date.

For now we are investigating if the backend can perform some checks on the Analytic event received, so we let all that logic of masking/removing PII data on the backend rather than on the frontend.

FabioRomeiro: Got it!

dimitra-adyen:
## PR Review: Analytics - Reporting Core Properties

### Summary

This PR adds a one-time `initialized` analytics event that captures the raw merchant checkout configuration at SDK init time, with sensitive fields obfuscated before transmission. The overall design is solid — the use of `satisfies Array<keyof ...>` for compile-time safety on exclusion/masking lists is a nice touch, and the test coverage is comprehensive. However, there is one potentially breaking behavior change that needs explicit attention.

---

### 🚨 Critical: Possible Behavior Regression in Sessions Flow

**File:** `packages/lib/src/core/core.ts`, `initializeCore()`

The PR removes the spread of `...rest` from the session setup response into `setOptions()`:

```ts
// Before
const { amount, shopperLocale, countryCode, paymentMethods, ...rest } = sessionResponse;
this.setOptions({ ...rest, amount: ..., locale: ..., countryCode: ... });

// After
const { amount, shopperLocale, countryCode, paymentMethods } = sessionResponse;
this.setOptions({ amount: ..., locale: ..., countryCode: ... });
```

`...rest` from `CheckoutSessionSetupResponse` contains `{ id, sessionData, expiresAt, returnUrl, configuration }`. The most concerning dropped field is **`configuration`** (`SessionConfiguration`: `installmentOptions`, `enableStoreDetails`), which was previously merged into `this.options`. If any component reads `this.options.installmentOptions` or `this.options.enableStoreDetails`, those would now resolve as `undefined` in the sessions flow.

`Session.setupSession()` does separately store `configuration` in `this.session.configuration`, so components that access it via `core.session.configuration` are fine — but the change is unrelated to this PR's stated goal and has no mention in the description.

**Action needed:** Either document why this is intentional (and verify no component reads these via `this.options`), or restore the spread selectively.

---

### ⚠️ Notable Issues

**1. Type unsafety in `serializeConfigValue` for numeric primitives**

`packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts`

The method is declared to return `string | boolean`, but numeric primitives (e.g. `amount: 5000`) fall through to the final `return value as string | boolean` — a lying cast. The `configData` property was already widened to `Record<string, string | boolean | number | null>` to accommodate numbers, so the fix is aligning the function signature and adding an explicit branch:

```ts
private serializeConfigValue(key: string, value: unknown): string | boolean | number {
    const MAX_STRING_LENGTH = 128;
    if (this.configDataMaskedFields.includes(key)) return '<masked>';
    if (typeof value === 'function') return '<function>';
    if (Array.isArray(value)) return value.join(', ').substring(0, MAX_STRING_LENGTH);
    if (typeof value === 'object' && value !== null) {
        try { return JSON.stringify(value).substring(0, MAX_STRING_LENGTH); } catch { return '[object]'; }
    }
    return value as string | boolean | number; // string, boolean, or number primitives
}
```

**2. `InfoEventType.Initialized` naming is inconsistent with the existing enum style**

`packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts`

All existing members use lowercase camelCase (`clicked`, `rendered`, `selected`, `validationError`, `sdkDownloadInitiated`, etc.), but the new entry uses PascalCase `Initialized`. The runtime string value `'initialized'` is correct, but the enum key should follow convention: `initialized = 'initialized'`.

**3. Getter allocation inside the serialization loop**

`packages/lib/src/core/Analytics/events/AnalyticsInfoEvent.ts`

`configDataExcludedFields` and `configDataMaskedFields` are getters that each allocate a new array on every access. They're called once per key in the loop. Snapshot them before iterating to avoid unnecessary allocations:

```ts
const excluded = this.configDataExcludedFields;
const masked = this.configDataMaskedFields;
for (const [key, value] of Object.entries(config)) {
    if (excluded.includes(key)) continue;
    result[key] = this.serializeConfigValue(key, value, masked);
}
```

---

### 💬 Nitpicks

- **PR description inaccuracy**: The description says `order` is "excluded entirely" but it is actually in `configDataMaskedFields` (masked with `<masked>`), not excluded.
- **Redundant null guard**: `if (!config) return {}` inside `createAnalyticsConfigData` is unreachable — the call site already guards on `props.configData` being truthy.
- **Shallow copy caveat**: `this.initialConfiguration = { ...props }` is a shallow copy. If a merchant mutates a nested config object between construction and `initialize()`, the analytics event reflects the mutation. Very low risk in practice, but worth a brief code comment to set expectations.

PR #4084, opened 2026-07-14
feat: propagate healthcare field to onBinLookup

dimitra-adyen:
## 📋 Pull Request Checklist

- [X] I have added unit tests to cover my changes.
- [X] I have tested the changes manually in the local environment.
---

## 📝 Summary

The BE is adding a healthcare boolean field at the brand level to the `/binLookup` response. Since the SDK re-maps the binLookup response before exposing it to merchants via `onBinLookup`, we will explicitly pass the field through.

The field is only included in the callback when the BE returns it, it is never defaulted to false when absent.
When at least one brand has the healthcare field (true/false), the healthcare array is returned inside the `onBinLookup`. In any other case the field gets omitted.

---

## 🧪 Tested scenarios

- healthcare: true in at least one brand returned by BE → included in onBinLookup callback
- healthcare absent from all brands from BE response → field omitted entirely from onBinLookup callback
- healthcare: false in at least one brand returned by BE → included in onBinLookup callback

---

**Closes**:  

COSDK-1266
---

### DIFF
--- a/packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
+++ b/packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
@@ -89,8 +89,10 @@ export default parent => {
                                 acc.detectedBrands.push(item.brand);
                                 // Also add the paymentMethodVariants (more granular description of the txvariant)
                                 acc.paymentMethodVariants.push(item.paymentMethodVariant);
-                                // Add healthcare to the healthcare array
-                                acc.healthcare.push(item.healthcare);
+                                // Add healthcare to the healthcare array, keyed by brand (only when the field is present)
+                                if (item.healthcare !== undefined) {
+                                    acc.healthcare.push({ [item.brand]: item.healthcare });
+                                }
 
                                 // Add supported brand objects to the supportedBrands array
                                 if (item.supported === true) {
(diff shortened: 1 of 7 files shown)


packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
```diff
@@ -87,6 +89,8 @@ export default parent => {
                                 acc.detectedBrands.push(item.brand);
                                 // Also add the paymentMethodVariants (more granular description of the txvariant)
                                 acc.paymentMethodVariants.push(item.paymentMethodVariant);
+                                // Add healthcare to the healthcare array
+                                acc.healthcare.push(item.healthcare);
```
sponglord: The other arrays that end up on the object that we send to the `onBinLookup` callback are all clearly aligned with the brand that they are associated with e.g.
```
{
    "type": "card",
    "detectedBrands": [
        "mc",
        "maestro"
    ],
    "supportedBrands": [
        "mc",
        "maestro"
    ],
    "paymentMethodVariants": [
        "mcdebit",
        "maestro"
    ],
    "issuingCountryCode": "BE"
}
```
With the addition of this healthcare array though, I guess we're just going to get
`"healthcare" : [true, false]`

Then the merchant needs to do the "heavy lifting" :-p of matching the booleans to a txVariant.

I'm wondering if the `healthcare` array should be objects, pre-matching txVariant & boolean? i.e.
`"healthcare" : [{"mc": true}, {"maestro": false}]`

dimitra-adyen: Applied the suggestion

sponglord (approved):

PR #4085, opened 2026-07-15
Fixed: aria-checked missing in the Dropin payment method list if openFirstPaymentMethod set to false

### LINKED ISSUE #3824: Accessibility: role="radio" payment option missing required aria-checked when no selection
**Describe the bug**
The payment payment option buttons are implemented with `role="radio"` but does not have `aria-checked="false"` when the user hasn’t selected any payment method yet. According to the WAI‑ARIA specification, when using role="radio", the aria-checked attribute is required (not optional).

Spec reference: 
https://www.w3.org/TR/wai-aria-1.2/#radio

**To Reproduce**
1. Go to any payment page which use `Dropin` from `@adyen/adyen-web/auto`.
2. Ensure no payment method is selected
3. Inspect the payment method element(s) that use role="radio" (e.g., in DevTools Elements panel).
4. Observe that `aria-checked` is missing (it should be present and set to "false" for unselected radios).

**Expected behavior**
All elements with `role="radio"` should always include `aria-checked`.

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: macOS 26.2
 - Browser chrome 144.0.7559.135 (Official Build) (arm64)
 - @adyen/adyen-web@6.29.0

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

This issue fixed #3824. The issue was reproducible when using `openFirstPaymentMethod` was set to false, in this case the `isSelected` condition would evaluate wrongly to null hiding the key from the element.

Fixed by swapping to optional chaining on the comparison.

---

## 🧪 Tested scenarios

---

**Closes**:  #3824 

---

### DIFF
--- a/.changeset/thin-spies-film.md
+++ b/.changeset/thin-spies-film.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed: aria-checked missing in the Dropin payment method list if openFirstPaymentMethod set to false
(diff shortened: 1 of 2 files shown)


dimitra-adyen (approved):
👌 tested

PR #4087, opened 2026-07-15
 PayPal v6 (Task 1) — Core SDK Infrastructure (Loading, Token & Eligibility)

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

Adds the foundational infrastructure for loading and initialising the PayPal JS SDK v6, gated behind a new `usePayPalV6` configuration option. This is phase 0 — it wires up SDK loading, client token retrieval, instance creation, and eligibility checks, but does not yet render a functional v6 button (placeholder only). The existing v5 integration is untouched.

- **`PayPalSdkLoader`** — injects the PayPal v6 SDK `<script>` via the shared `Script` util. The URL is **environment-aware**, resolved from config (`PAYPAL_SDK_URL_SANDBOX` / `PAYPAL_SDK_URL_PRODUCTION`) rather than hardcoded. Accepts an optional `nonce` attribute for CSP, exposes `isSdkLoaded()`, and throws `AdyenCheckoutError('SCRIPT_ERROR', ...)` on load failure.
- **`PayPalService`** — orchestrator for the init chain: awaits the script, fetches the client token, calls `paypal.createInstance({ clientToken, components, pageType })`, then runs `findEligibleMethods({ currencyCode, countryCode, paymentFlow })` (handling zero-auth / vault flows). Exposes `initialize()`, `isSdkLoaded()`, `getInstance()`, and `getEligiblePaymentMethods()`.
- **`request-paypal-oauth-token`** — `httpPost` helper that calls the Adyen utility token endpoint using `configuration.merchantId` and returns the `clientToken`, surfacing failures as fatal errors.
- **`Paypal.tsx`** — when `usePayPalV6` is set, instantiates the loader/service in the constructor (passing `environment`, `nonce`, `vault`), kicks off initialisation, and overrides `isAvailable()` to await SDK readiness and check `isEligible('paypal')` (hiding the component when ineligible or on failure).
- **`config.ts`** — renames v5 constants (`PAYPAL_JS_URL_V5`, `ADYEN_CLIENTID_V5_TEST/LIVE`) for clarity and adds the v6 SDK URL constants; updates `get-paypal-settings` / `get-paypal-url` references accordingly.
- **`types.ts`** — adds the `usePayPalV6` prop with `vault` and `nonce` options.
- Adds PayPal SDK type definitions (`paypal-js-types.ts`, `custom.d.ts`) and minor `ApplePay` loader/test cleanup.

---

## 🧪 Tested scenarios

- [x] `PayPalSdkLoader` — script load success/failure (`AdyenCheckoutError`), sandbox vs. production URL selection, `nonce` attribute set when provided and omitted when not.
- [x] `request-paypal-oauth-token` — endpoint called with correct path and `merchantId`.
- [x] `cd packages/lib && yarn test src/components/PayPal` passes.

---

- COSDK-1317

**Closes**:  

---


packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -36,11 +38,53 @@ class PaypalElement extends UIElement<PayPalConfiguration> {
...
+            });
+
+            this.paypalService = new PayPalService({
+                loadingContext: this.props.loadingContext ?? '',
+                clientKey: this.props.clientKey ?? '',
+                merchantId: this.props.configuration?.merchantId ?? '',
+                sdkLoader,
+                countryCode: this.props.countryCode ?? '',
+                amount: this.props.amount,
+                vault: Boolean(paypalV6Props.vault)
```
ribeiroguilherme, line 65: Do we need to apply these nullish coalescing operators here? All these values are required by default. 

Does this prevent something to happen downstream or would we have the same flow/behavior if we remove these assignments?

packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -36,11 +38,53 @@ class PaypalElement extends UIElement<PayPalConfiguration> {
...
+            });
+
+            void this.paypalService.initialize();
+        }
+    }
+
+    public override async isAvailable(): Promise<void> {
+        if (this.props.usePayPalV6) {
+            if (!this.paypalService) {
+                return Promise.reject(new Error('PayPal service not initialized'));
```
ribeiroguilherme: Instead of using the default`Error` object, we can throw `new AdyenCheckout('ERROR', 'PayPal is not available');`. We do this with ApplePay component for example.

packages/lib/config/rollup.prod.js
```diff
@@ -136,15 +136,15 @@ export default () => {
         {
             input: 'dist/temp-types/src/types.d.ts',
             output: [{ file: './dist/cjs/index.d.cts', format: 'commonjs' }],
-            external: [/\.scss$/u],
+            external: [/\.scss$/u, '@paypal/paypal-js/sdk-v6'],
```
ribeiroguilherme, line 139: Why do we need to put the paypal types here?

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -0,0 +1,117 @@
...
+        if (this.loadingPromise === undefined) {
+            return Promise.reject(new Error('PayPal SDK not loaded'));
+        }
+
+        return this.loadingPromise;
+    }
+
+    private async createPayPalSdkInstance(clientToken: string): Promise<PayPalSdkInstance> {
+        const paypal = window.paypal;
+        const createInstance = paypal?.v6?.createInstance || paypal?.createInstance;
```
ribeiroguilherme, line 74: What are the scenarios in which the `v6` namespace would be created?

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -0,0 +1,117 @@
...
+        this.sdkInstance = await createInstance({
+            clientToken,
+            components: ['paypal-payments', 'venmo-payments'] satisfies PayPalComponents,
+            pageType: 'checkout'
+        });
+
+        return this.sdkInstance;
+    }
+
+    private async createPayPalPaymentMethods(): Promise<void> {
```
ribeiroguilherme: Is it creating payment methods or is it creating the eligibility list?

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        const isSdkLoaderLoadedPromise = this.sdkLoader.isSdkLoaded();
+        const tokenDataPromise = requestPayPalOauthToken(this.loadingContext, { clientKey: this.clientKey, merchantId: this.merchantId });
+
+        this.loadingPromise = Promise.all([isSdkLoaderLoadedPromise, tokenDataPromise])
+            .then(([_loadedSdk, tokenData]) => {
+                return tokenData.clientToken;
+            })
+            .then(this.createPayPalSdkInstance)
+            .then(this.createPayPalPaymentMethods);
```
ribeiroguilherme: What happens if the token creation endpoint fails? Do we manage to propagate that to the `PayPal.isAvailable()` and then report that PayPal is not available at all?

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -0,0 +1,117 @@
+import { PaymentAmount } from '../../../types';
+import { PayPalSdkLoader } from './PayPalSdkLoader';
+import { PayPalComponents, PayPalEligiblePaymentMethods, PayPalPaymentFlow, PayPalSdkInstance } from '../paypal-js-types';
```
ribeiroguilherme: Minor: better to use `import type` for the type-only imports

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        return this.loadingPromise;
+    }
+
+    private async createPayPalSdkInstance(clientToken: string): Promise<PayPalSdkInstance> {
+        const paypal = window.paypal;
+        const createInstance = paypal?.v6?.createInstance || paypal?.createInstance;
+
+        if (!createInstance) {
+            return Promise.reject(new Error('PayPal SDK `createInstance` is not available'));
```
ribeiroguilherme: Better to use `AdyenCheckoutError` rather than the generic error object. There are few places in this PR that this is happening

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        this.paymentMethods = await this.sdkInstance.findEligibleMethods({
+            currencyCode: this.amount?.currency,
+            // @ts-expect-error: @paypal/paypal-js is missing countryCode in the types
+            countryCode: this.countryCode,
+            paymentFlow
+        });
+
+        return Promise.resolve();
```
ribeiroguilherme: The return `Promise.resolve()` here is redundant, it can be removed

packages/lib/src/components/PayPal/Paypal.tsx
johnayeni: So most of these props are actually marked as optional in our types https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/internal/UIElement/types.ts#L66-L88. The reason for this nullish coalescing is to stay true to what our types say as there is a chance these values could be omitted.

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Good point, I have added a catch block to the `initialise` PayPal class that would forwards the error to the onError callback if it happens https://github.com/Adyen/adyen-web/blob/ec564645c3af7573ed1cd4c35e0d3faf8bb094ab/packages/lib/src/components/PayPal/Paypal.tsx#L68-L75

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Good catch, done

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Done 👍🏿

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Removed 👍🏿

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: I documented the reason here https://github.com/Adyen/adyen-web/pull/4087/changes#diff-b479d87d32b74fa4ea1fdb66a8d8989de466f71d68e83202e03f5833092100d5R37-R40

packages/lib/src/components/PayPal/Paypal.tsx
johnayeni: Done

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: True, that makes more sense. I have renamed this method to createEligiblePaymentMehthods

packages/lib/config/rollup.prod.js
johnayeni: This issue happens because of the export map configuration in paypal-js:

```json
"exports": {
  ".": {
    "types": "./types/index.d.ts",
    "import": "./dist/esm/paypal-js.js",
    "require": "./dist/cjs/paypal-js.js"
  },
  "./sdk-v6": {
    "types": "./types/v6/index.d.ts",
    "import": "./dist/v6/esm/paypal-js.js"
  }
}
```

Because `@paypal/paypal-js/sdk-v6` is a mapped subpath rather than a physical directory on disk, Rollup cannot automatically resolve and inline its types into our generated `index.d.ts`.

By default, Rollup treats dependency imports as external packages. This works fine for the root import (`@paypal/paypal-js`) because Rollup can resolve its types properly through the export map. However, it fails to do so for the `/sdk-v6` subpath, triggering a build warning.

Explicitly adding `@paypal/paypal-js/sdk-v6` to our external array tells Rollup that this external resolution behavior is intentional and suppresses the warning.

packages/lib/src/components/PayPal/Paypal.tsx
ribeiroguilherme: Yeah, fair point.. They are "optional" because they are not required for the `UIElement` interface,  but internally we create these properties..  

Having  two difference interfaces such as `UIElement` and `InternalUIElement` would fix such problem I think.. but that is something for another time 😄

ribeiroguilherme (approved):

ribeiroguilherme (approved):

PR #4089, opened 2026-07-16
Fix: resolve TypeScript issues in triggerBinLookUp and CardElement

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

- Refactor `triggerBinLookUp` into a named export and type its argument as `CardElement | CustomCardElement`, replacing the untyped `parent` parameter and using `type` imports to avoid circular dependencies.
- Add proper types to `CardElement` (`clickToPayRef`, `setClickToPayRef`, `handleClickToPaySubmit`, `onBlur`) and allow `null` in `processBinLookupResponse`.
- Type `ClickToPayProvider`'s `setClickToPayRef` and initialize the ref with `setStatus` up front.
- Type `DoChallenge3DS2.render` props/state and guard `inputValue` against `undefined`.
- Widen `CardBinLookupData` types (`detectedBrands`/`supportedBrands` nullable, add `paymentMethodVariants`).
- Add test cases covering `triggerBinLookUp`.

---

## 🧪 Tested scenarios

- [x] `cd packages/lib && yarn test` passes for `triggerBinLookUp`
- [x] Type check / build passes
- [x] BIN lookup still works for Card and CustomCard components
- [x] 3DS2 challenge flow renders and submits correctly

---

**Closes**:  

---


packages/lib/src/components/Card/Card.tsx
```diff
@@ -266,7 +266,7 @@ export class CardElement extends UIElement<CardConfiguration> {
         }
     };
 
-    private onBlur = (obj: ComponentFocusObject) => {
+    private readonly onBlur = (obj: ComponentFocusObject) => {
```
sponglord, line 269: Since I presume the intent of adding this `readonly` is to assert that this arrow function is an instance property that can only be assigned once - for consistencies sake, the same type safety should also be applied to the `onConfigSuccess` and `onFocus` functions

johnayeni: Good point! Added this too as well for those methods 👍🏿

sponglord (approved):

PR #4091, opened 2026-07-16
fix(card): keep expiry date and CVC format hints visible in error state

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

Related to COSDK-1111 — when a field like Expiry date or CVC enters an error state, the error message replaces the contextual format hint ("Front of card in MM/YY format", "3 digits on back of card"), so screen reader users lose the format guidance exactly when they need it most. These fields are cross-origin secured iframes, so `aria-describedby` can't reference both a hint span and an error span across the frame boundary — the accepted alternative is to fold the hint into the error message itself.

**Approach:** bake the hint directly into the existing error strings, rather than composing them at runtime.
- Updated `cc.dat.910/911/912/913` and `cc.cvc.920/921` in `en-US.json` to include the format hint as a second sentence (e.g. `"Enter a valid expiry date. Front of card in MM/YY format."`). `cc.dat.914` ("about to expire") is a forward-looking warning, not a data-entry error, so left unchanged.
- CVC's hint is brand-dependent (Amex: 4 digits/front vs. everyone else: 3 digits/back), so added explicit `cc.cvc.920.amex`/`cc.cvc.921.amex` variants and moved the brand-aware key selection into `CVC.tsx` itself (previously this logic lived in the caller — centralizing it means both `CardFields.tsx` and `StoredCardFields.tsx` (one-click/stored card flow) pick up the fix consistently, instead of needing the same brand check duplicated in each caller).
- Source strings for the changed/new keys have been submitted to the translation team for the other locales; `en-US.json` is updated in this PR, other locale files will follow once translated.
- (new) Bumped SF version `5.3.0` to support branded error messages on the CVC field.
  - Edited the SF configuration object to pass the new branded error messages for `amex`.

---

## 🧪 Tested scenarios

- Expiry date shows `"<error message>. <format hint>."` when invalid (`cc.dat.910-913`).
- CVC shows the brand-correct hint: "3 digits on back of card" for non-Amex, "4 digits on front of card" for Amex — verified for both `CardFields` (new card) and `StoredCardFields` (stored/one-click card).
- Card number (no configured hint) shows only the plain error message, unchanged.
- A merchant's `customTranslations` override for a base error key renders verbatim, with no hint appended on top.
- New `CVC.test.tsx` and `CardFields.test.tsx` covering the above; `Language.test.ts` unchanged from main.
- `tsc --noEmit`, `eslint`, and the full `packages/lib` Jest suite pass (2056/2057, 1 pre-existing unrelated skip).

---

**Closes**: COSDK-1111


m1aw:
This PRs tests are meant to fail as this functionality is only fully supported with secured fields changes.

packages/lib/src/components/Card/components/CardInput/components/CardFields.tsx
```diff
@@ -35,7 +35,8 @@ export default function CardFields({
     const { i18n } = useCoreContext();
 
     const getError = (errors, fieldType) => {
```
sponglord, line 37: To remove the implicit `any`s this signature could be changed to 
` const getError = (fieldType: string) => {`

and then the function can just be called below:
```
 <CardNumber
    ...
    error={getError(ENCRYPTED_CARD_NUMBER)}
```
and
```
<ExpirationDate
    ...
    error={getError(ENCRYPTED_EXPIRY_DATE)}
```

It was redundant to pass the `errors` argument to the function, since it was already available as a destructured prop.

sponglord: Alternatively the signature could stay as it was and `errors` could just be typed as `SFPErrorMap`

.changeset/expiry-date-error-context-hint.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
```
ScottiBR: Since the SF gains a new feature here (branded CVC aria errors), should the changeset be minor instead of patch?

packages/e2e-playwright/tests/a11y/card/errorPanel.card.visible.spec.ts
```diff
@@ -12,7 +12,7 @@ test('#1 Error panel is present at start, when there are no errors, but is empty
 });
 
 test('#2 Click pay with empty fields and error panel is populated', async ({ card, srPanel, page, browserName }) => {
-    const expectedSRPanelTexts = ['Enter the card number-sr', 'Enter the expiry date-sr', 'Enter the security code-sr'];
+    const expectedSRPanelTexts = ['Enter the card number-sr', 'Enter the expiry date. Front of card in MM/YY format.-sr', 'Enter the security code. 3 digits on back of card.-sr'];
```
ScottiBR: small one - `prettier --check` fails on this line (202 chars vs `printWidth: 150`), it's the only file of the changed ones that does. there's no prettier job in `pr.yml` so CI won't catch it, but it'll show up as noise in the next `lint-staged` run on this file.


separately, feel free to ignore - the sibling spec you touched (`card.contextualTexts.spec.ts`) pulls its copy from `LANG['cc.cvc.920']`, maybe we do the same here so a future translation tweak doesn't break the test with no pointer back to the key.

ScottiBR (dismissed):
Just a few comments, Nice work!

.changeset/expiry-date-error-context-hint.md
m1aw: Done

packages/e2e-playwright/tests/a11y/card/errorPanel.card.visible.spec.ts
m1aw: Fixed the first one. Tbh I didn't notice the lint-staged, also surprised it didn't just auto format.

Yeah, we have tons of tests that need that treatment, I think a Tech Debt topic to tackle them, probably quite an easy fix with AI.

ScottiBR (approved):

Issue #4083, opened 2026-07-13
Drop-in billing address fields duplicate entered input of traditional Chinese characters


abrahamhurtado:
**Describe the bug**
We noticed that typing traditional Chinese characters in the drop-in's address field results in duplicated characters while still typing. This is true with other text-based fields except "Name on card".

This behavior was replicated on iOS 26.5 devices, on Adyen 6.21.0 and 5.56.1

**To Reproduce**

To install the Zhuyin keyboard on iOS:

Settings > General > Keyboard > tap Keyboards > Add New Keyboard > Chinese, Traditional > Zhuyin - Standard

1. Load the Adyen drop-in with billing address form enabled.
2. Start typing on any of the billing address fields, such as Street Name, try this sequence: `ㄘㄜㄕ`, you’ll notice that extra characters get inserted into the field.
3. This however, does not seem to impact the composition of characters, as hitting enter does insert the correct characters.

**Expected behavior**
The expectation is that the entered text would remain as is.

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: iPhone 15 Pro Max
 - OS: 26.5
 - Browser: Safari
 - Version: 26.5

**Additional context**

https://github.com/user-attachments/assets/f0b15aa2-339f-4d1b-8034-aa47f06b40b9

dimitra-adyen:
@abrahamhurtado Thanks for reporting this. This issue should be fixed with the [6.41.1](https://github.com/Adyen/adyen-web/releases/tag/v6.41.1) release.

PR #4095, opened 2026-07-20
fix: duplicate characters issue in address input fields

dimitra-adyen:
## 📋 Pull Request Checklist

---

## 📝 Summary

This PR fixes a bug where address input fields would produce duplicate characters when typing with an IME (used for Chinese, Japanese, Korean, and other languages with composing input systems)

**Root cause**: The formattingFn applied to address fields and specifically `trimValWithOneSpace`, which was modifying the input value mid-composition (removing spaces from the start and end of the input). This interfered with the browser's IME composition lifecycle and caused the composed text to be appended again after the `compositionend` event fired, resulting in duplicate characters.

**Fix**: Removed trimValWithOneSpace from the address formattingFn. The formatter now only strips special characters, handling the space characters `onBlur` only now

E2E fix for flaky screenshot tests was included in this PR.

---

## 🧪 Tested scenarios

- on iPhone 16 in Chrome, Safari and Firefox
  - Typing with an IME (Zhuyin keyboard) in Street, House Number and City fields no longer produces duplicate characters.
  - The above address fields with special characters are still correctly stripped.

---

**Closes**:  #4083 

---

### DIFF
--- a/packages/lib/src/components/internal/Address/validate.formats.ts
+++ b/packages/lib/src/components/internal/Address/validate.formats.ts
@@ -1,6 +1,6 @@
 import { CountryFormatRules, FormatRules } from '../../../utils/Validator/types';
 import { Formatter } from '../../../utils/useForm/types';
-import { getFormattingRegEx, SPECIAL_CHARS, trimValWithOneSpace } from '../../../utils/validator-utils';
+import { getFormattingRegEx, SPECIAL_CHARS } from '../../../utils/validator-utils';
 
 const asCountryFormatRules = <T extends CountryFormatRules>(rules: T): T & CountryFormatRules => rules;
 
@@ -15,7 +15,7 @@ const createFormatByDigits = (digits: number): Formatter => {
 };
 
 const specialCharsRegEx = getFormattingRegEx(SPECIAL_CHARS);
-const formattingFn = val => trimValWithOneSpace(val).replace(specialCharsRegEx, '');
+const formattingFn = (val: string) => val.replace(specialCharsRegEx, '');
 
 export const addressFormatters: FormatRules = {
     postalCode: {
--- a/.changeset/ninety-boats-open.md
+++ b/.changeset/ninety-boats-open.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed: Duplicate characters when typing in address fields using an IME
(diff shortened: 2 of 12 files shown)


packages/lib/src/components/internal/Address/validate.formats.ts
```diff
@@ -1,6 +1,6 @@
 import { CountryFormatRules, FormatRules } from '../../../utils/Validator/types';
 import { Formatter } from '../../../utils/useForm/types';
-import { getFormattingRegEx, SPECIAL_CHARS, trimValWithOneSpace } from '../../../utils/validator-utils';
```
ribeiroguilherme, line 3: The function `trimValWithOneSpace` is now a dead code. It can be removed from the codebase

packages/lib/src/components/internal/Address/validate.formats.ts
```diff
@@ -15,7 +15,7 @@ const createFormatByDigits = (digits: number): Formatter => {
 };
 
 const specialCharsRegEx = getFormattingRegEx(SPECIAL_CHARS);
-const formattingFn = val => trimValWithOneSpace(val).replace(specialCharsRegEx, '');
+const formattingFn = (val: string) => val.replace(specialCharsRegEx, '');
```
ribeiroguilherme, line 18: It is better to add some unit tests to guard the fix, otherwise any future "clean up" can silently revert the problem, reintroducing the bug

ribeiroguilherme (changes requested):

packages/lib/src/components/internal/Address/validate.formats.ts
dimitra-adyen: Updated the validate.format.test.ts with tests to cover the above change.

packages/lib/src/components/internal/Address/validate.formats.ts
dimitra-adyen: Kept the `trimValWithOneSpace` method, to use it inside the `InputBase.tsx` -> `handleBlur` method.
With the previous `.trim()` method, only the leading and trailing spaces were trimmed, meaning that multiple spaces in between would remain intact. 
Since `trimOnBlur` is passed as true only by the `Address` component, I replaced the `.trim()` with the `trimValWithOneSpace` to achieve the same result as before.

ribeiroguilherme (approved):

PR #4096, opened 2026-07-21
feat: add conditional card component funding source validation

FabioRomeiro:
## 📋 Pull Request Checklist

---

## 📝 Summary

Adds conditional **funding-source validation** to the Card component, driven by the BIN lookup result.

**Motivation:**
- Merchants can split the card payment method into separate credit / debit entries via [splitCardFundingSources](https://docs.adyen.com/api-explorer/Checkout/71/post/paymentMethods#request-splitCardFundingSources), but the split is cosmetic today: any card is accepted in any entry.
- The component reports its *configured* funding source on the payment request, not the one BIN lookup resolved, so a credit card typed into the debit component is submitted as `debit`.
- That matters for **payment method personalization**, whose model learns from what shoppers actually pay with. Steering the shopper to the right component makes the reported funding source match the card entered.

**What it does:**

Driven by a new `allowedFundingSources` property in the scheme's `configuration` object in the `/paymentMethods` response.

- Brands returned by BIN lookup are **reordered, not filtered**, so the first brand with an allowed funding source is the one pre-selected and handed to SecuredFields.
- On a **dual-branded card both brands stay selectable**. Picking the disallowed one shows the `cc.num.904` error but keeps the selector on screen, so the shopper can switch back and the card re-encrypts.
- If **every** brand is disallowed, the card is rejected with the same error and no selector is offered, reusing the existing unsupported-card UX.
- Brands that report no funding source are always treated as allowed.
- `onBinLookup` is **unchanged**: it always reports what BIN lookup found, never a filtered or nulled view.
- BIN lookup returns `fundingSource` as an array of unique values per brand; a brand is allowed when any returned value matches `allowedFundingSources`.

**Dependencies:**
- `allowedFundingSources` is populated by the backend in `/paymentMethods`. That side is not implemented yet, so this ships ahead of its data source and is a no-op for every current integration.

**Scope and limitations:**
- Validation is skipped, and the card accepted, whenever the funding source cannot be established: `doBinLookup: false`, unknown BINs, failed lookups, and brands reporting no funding source. Types outside `credit | debit | prepaid` (charge, `deferred_debit`) currently arrive without one and fall into this group.
- A misconfigured `allowedFundingSources` fails open rather than rejecting every card.
- Not applied to Click to Pay, Fastlane, stored cards, or CustomCard. The rule lives in code shared with CustomCard but only the Card component supplies the configuration, so it is inert there.

**Notable changes:**
- `binLookup/extensions.ts`: owns the entire rule, applied at two points — on the BIN lookup result (*is any brand usable?*) and on brand selection (*is the picked brand usable?*). Resetting the selector on rejection also fixes a case where a previous card's selector could linger under an error.
- `binLookup/triggerBinLookUp.ts`: **deliberately unchanged from `main`**. Keeping it purely network-plus-mapping is what preserves the `onBinLookup` contract.
- `binLookup/fundingSource.ts` *(new)*: pure, unit-tested helpers that normalize the configured value, evaluate a brand, and order a brand set.
- `CardNumber.tsx` / `CardFields.tsx`: the untranslated `errorCode` is plumbed down so the brand selector survives this one error, where every other PAN error hides it.
- `Card.tsx` / `Card/types.ts`: `allowedFundingSources` normalized once when props are built; `fundingSource` added to `BrandObject`.
- `core/Errors/constants.ts` + `CardInput/utils.ts`: new error code, registered as a non-blur error.
- Storybook: `SplitFundingSourceTest` gains an `applyFundingSourceValidation` toggle.

**Open items:**
- Merchants get no callback signal for a rejected funding source, since `onBinLookup` reports raw findings by design. Follow-up ticket raised.

---

## 🧪 Tested scenarios

**Unit** (`fundingSource.test.ts`, `Card.test.tsx`) — configured-value normalization (whitespace, casing, unrecognized values failing open); brand evaluation including missing funding source and case-insensitive matching; brand ordering, including that no brand is ever dropped when none is allowed.

**Layer contract** (`extensions.test.ts`, `triggerBinLookUp.test.tsx`) — the error payload handed to SecuredFields and its `detectedBrands`; that a rejected card is never passed to SecuredFields; that a disallowed *selection* still is, then flagged; and that the trigger applies no funding-source logic at all.

**Component** (`CardInput.dualBrandUI.test.tsx`) — allowed brand pre-selected; disallowed pick shows the error and keeps the selector; switching back clears it; no brand allowed shows the error with no selector; a rejected card tears down the previous card's selector; no configuration means no change in behaviour.

**E2E** (`dualBranding.fundingSource.spec.ts`) — one browser test for what jsdom cannot show, since CSF is mocked there: the reordered brand actually reaches SecuredFields, encryption stops when the disallowed brand is picked, and resumes on switching back.

**Regression** — healthcare, `issuingCountryCode`, `paymentMethodVariants` and detected-but-unsupported brands all still propagate.

---

**Closes**: COPER-477

---


packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
```diff
@@ -119,6 +127,36 @@ export const triggerBinLookUp = (element: CardElement | CustomCardElement) => {
...
+                             * the allowed funding sources, reject the card as unsupported. When `allowedFundingSources` is absent or
+                             * empty, or when binLookup resolves no funding source, no validation is performed.
+                             */
+                            const allowedFundingSources = 'allowedFundingSources' in element.props ? element.props.allowedFundingSources : undefined;
+
+                            if (allowedFundingSources?.length && hasDisallowedFundingSource(mappedResponse.supportedBrands, allowedFundingSources)) {
+                                const errObj: CardErrorData = {
+                                    type: 'card',
+                                    fieldType: 'encryptedCardNumber',
+                                    error: SF_ErrorCodes.ERROR_MSG_UNSUPPORTED_CARD_ENTERED,
```
FabioRomeiro: The final error message is still TBD

FabioRomeiro: And on the new validation error we should send the allowedFundingSources value and the fundingSource returned from BinLookup

packages/lib/src/components/internal/SecuredFields/binLookup/triggerBinLookUp.ts
```diff
@@ -119,6 +127,36 @@ export const triggerBinLookUp = (element: CardElement | CustomCardElement) => {
...
+                                    type: callbackObj.type,
+                                    detectedBrands: mappedResponse.detectedBrands,
+                                    supportedBrands: null,
+                                    paymentMethodVariants: mappedResponse.paymentMethodVariants,
+                                    ...(hasHealthcareData && { healthcare: mappedResponse.healthcare }),
+                                    brands: element.props.brands || DEFAULT_CARD_GROUP_TYPES
+                                });
+
+                                return;
+                            }
```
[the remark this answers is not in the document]
FabioRomeiro: That is a good point, it doesn't make sense to show the option of selecting a brand that has a disallowed funding source. The `onBinLookup` data should stay the same though.

PR #4097, opened 2026-07-22
 PayPal v6 (Task 2) — Button Rendering & Payment Flow

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

Builds the PayPal v6 rendering layer on top of the core SDK infrastructure. This adds the actual buttons shoppers interact with, the logic to start payment sessions, and controls for which buttons show and how the PayPal flow is presented.

## What's included

**Button components**
- New `PayPalButton`, `PayPalCreditButton`, `PayPalPayLaterButton`, and `VenmoButton` v6 web components.
- New `PayPalComponentV6` that renders the buttons together and shows a loading spinner (`PayPalSpinner`) until the SDK is ready.
- Buttons only render when the funding source is eligible, and can be individually hidden via `blockPayPalCreditButton`, `blockPayPalPayLaterButton`, and `blockPayPalVenmoButton`.

**Payment session logic (hooks)**
- `usePayPalOneTimeSession` / `usePayPalSaveSession` — start a one-time payment or a save (zero-auth) session on click.
- `useCreateOrder` / `useCreateVaultSetupToken` — create the order or vault setup token from the merchant's `onSubmit`.
- `usePayPalSessionOptions` and `usePayPalButtonEligibility` — build session options and check button eligibility.

**Configuration**
- New `presentationModeOptions` prop to customize how the PayPal payment UI is presented, defaulting to `{ presentationMode: 'auto' }`.
- Support for per-button `style` (PayPal and Venmo).

**Other**
- Added tests for all new components and hooks, plus Storybook stories.
- Fixed strict TypeScript issues across the PayPal component types.

## 🧪 Tested scenarios

- [x] `yarn test` passes for PayPal components and hooks
- [ ] Verify buttons render, respect eligibility/block flags, and start the correct session (one-time vs. save) on click

---

- COSDK-1318

**Closes**:  

---


packages/lib/src/components/PayPal/components/PayPalProcessingSpinner.tsx
```diff
@@ -0,0 +1,16 @@
...
+import Spinner from '../../internal/Spinner';
+
+export const PayPalProcessingSpinner = ({ withReviewPage }: Readonly<{ withReviewPage: boolean }>) => {
+    const { i18n } = useCoreContext();
+
+    return (
+        <div className="adyen-checkout__paypal">
+            <div className="adyen-checkout__paypal__status adyen-checkout__paypal__status--processing">
+                <Spinner size="medium" inline />
+                {withReviewPage && i18n.get('paypal.processingPayment')}
```
ribeiroguilherme: I think the logic here is inverted: the message should be displayed if there is no review page.

packages/lib/src/components/ApplePay/components/ApplePayButton.scss
```diff
@@ -1,5 +1,7 @@
+@use 'styles/variable-generator';
+
 apple-pay-button {
   --apple-pay-button-width: 100%;
-  --apple-pay-button-height: 48px;
-  --apple-pay-button-border-radius: 4px;
+  --apple-pay-button-height: variable-generator.token(spacer-120);
+  --apple-pay-button-border-radius: variable-generator.token(border-radius-s);
```
ribeiroguilherme: Apple Pay button height is broken

packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -225,6 +239,32 @@ class PaypalElement extends UIElement<PayPalConfiguration> {
...
+                data: {
+                    details: {
+                        orderID: orderId,
+                        payerID: payerId,
+                        ...restData
+                    },
+                    paymentData: this.paymentData ?? undefined
+                }
+            };
+        }
```
ribeiroguilherme, line 295: What is the use-case for this to have two different state data objects for the /details call? 
Perhaps adding a comment might help to clarify that

packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -225,6 +239,32 @@ class PaypalElement extends UIElement<PayPalConfiguration> {
...
+
+        this.handleAdditionalDetails(state);
+        return Promise.resolve();
```
ribeiroguilherme: Is this `handleOnApproveV6` async? Seems like it can be just marked as `void` and there is no need to return a Promise

packages/lib/src/components/PayPal/types.ts
```diff
@@ -213,6 +229,91 @@ export interface PayPalConfiguration extends UIElementProps {
...
+         */
+        /**
+         * Callback called when PayPal authorizes the payment.
+         * Must be resolved/rejected with the action object. If resolved, the additional details will be invoked. Otherwise it will be skipped
+         *
+         * @param data - Contains the raw event from PayPal, along with the billingAddress and deliveryAddress parsed by Adyen based on the raw event data
+         * @param actions - Used to indicate that payment flow must continue or must stop
+         */
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        onAuthorized?: (data: any, actions: { resolve: () => void; reject: () => void }) => void;
```
ribeiroguilherme: Is there a way to the the `data` here?

packages/lib/src/components/PayPal/types.ts
```diff
@@ -228,3 +329,26 @@ export type SupportedPayPalFundingSources = 'paypal' | 'credit' | 'paylater' | '
...
+    type?: PayPalButtonType;
+    class?: VenmoButtonClass;
+};
+
+export type PayPalPresentationModeOptions =
+    | PayPalPresentationModeOptionsForPopup
+    | PayPalPresentationModeOptionsForModal
+    | PayPalPresentationModeOptionsForRedirect
+    | PayPalPresentationModeOptionsForPaymentHandler
+    | PayPalPresentationModeOptionsForAuto;
```
ribeiroguilherme, line 377: Is it needed to export/expose all these types? 

Feels like there is no need for that. They will be used for auto-completion but there is no point in actually letting merchant import these.

packages/lib/src/components/PayPal/stories/PayPal.stories.tsx
```diff
@@ -31,3 +31,28 @@ export const Default: Story = {
...
+                        class: 'paypal-blue'
+                    },
+                    venmo: {
+                        type: 'pay',
+                        class: 'venmo-black'
+                    }
+                },
+                vault: false
+            }
+        }
```
ribeiroguilherme, line 60: There is a noticeable difference between the loading times for the old PayPal and PayPal v6  👀  V6 seems to take a while in order to verify the eligible methods

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -74,13 +99,18 @@ class PayPalService {
         const createInstance = paypal?.v6?.createInstance || paypal?.createInstance;
 
         if (!createInstance) {
-            return Promise.reject(new AdyenCheckoutError('ERROR', 'PayPal SDK `createInstance` is not available'));
+            throw new AdyenCheckoutError('ERROR', 'PayPal SDK `createInstance` is not available');
         }
 
+        const isTestEnvironment = this.environment?.toLowerCase() === 'test';
```
ribeiroguilherme: Better to check that environment does not include `live`  rather than comparing it to `test`.

I imagine that if we want to run this against a beta environment for example, then it will be considered live.

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -74,13 +99,18 @@ class PayPalService {
...
+
         this.sdkInstance = await createInstance({
             clientToken,
-            components: ['paypal-payments', 'venmo-payments'] satisfies PayPalComponents,
-            pageType: 'checkout'
+            components: ['paypal-payments', 'venmo-payments', 'paypal-messages'] satisfies PayPalComponents,
```
ribeiroguilherme: Do we always pass all the available components, or do we intend to apply some filtering? (e.g. pass `paypal-messages` only if the messaging callback is available).

I am assuming that each Component might increase the loading time of the SDK since more resources are needed.

packages/lib/src/components/PayPal/services/PayPalSdkLoader.ts
```diff
@@ -18,7 +18,7 @@ class PayPalSdkLoader {
     public async load(): Promise<typeof window.paypal> {
         try {
             const scriptElement = new Script({
-                src: this.environment.toLowerCase() === 'test' ? PAYPAL_SDK_URL_SANDBOX : PAYPAL_SDK_URL_PRODUCTION,
+                src: this.environment?.toLowerCase() === 'test' ? PAYPAL_SDK_URL_SANDBOX : PAYPAL_SDK_URL_PRODUCTION,
```
ribeiroguilherme: Same as in my other comment - Perhaps we can change this to use the URL sandbox if the environment does not include 'live'.

packages/lib/src/components/PayPal/Paypal.tsx
```diff
@@ -289,11 +329,63 @@ class PaypalElement extends UIElement<PayPalConfiguration> {
...
+     *
+     * @param data - The shipping options change data
+     */
+    private handleOnShippingOptionsChangeV6(data: PayPalV6OnShippingOptionsChangeData): Promise<void> {
+        const { usePayPalV6 } = this.props;
+
+        if (!usePayPalV6?.onShippingOptionsChange) return Promise.resolve();
+
+        return usePayPalV6.onShippingOptionsChange(data, this);
+    }
```
ribeiroguilherme, line 441: These callbacks are unbound. When they are triggered from the Paypal component, the reference for the `prop` and `this` doesn't exist.

Might be good to also add a test case to cover this.

packages/lib/src/components/PayPal/types.ts
ribeiroguilherme: `onAuthorized` does not seem to be triggered when Payment is authorized. Is this expected?

packages/lib/src/components/PayPal/hooks/usePayPalSessionOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const usePayPalSessionOptions = ({
+    onApprove,
+    onShippingAddressChange,
+    onShippingOptionsChange,
+    onCancel,
+    onError,
+    commit,
+    vault
+}: Omit<PayPalComponentV6Props, 'onSubmit' | 'style' | 'setComponentRef' | 'paymentSessionOptions'>) => {
```
ribeiroguilherme: `paymentSessionOptions` does not exist on  `PayPalComponentV6Props`

packages/lib/src/components/PayPal/components/PayPalProcessingSpinner.tsx
johnayeni: Fixed the logic here

packages/lib/src/components/ApplePay/components/ApplePayButton.scss
johnayeni: I reverted this change for now and would do in a separate PR

packages/lib/src/components/PayPal/Paypal.tsx
johnayeni: I have added comments for this

packages/lib/src/components/PayPal/Paypal.tsx
johnayeni: This `handleOnApproveV6` is async. I just added the implementation to fetch the order details with the checkoutshopper API

packages/lib/src/components/PayPal/types.ts
johnayeni: This has been implemented now

packages/lib/src/components/PayPal/types.ts
johnayeni: I marked these types as `@internal`. Do you think this is enough? I exported these types still because I am using them in other files

packages/lib/src/components/PayPal/stories/PayPal.stories.tsx
johnayeni: The delay actually comes from the fact that the `isAvailable` check is now asynchronous and we do this check in the `ComponentContainer` for storybook. I have updated the component to show a spinner until this check is done https://github.com/Adyen/adyen-web/blob/feature/paypal-sdk-v6-phase1/packages/lib/storybook/components/ComponentContainer.tsx#L49-L53

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Makes sense. Done

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Good point. I have implemented a change to only include the needed components based on the merchant configuration

packages/lib/src/components/PayPal/services/PayPalSdkLoader.ts
johnayeni: Makes sense. Done

packages/lib/src/components/PayPal/Paypal.tsx
johnayeni: Fixed

packages/lib/src/components/PayPal/types.ts
johnayeni: I just added the implementation to fetch the order details with the checkoutshopper API so this is now being triggered

packages/lib/src/components/PayPal/hooks/usePayPalSessionOptions.ts
johnayeni: Fixed

packages/lib/src/components/PayPal/types.ts
ribeiroguilherme: It is not a big ideal honestly. But if they don't need to be exposed externally, then they could be kept in the `components/types.ts` file and not be imported in this file. Marking as `@internal` currently does not do anything

packages/lib/src/components/PayPal/services/PayPalService.ts
```diff
@@ -74,22 +100,31 @@ class PayPalService {
         const createInstance = paypal?.v6?.createInstance || paypal?.createInstance;
 
         if (!createInstance) {
-            return Promise.reject(new AdyenCheckoutError('ERROR', 'PayPal SDK `createInstance` is not available'));
+            throw new AdyenCheckoutError('ERROR', 'PayPal SDK `createInstance` is not available');
         }
 
+        console.log('PayPal SDK component', this.components);
```
ribeiroguilherme, line 106: Some console.logs left , but not big issue since it is not final branch

ribeiroguilherme (approved):
👏

packages/lib/src/components/PayPal/services/PayPalService.ts
johnayeni: Will remove this in a subsequent PR

PR #4098, opened 2026-07-22
fix(a11y): announce loading spinner to screen readers for QR/redirect methods and secured fields

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

The loading spinner shown while waiting for a payment status check (Swish, WeChat Pay, PromptPay, Pix, PayNow, PayMe, DuitNow, BCMC Mobile) and the spinner shown while Card / Stored Card / MB WAY / ANCV secured fields load had no accessible name, so screen readers announced nothing while it was visible. Reported by a merchant testing Swish.

- Extracted the QR loader's pending state into `QRLoaderPendingState`. `QRLoader` reports `loading`/`loaded` to the shared SRPanel, since the pending state unmounts before it could announce that loading finished.
- `LoadingWrapper` reports the same via the new `useLoadingA11yReporter` hook, fixing Card, Stored Card, MB WAY and ANCV in one place with no changes needed at their call sites.
- Made `useA11yReporter` defensive against a missing `SRPanelProvider` (some components/tests render without one); it now no-ops instead of throwing.

The announcements were still being dropped, because the SRPanel is a singleton shared by loading messages and form validation errors, and both cleared it unconditionally. A Card re-render with no validation errors wiped a `Loading…` message written ~10ms earlier, well before a screen reader could read it.

- Error reporters only clear the panel when they have errors of their own on display (`hasDisplayedErrors`).
- Status reporters no longer clear on unmount; a message stays until another replaces it. This is why a terminal `loaded` message is required rather than cosmetic.
- Re-writing the message already on display remounts it, so an identical message is announced again instead of diffing to nothing.

Adds one translation key, `loaded`.

## 🧪 Tested scenarios

- `SRPanel.ownership.test.tsx` — 8 cases pinning down when the shared panel may and may not be cleared, asserted against the DOM mutations the live region emits.
- `useLoadingA11yReporter.test.tsx` — loading, terminal, and result-message behaviour.
- `QRLoader.test.tsx` — loading message is superseded once the QR code shows, and the payment result wins over the generic `loaded` message on completion.
- `LoadingWrapper.test.tsx` covering both loading and non-loading states.
- Verified manually with VoiceOver on the Drop-in story.
- Full `yarn jest` (270 suites / 2085 tests), `yarn lint`, `tsc --noEmit` and `yarn build` pass.

---

Reported via Slack thread (Chanel, Swish loader not announced to screen readers).

**Closes**: COSDK-1212


packages/lib/src/components/internal/LoadingWrapper/LoadingWrapper.tsx
```diff
@@ -9,6 +11,11 @@ interface LoadingWrapperProps {
 }
 
 const LoadingWrapper = ({ children, status }: Readonly<LoadingWrapperProps>) => {
+    const { i18n } = useCoreContext();
+
+    // Announce the loading status to screen readers, since the spinner itself has no accessible name
+    useA11yReporter(status === 'loading' ? i18n.get('loading') : null);
+
```
ribeiroguilherme, line 15: FYI On Drop-in it is not reading it properly for every payment method that has Secured fields. It looks like it reads only for the first one or a single time.

ribeiroguilherme (dismissed):

ribeiroguilherme (dismissed):

ribeiroguilherme (approved):

PR #4109, opened 2026-07-27
Fixed: focus moving to the wrong component when multiple OpenInvoice components

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

On OpenInvoice-based components (AfterPay, RatePay, FacilyPay, Atome, ...), pressing Pay with an invalid form moves focus to the first field in error. This focus call used a global CSS selector (`.adyen-checkout__open-invoice`) resolved via `document.querySelector`, which always returns the **first** matching element in the DOM.

When multiple OpenInvoice-based components are rendered on the same page, submitting any instance other than the first one moved focus into the **first** component's fields instead of the submitted component's own fields.

Fixed by scoping the focus lookup to the submitted component's own container, using a `containerRef` (the same pattern already used by `IssuerList` and the ClickToPay inputs) instead of a page-wide selector.

---

## 🧪 Tested scenarios

- Added a unit test in `AfterPay.test.ts` that renders two independent `AfterPay` instances on the same page, submits the second one with an empty form, and asserts focus lands on the second instance's field (not the first's).
- Verified the new test fails against the old code (reproducing the bug) and passes with the fix.
- Manually reproduced with a temporary debug Storybook story rendering two `AfterPay` components side by side (removed before opening this PR).

---

**Closes**:   COSDK-1417

---

### DIFF
--- a/.changeset/fix-openinvoice-focus-multiple-instances.md
+++ b/.changeset/fix-openinvoice-focus-multiple-instances.md
@@ -1,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed: focus moving to the wrong component when multiple OpenInvoice-based components (e.g. AfterPay, RatePay, FacilyPay, Atome) are rendered on the same page and a validation error occurs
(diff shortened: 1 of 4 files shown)


ScottiBR (approved):
LGTM!

Issue #4076, opened 2026-07-10
US postal code: error message expects ZIP+4 but regex only checks /\d{5}/


sun-jingtao:
## Describe the bug

For `country = US`, the postal code **error message** and **validation regex** do not match.

- Error message (from `validate.formats.ts`): `99999 or 99999-9999`
- Actual regex (from `validate.ts`): `US: createPatternByDigits(5)` → `/\d{5}/` (unanchored)

So any value that **contains** 5 consecutive digits passes validation, even with extra junk characters.

## Steps to reproduce

1. Use Card Component with:
   ```js
   {
     billingAddressRequired: true,
     billingAddressMode: 'partial',
     data: { billingAddress: { country: 'US' } }
   }

sponglord:
@sun-jingtao This issue should now be fixed with our release `v6.42.0`

PR #4111, opened 2026-07-28
fix: US postal code validation

dimitra-adyen:
## 📋 Pull Request Checklist

---

## 📝 Summary

**Align US postal code validation with the error message.**

**Issue:** 
`validate.formats.ts` states that the US postal code format is `99999 or 99999-9999`, but `validate.ts` uses
`US: createPatternByDigits(5)`, which builds `new RegExp('\d{5}')`. This means, that any value
containing five consecutive digits passes, so `1234567` and values with surrounding non-digit
characters are accepted despite the error message stating otherwise.

**Solution:**
Replace the `createPatternByDigits` helper for `US` with an explicit pattern that
matches what the error message actually advertises:
 
```diff
- US: createPatternByDigits(5)
+ US: { pattern: /^\d{5}(?:[-\s]\d{4})?$/ }
```

**Enable the address formatter for AVS**

**Issue**
- The formatter was not applied to the Address fields for AVS, since no country was inside the form. 
**Solution**
- Add an optional `fallbackCountry` parameter to `addressFormatters`, and pass the `props.data.country` inside the `Address.tsx`, enabling the formatter for partial address as well.
---

## 🧪 Tested scenarios

---

**Closes**:  #4076 

---

### DIFF
--- a/packages/lib/src/components/internal/Address/Address.tsx
+++ b/packages/lib/src/components/internal/Address/Address.tsx
@@ -9,7 +9,7 @@ import { AddressData } from '../../../types/global-types';
 import FieldContainer from './components/FieldContainer';
 import useForm from '../../../utils/useForm';
 import Specifications from './Specifications';
-import { ADDRESS_SCHEMA, FALLBACK_VALUE } from './constants';
+import { ADDRESS_SCHEMA, COUNTRY, FALLBACK_VALUE } from './constants';
 import { getMaxLengthByFieldAndCountry } from '../../../utils/validator-utils';
 import { useCoreContext } from '../../../core/Context/CoreProvider';
 import AddressSearch from './components/AddressSearch';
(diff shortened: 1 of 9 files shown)


ribeiroguilherme (commented):
There is a discrepancy between the 'Partial AVS' and 'AVS' vailidation. 
Let's take a look into that and try to align both mechanisms

packages/lib/src/components/internal/Address/validate.formats.ts
```diff
@@ -17,10 +18,10 @@ const createFormatByDigits = (digits: number): Formatter => {
 const specialCharsRegEx = getFormattingRegEx(SPECIAL_CHARS);
 const formattingFn = val => trimValWithOneSpace(val).replace(specialCharsRegEx, '');
 
-export const addressFormatters: FormatRules = {
+export const addressFormatters = (fallbackCountry?: string): FormatRules => ({
```
m1aw: Why do we need to handle fallback country here?

dimitra-adyen: Since for AVS there's no country field in the form, data.state.country is empty. This was the reason why the formatters aren't working there

dimitra-adyen: Addressed the issue inside the `Address` component.

packages/lib/src/components/internal/Address/Address.tsx
```diff
@@ -145,19 +148,14 @@ export default function Address(props: Readonly<AddressProps>) {
             const isOptional = optionalFields.includes(cur);
             const isRequired = requiredFields.includes(cur);
             const newValue = data[cur];
-            const initialValue = props.data[cur];
+            const initialValue = defaultData[cur];
             // recover default data values which are not requiredFields, or prefill with 'N/A'
-            const fallbackValue = !isRequired && !newValue && !!initialValue ? initialValue : FALLBACK_VALUE;
+            const fallbackValue = !isRequired && initialValue ? initialValue : FALLBACK_VALUE;
```
dimitra-adyen, line 153: @m1aw  Could you verify if this change is valid?

m1aw: If I follow through the logic I think this makes sense.
Then it gets evaluated if there's or not `newValue` on the next line anyway. 

So I think this makes sense?

m1aw (approved):

PR #4112, opened 2026-07-29
Select component - Adjust secondary text position and add Tag component

ScottiBR:
## 📋 Pull Request Checklist

---

### secondaryText moved below the option name

The **collapsed button** now renders tags only, and never `secondaryText`. Both competed for the same narrow strip and only one could win, so rather than adding a config prop to pick between them we made it unconditional — tags are the thing that needs to survive collapse. If a consumer needs to convey state on the closed button, it should do that through `disabled` or the item name, not `secondaryText`.

### Tags for EMI

Emi requires a new Tag component to display extra information about EMI plans, it's possible to have multiple tags in the same select row.

image

### Smaller fixes picked up along the way

- The selected-option checkmark was `position: absolute; right: spacer-040` inside the option, which overlapped content and broke in RTL. It's now `flex-shrink: 0` and flows naturally in the row.
- Filterable click handling: clicking the input of an already-open list used to close it. `handleClick` now returns early in that case.
- The screen-reader status region had been borrowing `adyen-checkout-sr-panel--sr-only`; it now has its own `__dropdown__status` class using the `a11y-visually-hidden` mixin.
- `readonly` is destructured in `SelectButtonElement` so it stops being spread onto the filterable wrapper `<div>` as an invalid DOM attribute.
- Option padding is `spacer-060` and `__element__text` is `text-subtitle-font-size` — these are the values design specified, the previous ones were the defect.
- Truncate primary text for smaller screens

---

## 🧪 Tested scenarios

- New `Select.tags.test.tsx`
- One test explicitly encodes the collapsed-button decision: `never renders the secondaryText of the selected option`. It inverts what the previous test asserted, so the behaviour can't regress silently.
- New `Tag.test.tsx` and `TagList.test.tsx` cover rendering, ordering, the null/empty guard, and that neither adds list semantics.
- New Click to Pay regression test: `should not select an expired card when its option is clicked` — clicks the expired option and asserts the selection and Pay button label don't change. This is what makes the `disabled` compensation real rather than incidental. Also updated the existing all-expired assertion from `'Mastercard •••• 3456 Expired'` to `'Mastercard •••• 3456 '` to match the new collapsed button.

---

**Closes**:  

---


packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
```diff
@@ -35,6 +36,23 @@ function SelectButtonElement({ filterable, toggleButtonRef, showList, selectList
...
+                <div className="adyen-checkout__dropdown__button__tags">
+                    {selected.tag.map(tag => (
+                        <Tag key={tag.label} label={tag.label} variant={tag.variant} />
+                    ))}
+                </div>
+            );
+        case SecondaryContentVariant.SECONDARY_TEXT:
+            if (!selected.secondaryText) return null;
+            return <span className="adyen-checkout__dropdown__button__secondary-text">{selected.secondaryText}</span>;
+    }
```
ScottiBR: we can do an early return for both before the switch:
`if(!selected[variant].length) return null`

packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
```diff
@@ -35,6 +36,23 @@ function SelectButtonElement({ filterable, toggleButtonRef, showList, selectList
...
+function SecondaryContent({ selected, variant }: Readonly<{ selected: SelectItem; variant: SecondaryContentVariant }>) {
+    switch (variant) {
+        case SecondaryContentVariant.TAG:
+            if (!selected.tag?.length) return null;
+            return (
+                <div className="adyen-checkout__dropdown__button__tags">
+                    {selected.tag.map(tag => (
+                        <Tag key={tag.label} label={tag.label} variant={tag.variant} />
+                    ))}
+                </div>
```
ScottiBR: maybe we should create a TagList component and unify the css from `adyen-checkout__dropdown__button__tags` and `adyen-checkout__dropdown__element__tags`

packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
dimitra-adyen: Added! ✅

packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
dimitra-adyen: Created the `Taglist` component, with the classname as parameter since there were small differences between the `SelectListItem` and the `SelectButton` ✅

packages/lib/src/components/internal/FormFields/Select/components/SelectButton.tsx
```diff
@@ -53,19 +54,26 @@ function SelectButton(props: Readonly<SelectButtonProps>) {
...
 
-    const setFocus = (e: Event) => {
+    const handleClick = (e: Event) => {
         e.preventDefault();
-        if (document.activeElement === props.filterInputRef.current) {
-            if (!props.showList) {
-                props.toggleList(e);
-            }
-        } else if (props.filterInputRef.current) props.filterInputRef.current.focus();
+        props.filterInputRef.current?.focus();
```
m1aw, line 59: Good clean up.

m1aw (approved):

PR #4119, opened 2026-08-02
fix: improve Japanese card and address translations

m1aw:
## 📋 Pull Request Checklist

---

## 📝 Summary

Improve Japanese localization for Checkout:
- Updated translations for Card and Address
- Added a dedicated JP address form (country, postal code/prefecture, city, street, building name/room number), with Prefecture as a free-text field

---

## 🧪 Tested scenarios

Unit tests added for the new JP address schema and free-text field behavior.

---

**Closes**: COSDK-1213


m1aw:
@ScottiBR adding you back as a reviewer (removed it because to put it back in Draft).

Code is ready to review, the tests are currently not passing because I'm going to remove the Japan story card story. Right now is only there to use the mocks so the Localization team can validate.

.changeset/warm-plums-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+'@adyen/adyen-web': patch
+---
+
+Fixed: improved Japanese translations for Card and updated the Address component for JP with a localized field set (country, postal code, prefecture, city, street, building name/room number)
```
ScottiBR, line 5: This looks like a breaking change rather than a minor fix.

Before this PR, JP defaulted to auto-filling stateOrProvince with 'N/A' without rendering StateField. Now, it prompts the user, marks it required, and submits the prefecture ID ('13'). That breaks contract for anyone already collecting JP billing addresses since both the UI and payment payload will change.

packages/lib/src/components/internal/Address/validate.ts
```diff
@@ -55,6 +55,7 @@ const postalCodePatterns = {
     IE: { pattern: /(?:^[AC-FHKNPRTV-Y][0-9]{2}|D6W)[ -]?[0-9AC-FHKNPRTV-Y]{4}/ },
     IS: createPatternByDigits(3),
     IT: createPatternByDigits(5),
+    JP: { pattern: /^\d{3}-?\d{4}$/ },
```
ScottiBR, line 58: Japanese keyboards usually give you full-width digits pretty often (`１０７−００５２`). should we add a small JP `formatterFn` that maps full-width digits to ascii and strips any JP character like `〒`, similar of what we have in BR? that way we keep the stricter validation without breaking the common local input.

packages/lib/src/components/Card/stories/CardWithAVSJapan.stories.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const meta: MetaConfiguration<CardConfiguration> = {
+    title: 'Components/Cards'
+};
+
+/**
+ * Mocks the `datasets/states/JP/:locale.json` endpoint so the AVS Prefecture dropdown can be
+ * demoed before the dataset files are hosted by checkoutshopper.
+ * The data below mirrors what will be served in production (47 JP prefectures).
+ */
```
ScottiBR, line 17: Since we have the endpoint, why not use it at this story? would it be better for testing and catching problems?

packages/lib/src/components/internal/Address/Address.test.tsx
```diff
@@ -263,6 +263,55 @@ describe('Address', () => {
         expect(receivedData.stateOrProvince).toBe(undefined);
     });
 
+    describe('JP address specification', () => {
+        const requiredFields = ['country', 'postalCode', 'stateOrProvince', 'city', 'street', 'houseNumberOrName'];
+
+        test('should render the fields in the expected order: country, postalCode/prefecture, city, street, building', async () => {
```
ScottiBR, line 269: the name says order but every assertion is `toBeInTheDocument`, so this would still pass if the JP schema got reordered back to the default layout

ScottiBR (changes requested):

.changeset/warm-plums-jog.md
m1aw: This is correct, I changes this to a minor, the scope grew a bit.

However indeed we are sending a new value for `stateOrProvince`, this could break some integrations yes, but they would need to be making a specific check on `stateOrProvince` being ONLY `N/A`. For a merchant on advanced flow just forwarding the `billingAddress` field or processing our address field it should be unfected.

packages/lib/src/components/Card/stories/CardWithAVSJapan.stories.tsx
m1aw: As mentioned on the comment above, this was for the local team to validate. We have re-ordered the dataset on the backend so we are waiting for this to be updated.

packages/lib/src/components/internal/Address/Address.test.tsx
m1aw: Oh yeah, great catch I fill fix this.

packages/lib/src/components/internal/Address/validate.ts
m1aw: Fixed.

packages/lib/src/components/Card/stories/CardWithAVSJapan.stories.tsx
m1aw: Removed

ScottiBR (approved):
LGTM

PR #4120, opened 2026-08-03
feat: support custom display mode for payment methods in Drop-in

FabioRomeiro:
## 📋 Pull Request Checklist

---

## 📝 Summary

Today the Drop-in instant payments section is driven **only** by the merchant's `instantPaymentTypes` configuration. There is no way to influence it from the payment methods response, so promoting a payment method into that section requires every merchant to change their integration.

This PR adds a **custom display mode**: an optional `displayMode` property inside each payment method's `configuration` object in the `/paymentMethods` response, which decides the section the method renders in.

### Behaviour

| Response | Result |
|---|---|
| `configuration.displayMode: 'instant'` | renders in the instant payments section |
| `configuration.displayMode: 'regular'` | renders in the regular payment method list |
| **any** payment method carries a `displayMode` | the custom display mode owns the decision for the whole list and `instantPaymentTypes` is **ignored** |
| no payment method carries a `displayMode` | unchanged, `instantPaymentTypes` still applies |

Additional rules:

- The existing type allowlist still applies. A payment method outside `SUPPORTED_INSTANT_PAYMENTS` (`paywithgoogle`, `googlepay`, `applepay`) never enters the instant section, even when marked `instant`.
- Unrecognised values (e.g. a future `promoted`) resolve to `regular`. Note they still activate the custom display mode, so `instantPaymentTypes` is ignored in that case too.
- `fastlane` continues to be sectioned by payment method type and is unaffected by `displayMode`.

### Implementation notes

- All the logic lives in a single seam, `splitPaymentMethods`, which both the rendering call site (`Dropin.handleCreate`) and the analytics call site (`DropinComponent.reportPaymentMethodList`) already funnel through. Rendering and the `paymentListDisplayed` event therefore cannot drift.
- Each payment method is now sectioned exactly once, in a single grouping pass extracted into `groupByDisplayMode`, then read back per display mode. The previous code made several passes over the list and evaluated `isInstantPaymentMethod` twice per payment method.
- Introduced `DisplayMode` in a new `Dropin/constants.ts` as the single definition of the display mode vocabulary, and derived `PaymentMethodDisplayMode` from it. This vocabulary was already duplicated as string literals between the sectioning logic and the analytics event.
- A `const` object was chosen over a TS `enum` deliberately: string literals are not assignable to string enums, which would have forced churn across existing `displayMode: 'regular'` analytics fixtures.
- `ResolvedDisplayMode` (`Exclude<PaymentMethodDisplayMode, 'stored'>`) documents and enforces that `stored` is never a sectioning outcome. It describes the separate `storedPaymentMethods` array and exists only so the analytics event can label those entries.
- `RawPaymentMethod.configuration` is narrowed from `object` to `Record<string, string>`, mirroring the backend's `Map<String, String>`. This is what makes `configuration.displayMode` readable without a cast.

No changes to what is sent in analytics: only the string literals were replaced with the shared constants, and the `toAnalyticsPaymentMethod` allowlist is untouched.

---

## 🧪 Tested scenarios

**Unit tests** (`yarn jest src/components/Dropin` → 11 suites, 111 tests passing):

- `splitPaymentMethods.test.ts`, 13 tests, covering: `instant` honoured without `instantPaymentTypes`; `instantPaymentTypes` fully ignored once the response sets a `displayMode`; response sending only `regular`; non-allowlisted type marked `instant`; unknown values falling back to `regular`; empty/`null` `displayMode` not activating the custom mode; a single `displayMode` taking over the whole list; `fastlane` staying out of the instant section when marked `instant`; legacy `instantPaymentTypes` path when `configuration` has no `displayMode`; response order preserved within every section.
- `Dropin.test.ts`, 4 integration cases under *"display mode driven by the payment methods response"*, asserting the **rendered DOM** rather than the analytics payload:
  - a payment method the response marks `instant` renders in the instant payments area;
  - a payment method renders in the regular list when the response overrides `instantPaymentTypes`;
  - `instantPaymentTypes` still applies when the response carries no `displayMode`;
  - the response order is preserved within both the instant area and the regular list.

  These mount a real Drop-in through `AdyenCheckout` with a local `UIElement` subclass standing in for an instant-capable payment method, so the assertions verify what the shopper actually sees. Each was verified by mutation testing (forcing detection off, and swapping `push` for `unshift`).

**Storybook**: new `Drop-in/Drop-in Component → Custom Display Mode` story. It marks iDEAL and Apple Pay `regular` while setting `instantPaymentTypes: ['applepay']`, so it demonstrates the response overriding the merchant configuration (Apple Pay does *not* render in the instant section).

**Static checks**: `tsc --noEmit` and `eslint .` both clean.

---

**Closes**:  COPER-586

---

### DIFF
--- a/packages/lib/src/components/Dropin/elements/splitPaymentMethods.ts
+++ b/packages/lib/src/components/Dropin/elements/splitPaymentMethods.ts
@@ -15,7 +15,7 @@ interface SplitPaymentMethods {
  */
 type ResolvedDisplayMode = Exclude<PaymentMethodDisplayMode, typeof DisplayMode.stored>;
 
-const getRequestedDisplayMode = (paymentMethod: PaymentMethod): string | undefined => {
+const getCustomDisplayMode = (paymentMethod: PaymentMethod): string | undefined => {
     return paymentMethod.configuration?.displayMode;
 };
 
@@ -24,7 +24,7 @@ const getRequestedDisplayMode = (paymentMethod: PaymentMethod): string | undefin
  */
 const resolveDisplayMode = (
     paymentMethod: PaymentMethod,
-    hasRequestedDisplayMode: boolean,
+    hasCustomDisplayMode: boolean,
     instantPaymentTypes: InstantPaymentTypes[]
 ): ResolvedDisplayMode => {
     if (paymentMethod.type === 'fastlane') {
@@ -35,8 +35,8 @@ const resolveDisplayMode = (
         return DisplayMode.regular;
     }
 
-    const isInstant = hasRequestedDisplayMode
-        ? getRequestedDisplayMode(paymentMethod) === DisplayMode.instant
+    const isInstant = hasCustomDisplayMode
+        ? getCustomDisplayMode(paymentMethod) === DisplayMode.instant
         : instantPaymentTypes.includes(paymentMethod.type as InstantPaymentTypes);
 
     if (isInstant) {
@@ -53,7 +53,7 @@ const groupByDisplayMode = (
     paymentMethods: PaymentMethod[],
     instantPaymentTypes: InstantPaymentTypes[]
 ): Record<ResolvedDisplayMode, PaymentMethod[]> => {
-    const hasRequestedDisplayMode = paymentMethods.some(getRequestedDisplayMode);
+    const hasCustomDisplayMode = paymentMethods.some(getCustomDisplayMode);
 
     const groups: Record<ResolvedDisplayMode, PaymentMethod[]> = {
         [DisplayMode.fastlane]: [],
@@ -62,7 +62,7 @@ const groupByDisplayMode = (
     };
 
     paymentMethods.forEach(paymentMethod => {
-        const displayMode = resolveDisplayMode(paymentMethod, hasRequestedDisplayMode, instantPaymentTypes);
+        const displayMode = resolveDisplayMode(paymentMethod, hasCustomDisplayMode, instantPaymentTypes);
         groups[displayMode].push(paymentMethod);
     });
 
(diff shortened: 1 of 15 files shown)


packages/lib/src/components/Dropin/elements/splitPaymentMethods.ts
```diff
@@ -9,14 +10,87 @@ interface SplitPaymentMethods {
...
 
-function splitPaymentMethods(paymentMethods: PaymentMethods, instantPaymentTypes: InstantPaymentTypes[]): SplitPaymentMethods {
-    const isFastlane = ({ type }: PaymentMethod) => type === 'fastlane';
-    const isInstantPaymentMethod = ({ type }: PaymentMethod) => instantPaymentTypes.includes(type as InstantPaymentTypes);
+/**
+ * The display modes a payment method from the 'paymentMethods' array can resolve to.
+ */
+type ResolvedDisplayMode = Exclude<PaymentMethodDisplayMode, 'stored'>;
+
+const getCustomDisplayMode = (paymentMethod: PaymentMethod): string | undefined => {
```
johnayeni: Do we really need to have this utility or can we just simply check if the displayMode value is equal to the supported list of display modes?

```ts
[DisplayMode.instant, DisplayMode.regular].includes(paymentMethod.configuration.displayMode)
```

FabioRomeiro: Good point, I was being unnecessarily defensive there. I simplified the validation without explicitly covering edge cases, but I had to keep the utility for typing consistency. The gymnastics we have to do to keep typescript happy :)

johnayeni (approved):
LGTM

johnayeni (dismissed):
Requesting changes to prevent accidental merging for now

PR #4147, opened 2026-08-19
fix: resolve ts issues in internal components

dimitra-adyen:
## 📋 Pull Request Checklist

---

## 📝 Summary

Continues the strict TypeScript migration by removing `no-explicit-any` violations from the internal components, so these folders can be dropped from the ESLint ignore list in `packages/lib/eslint.config.js`:
`Address`, `ClickToPay`, `FormFields`, `IbanInput`, `IssuerList`, `OpenInvoice`.

**How:**

- **Address**: typed `reject` callbacks with `RejectionReason` and `onManualAddress`, narrowed `countryHasOptionalField` to `AddressField`, dropped the unused `ref?: any` prop.
- **ClickToPay**: introduced `ISchemeSdk` plus scheme-specific `IVisaSchemeSdk` / `IMastercardSchemeSdk` interfaces to replace `schemeSdk: any`; `children` typed as `ComponentChildren`.
- **Card**: `ClickToPayWrapper` now derives its props from the provider/holder types and spreads them, instead of re-declaring and forwarding each prop manually.
- **IbanInput / OpenInvoice / IssuerList / Field**: concrete types for state, `payButton`, `onChange` and render callbacks; removed dead `setData` / `setValid` and the unused `ibanNumber` ref.

Type-only changes — no runtime behaviour is intended to change.

---

## 🧪 Tested scenarios

- `yarn type-check` and `yarn lint` pass with the reduced ignore list.
- Manual smoke test of the touched flows: Card + Click to Pay, AVS address lookup, SEPA (IBAN input), issuer list and open invoice.

---

**Closes**:

COSDK-1069


packages/lib/src/components/internal/IssuerList/types.ts
```diff
@@ -9,7 +9,7 @@ export interface IssuerListProps {
     type: string;
     showPayButton: boolean;
     payButton(props: PayButtonProps): ComponentChildren;
-    onChange(payload: any): void;
+    onChange(payload: unknown): void;
```
ribeiroguilherme, line 12: Can that be more strictly typed?  [Perhaps something such as this](https://github.com/Adyen/adyen-web/blob/main/packages/lib/src/components/internal/PhoneInput/types.ts#L14) .

packages/lib/src/components/internal/IbanInput/IbanInput.tsx
```diff
@@ -100,18 +99,10 @@ class IbanInput extends Component<Readonly<IbanInputProps>, IbanInputState> {
         this.props.onChange(data);
     }
 
-    public setData = (key, value, cb?) => {
```
ribeiroguilherme: Were these methods (`setData` and `setValid`) not being used anywhere? 😱 
I tested SEPA and the payload seems to be ok and contain the IBAN value

packages/lib/src/components/internal/ClickToPay/services/sdks/MastercardSdk.ts
```diff
@@ -15,8 +16,17 @@ const IdentityTypeMap = {
...
+
+export interface IMastercardSchemeSdk extends ISchemeSdk {
+    init(params: MastercardSdkInitParams): Promise<void>;
+    identityLookup(params: { consumerIdentity: { identityValue: string; identityType: string } }): Promise<SrciIdentityLookupResponse>;
+    completeIdentityValidation(params: { validationData: string }): Promise<SrciCompleteIdentityValidationResponse>;
+}
+
 class MastercardSdk extends AbstractSrcInitiator {
     public readonly schemeName = 'mc';
+    declare public schemeSdk: IMastercardSchemeSdk;
```
ribeiroguilherme, line 29: question: Do we need to use `declare` here and also in the `VisaSdk` file? What does it fix/help in this case?

ribeiroguilherme (dismissed):

packages/lib/src/components/internal/ClickToPay/services/sdks/MastercardSdk.ts
dimitra-adyen: schemeSdk already exists in the base class, so we just want to narrow the type down to the scheme's interface (different for mc and visa), thus the use of declare.

johnayeni (approved):
LGTM

PR #4149, opened 2026-08-20
Fix: remove implicit and explicit any from BaseElement and UIElement

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary

Tightens the typing of the core `BaseElement`/`UIElement` hierarchy so the base classes no longer rely on implicit or explicit `any`, and removes both from the ESLint `no-explicit-any` exemption list.

### Key changes

- `BaseElement` is now generic over its state (`BaseElement<P, S>`), with `state`, `_component`, and `formatData()` properly typed instead of `any`. `IBaseElement` is likewise parameterised by props and state.
- `UIElement.componentRef` is typed as `ComponentMethodsRef`, and ref access is narrowed via optional chaining rather than truthiness checks.
- The unused second `props` argument was dropped from `setStatus()` / `setElementStatus()`, and the dead `resolveFinalResult()` helper was removed.
- `assertIsDropin()` now narrows with `in` checks against a typed `UIElement` instead of accepting `any`.
- Payment method components were updated to match the stricter base signatures, mostly by typing callback payloads and component refs.
- Components that previously assigned methods onto `this` from a function component (e.g. `RedirectButton`, `BacsInput`) now expose their imperative API through the standard `setComponentRef` + `useRef` pattern.
- `PaymentData.clientStateDataIndicator` is now optional, matching how it is actually populated.

### Notes

- `setStatus(status, props)` and `setElementStatus(status, props)` silently ignored the second argument, so dropping it is behaviour-preserving for JS callers, but it is a signature change for TypeScript users passing it. Released as a patch.
- Unit tests for Bacs were updated alongside the ref refactor; the full suite passes

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
+++ b/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
@@ -1,11 +1,12 @@
 import { h } from 'preact';
-import { useState } from 'preact/hooks';
+import { useEffect, useRef, useState } from 'preact/hooks';
 import { useCoreContext } from '../../../../core/Context/CoreProvider';
 import { MBWayInputProps } from './types';
 import './MBWayInput.scss';
 import PhoneInputForm from '../../../internal/PhoneInput';
 import LoadingWrapper from '../../../internal/LoadingWrapper';
 import usePhonePrefixes from '../../../internal/PhoneInput/usePhonePrefixes';
+import { ComponentMethodsRef } from '../../../types';
 
 function MBWayInput(props: Readonly<MBWayInputProps>) {
     const { i18n, loadingContext } = useCoreContext();
@@ -15,6 +16,14 @@ function MBWayInput(props: Readonly<MBWayInputProps>) {
     const [status, setStatus] = useState<string>('ready');
 
     this.setStatus = setStatus;
+
+    const mbWayInputRef = useRef<ComponentMethodsRef>({
+        setStatus: setStatus
+    });
+
+    useEffect(() => {
+        props.setComponentRef(mbWayInputRef.current);
+    }, [props.setComponentRef]);
 
     const { loadingStatus: prefixLoadingStatus, phonePrefixes } = usePhonePrefixes({ allowedCountries, loadingContext, handleError: props.onError });
 
--- a/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
+++ b/packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
@@ -15,8 +15,6 @@ function MBWayInput(props: Readonly<MBWayInputProps>) {
 
     const [status, setStatus] = useState<string>('ready');
 
-    this.setStatus = setStatus;
-
     const mbWayInputRef = useRef<ComponentMethodsRef>({
         setStatus: setStatus
     });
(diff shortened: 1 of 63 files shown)


packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -16,6 +17,14 @@ function MBWayInput(props: Readonly<MBWayInputProps>) {
 
     this.setStatus = setStatus;
```
m1aw: I think this probably should be cleaned up.
I also found a bug on MBWay when you press it for the first time it doesn't trigger validation.

packages/lib/src/components/internal/UIElement/UIElement.tsx
```diff
@@ -220,18 +220,16 @@ export abstract class UIElement<P extends UIElementProps = UIElementProps> exten
      * - If Drop-in, will set status for Dropin component, and then it will propagate the new status for the active payment method component
      * - If Component, it will set its own status
      */
-    public setElementStatus(status: UIElementStatus, props?: any): this {
-        this.elementRef?.setStatus(status, props);
+    public setElementStatus(status: UIElementStatus): this {
+        this.elementRef?.setStatus?.(status);
```
m1aw, line 225: Any idea what was props used for here? As in I see it's mentioned in the PR description, but I just think it wararnts some discussion.

packages/lib/src/types/global-types.ts
```diff
@@ -290,7 +290,7 @@ export interface PaymentData extends PaymentMethodData {
         orderData: string;
         pspReference: string;
     };
-    clientStateDataIndicator: boolean;
+    clientStateDataIndicator?: boolean;
```
m1aw, line 294: This is always true.

packages/lib/src/components/Sepa/Sepa.tsx
```diff
@@ -49,14 +49,12 @@ class SepaElement extends UIElement<SepaConfiguration> {
             <Fragment>
                 <FormInstruction />
 
-                {/* @ts-ignore TODO: add props */}
                 <IbanInput
                     ref={ref => {
```
m1aw, line 53: Why wasn't this one changed?

m1aw (commented):
I have not finished reviewing, but I noticed this so far.

packages/lib/src/components/Sepa/Sepa.tsx
johnayeni: The IbanInput component is still a class component

packages/lib/src/types/global-types.ts
johnayeni: In theory it should, but not in practice. UIElement for example does not set this value when formating the data object, leaving it to extended components to set it

johnayeni: Do you mean that this value should always be `true`?

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
johnayeni: fixed

packages/lib/src/components/internal/UIElement/UIElement.tsx
johnayeni: I dug into this a bit more and it seems this was because of Dropin. Dropin overrides `setStatus` and actually uses the props parameter. What I have done now is to also override `setElementStatus ` which would support the `props` param. Let me know what you think

packages/lib/src/components/MBWay/components/MBWayInput/MBWayInput.tsx
```diff
@@ -25,7 +32,18 @@ function MBWayInput(props: Readonly<MBWayInputProps>) {
...
             <div className="adyen-checkout__mb-way">
-                <PhoneInputForm setComponentRef={props.setComponentRef} {...props} items={phonePrefixes} onChange={onChange} data={props.data} />
+                <PhoneInputForm
+                    setComponentRef={ref => {
+                        mbWayInputRef.current = {
+                            ...mbWayInputRef.current,
+                            ...ref
+                        };
+                    }}
+                    {...props}
```
dimitra-adyen: I can still reproduce the issue reported by @m1aw. Moving the props above `setComponentRef` avoids the issue with overriding the callback
```suggestion
                    {...props}
                    setComponentRef={ref => {
                        mbWayInputRef.current = {
                            ...mbWayInputRef.current,
                            ...ref
                        };
                    }}
```

johnayeni: I had actually already fixed this but had not pushed my last changes. Done now

m1aw (approved):

dimitra-adyen (approved):

PR #4150, opened 2026-08-20
chore(useForm): type useForm hook and remove implicit anys

johnayeni:
## 📋 Pull Request Checklist

---

## 📝 Summary
Removes `src/utils/useForm/**` from the eslint `no-explicit-any`/implicit-any exemption list by giving the `useForm` hook, its reducer and the related validation types proper typings. This is an internal typing cleanup with no runtime behaviour change.

- Typed `useForm`, `getReducer`/`init` and the dispatched actions via new `FormAction`, `FormInitArg`, `ProcessFieldType` and `FormChangeEventTarget` types; `FormState` now uses `ValidationRuleResult` for `errors` and `Record<string, boolean>` for `valid`.
- Exported `ValidationResult` from `Validator.ts`, made `getError()` return `ValidationRuleResult | null` and typed `setRules(newRules: ValidatorRules)`.
- Extracted `OnChangeDataErrors` in `core/types.ts` and reused it instead of the inline error shape; deleted the duplicate `ValidationResult` interface in `PersonalDetails/types.ts` in favour of `ValidationRuleResult`.
- Adjusted call sites (ANCV, ACH, Blik, Card, PreAuthorizedDebitCanada, CtP login, PhoneInputForm) to narrow `errorMessage` before passing it to `i18n.get`.

---

## 🧪 Tested scenarios

---

**Closes**:  

---

### DIFF
--- a/packages/lib/src/components/Ach/components/AchComponent.tsx
+++ b/packages/lib/src/components/Ach/components/AchComponent.tsx
@@ -1,4 +1,4 @@
-import { h } from 'preact';
+import { h, TargetedInputEvent } from 'preact';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/hooks';
 import { useCoreContext } from '../../../core/Context/CoreProvider';
 import FormInstruction from '../../internal/FormInstruction';
@@ -92,7 +92,7 @@ function AchComponent({
         setComponentRef(achRef.current);
     }, [setComponentRef, achRef.current]);
 
-    useSRPanelForAchErrors({ errors: errors as AchStateErrors, data, isValidating });
+    useSRPanelForAchErrors({ errors: errors as unknown as AchStateErrors, data, isValidating });
 
     useEffect(() => {
         onChange({ data, valid, errors, isValid, storePaymentMethod });
@@ -103,7 +103,7 @@ function AchComponent({
      * its validation when there is any change done to the "Account number" field
      */
     const onAccountNumberInput = useCallback(
-        (event: h.JSX.TargetedInputEvent<HTMLInputElement>) => {
+        (event: TargetedInputEvent<HTMLInputElement>) => {
             handleChangeFor('accountNumber', 'input')(event);
 
             const hasAccountVerificationError = !!errors.accountNumberVerification;
@@ -124,13 +124,19 @@ function AchComponent({
                     onSelect={handleChangeFor('selectedAccountType')}
                     selectedAccountType={data.selectedAccountType}
                     disabled={isFormDisabled}
-                    errorMessage={!!errors.selectedAccountType && i18n.get(errors.selectedAccountType.errorMessage)}
+                    errorMessage={
+                        !!errors.selectedAccountType &&
+                        typeof errors.selectedAccountType.errorMessage === 'string' &&
+                        i18n.get(errors.selectedAccountType.errorMessage)
+                    }
                 />
 
                 {hasHolderName && (
                     <Field
                         label={i18n.get('ach.accountHolderNameField.title')}
-                        errorMessage={!!errors.ownerName && i18n.get(errors.ownerName.errorMessage)}
+                        errorMessage={
+                            !!errors.ownerName && typeof errors.ownerName.errorMessage === 'string' && i18n.get(errors.ownerName.errorMessage)
+                        }
                         isValid={!!valid.ownerName}
                         name={'ownerName'}
                     >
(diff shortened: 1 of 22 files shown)


packages/lib/src/utils/useForm/useForm.ts
```diff
@@ -38,25 +46,25 @@ function useForm<FormSchema>(props: FormProps): Form<FormSchema> {
...
 
-    const getTargetValue = (key, e) => {
-        if (!e.target) return e;
+    const getTargetValue = (key: string, e: unknown) => {
+        if (!hasEventTarget(e)) return e;
 
-        if (e.target.type === 'checkbox') {
-            return !state.data[key];
+        if (e.target?.type === 'checkbox') {
+            return !(state.data as Record<string, unknown>)[key];
```
m1aw: I believe this should be possible to type right instead of using Record here?

packages/lib/src/utils/useForm/useForm.ts
```diff
@@ -19,16 +27,16 @@ function useForm<FormSchema>(props: FormProps): Form<FormSchema> {
     } = props;
     const rules = providedRules ?? {};
     const formatters = providedFormatters ?? {};
-    const defaultData = providedDefaultData ?? {};
-    const fieldProblems = providedFieldProblems ?? {};
+    const defaultData = (providedDefaultData ?? {}) as Record<string, unknown>;
```
m1aw: What did we decide to go with `unkown` here?

m1aw (commented):
Already discussed this with @johnayeni, I believe we should try an approach were we use the generic `FormSchema`.

packages/lib/src/utils/useForm/useForm.ts
johnayeni: Fixed

packages/lib/src/utils/useForm/useForm.ts
johnayeni: Fixed

johnayeni:
> Already discussed this with @johnayeni, I believe we should try an approach were we use the generic `FormSchema`.

the approach has been adjusted for `FormSchema` to be generic 👍🏿

packages/lib/src/components/Ach/components/AchComponent.tsx
```diff
@@ -124,13 +124,13 @@ function AchComponent({
                     onSelect={handleChangeFor('selectedAccountType')}
                     selectedAccountType={data.selectedAccountType}
                     disabled={isFormDisabled}
-                    errorMessage={!!errors.selectedAccountType && i18n.get(errors.selectedAccountType.errorMessage)}
+                    errorMessage={typeof errors.selectedAccountType?.errorMessage === 'string' && i18n.get(errors.selectedAccountType.errorMessage)}
```
m1aw, line 127: I wonder if here we should move this check to a wrapper function. I don't quite like that we are doing quite a bit, I understand is for type-safety but I wonder we can improve this.

packages/lib/src/components/PayTo/components/PayToNameFields.tsx
```diff
@@ -28,7 +25,7 @@ export default function PayToNameFields({ i18n, errors, data, handleChangeFor, p
             <Field
                 label={i18n.get('payto.label.firstName')}
                 classNameModifiers={['col-50', 'firstName']}
-                errorMessage={getErrorMessage(i18n, errors.firstName, i18n.get('firstName'))}
+                errorMessage={getErrorMessage(i18n, errors.firstName ?? undefined, i18n.get('firstName'))}
```
m1aw, line 28: I guess we could move this logic to inside of `getErrorMessage`.

packages/lib/src/utils/useForm/reducer.ts
```diff
@@ -53,36 +54,39 @@ export function init({ schema, defaultData, processField, fieldProblems }) {
...
                             fieldProblems: { ...state['fieldProblems'], [key]: fieldProblems?.[key] ?? null },
-                            valid: { ...state['valid'], [key]: state['valid']?.[key] && !fieldProblems[key] }
+                            valid: { ...state['valid'], [key]: state['valid']?.[key] && !fieldProblems?.[key] }
                         }),
                         state
                     ) ?? state
                 );
             }
             case 'updateField': {
+                const { key, value, mode } = action;
```
m1aw, line 89: I like this change makes it way more clear, nicely spotted.

m1aw (dismissed):
Nice PR, I would threat the error -> getMessage stuff as a nitpick so feel free to just merge if you want.

packages/lib/src/components/Ach/components/AchComponent.tsx
johnayeni: This is a good idea. Maybe for now I could keep it this way to prevent this PR from becoming bigger since some of checks have some small differences.

packages/lib/src/components/PayTo/components/PayToNameFields.tsx
johnayeni: Maybe also good for a follow up

m1aw (dismissed):

m1aw (approved):

PR #4179, opened 2026-09-10
Feature/disclaimer message component

dimitra-adyen:
## 📋 Pull Request Checklist

---

## 📝 Summary

- Adds configurable disclaimer messages above payment buttons.
- Centralizes `showPayButton` handling in `PayButton`, keeping disclaimers visible when merchants provide their own button.
- Supports plain text and link placeholders and updates the disclaimer link color to match the design.

---

## 🧪 Tested scenarios

- Disclaimer rendering with text and links.
- Disclaimer visibility when the SDK pay button is hidden.
- Existing payment and review button behavior.

---

**Closes**:

- COSDK-1286

---


packages/lib/src/styles/mixins.scss
```diff
@@ -99,7 +99,7 @@ $adyen-checkout-media-query-xl-min: 1280px;
 }
 
 @mixin adyen-checkout-link {
-    --text-decoration: none;
+    --text-decoration: underline;
```
ScottiBR: Thinking about the size of this PR, maybe this should be a separated PR just refactoring link style, together with `color-label-highlight: #00112C`?

packages/lib/src/components/internal/PayButton/PayButton.test.tsx
```diff
@@ -46,6 +46,30 @@ describe('PayButton', () => {
...
+
+        const button = screen.getByRole('button');
+        expect(termsLink.compareDocumentPosition(button) & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    });
+
+    test('should not render a disclaimer message when none is configured', () => {
+        renderPayButton();
+        expect(screen.queryByRole('link')).toBeNull();
+    });
+
```
ScottiBR, line 75: Should we add something like `should not render the button when showPayButton is false` and `should still render the disclaimer message when showPayButton is false`?

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.tsx
```diff
@@ -6,11 +6,11 @@ import Link from '../Link';
 
 export interface DisclaimerMsgObject {
     message: string;
-    linkText: string;
-    link: string;
+    linkText: string | Array<string>;
+    link: string | Array<string>;
```
ScottiBR: It's possible for a merchant to pass a disclaimer message without a link like `Payments are processed securely.`? it's flexible enough?

packages/lib/src/components/internal/RedirectButton/RedirectButton.tsx
```diff
@@ -11,11 +11,10 @@ export interface RedirectButtonProps {
     payButton: Function;
     onSubmit: Function;
     name: string;
-    showPayButton: boolean;
     ref?: Ref<typeof RedirectButton>;
 }
 
-function RedirectButton({ label = null, icon = null, payButton, onSubmit, name, showPayButton, ...props }: Readonly<RedirectButtonProps>) {
+function RedirectButton({ label = null, icon = null, payButton, onSubmit, name, ...props }: Readonly<RedirectButtonProps>) {
```
ScottiBR, line 17: Maybe we should explicit pass showPayButton prop instead of just trust the `...props` spread

packages/lib/src/components/internal/DisclaimerMessage/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Converts the merchant facing disclaimer message config into the props of the DisclaimerMessage component.
+ *
+ * Each `%{placeholder}` in the message is replaced, in order of appearance, by the corresponding entry of
+ * `linkText` wrapped in the `%#` token pair that DisclaimerMessage renders as an anchor element. Slots left
+ * without a linkText or a link are kept as they are, so the message degrades to plain text.
+ *
+ * @param disclaimerMessage - merchant provided disclaimer message
+ * @returns the message with interpolation tokens and the urls matching the order of the tokens
+ */
+export function formatDisclaimerMessage({ message, linkText, link }: DisclaimerMsgObject): DisclaimerMessageProps {
```
ScottiBR, line 20: What happens if a merchant writes `message: 'accept the %{terms}'` and forgets `linkText`/`link`? `toArray(undefined)` gives `[undefined]`, this guard returns the slot, and the shopper ends up seeing a literal `%{terms}` on the checkout page. 

Would you rather leak %{terms} to the shopper or strip the slot?

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.tsx
```diff
@@ -21,17 +21,20 @@ interface InternalDisclaimerMsgObject {
...
  *  String inside the '%#' token pair will be rendered as an anchor element.
+ *
+ *  Merchant configured disclaimers use the `%{placeholder}` format instead, and must be passed
+ *  through `formatDisclaimerMessage` to obtain these props.
  */
 
-export default function DisclaimerMessage({ message, urls = [] }: Readonly<InternalDisclaimerMsgObject>) {
+export default function DisclaimerMessage({ message, urls = [] }: Readonly<DisclaimerMessageProps>) {
     return (
         <span className="adyen-checkout-disclaimer__label">
```
ScottiBR: should we have an `aria-describedby`?

ScottiBR (changes requested):
Left a few comments

packages/lib/src/components/internal/PayButton/PayButton.test.tsx
dimitra-adyen: added a test to verify that the button stays hidden, but the disclaimer gets displayed in case of showPayButton false.

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.tsx
dimitra-adyen: made both `linkText` and `link` optional and added tests

packages/lib/src/components/internal/DisclaimerMessage/utils.ts
dimitra-adyen: I updated the solution to not display unresolved placeholders. If only `link` is missing, the `linkText` will get rendered. If `linkText` is missing, the placeholder gets removed.

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.tsx
dimitra-adyen: good catch! I added a unique id to the disclaimer, and an aria-describedby to the associated paybutton

packages/lib/src/styles/mixins.scss
dimitra-adyen: I see your point. I opened a new PR that includes that change only: https://github.com/Adyen/adyen-web/pull/4185

packages/lib/src/components/internal/RedirectButton/RedirectButton.tsx
dimitra-adyen: I was doubting if I should keep it or not. I removed it, since it already gets forwarded through `...props`. I thought that if I explicitly pass it, it would basically duplicate the existing behavior. I don't have a strong opinion though, so if you think that making it explicit makes it more clear I can make the change, both here and in the components that call the RedirectButton.

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.tsx
```diff
@@ -21,17 +22,20 @@ interface InternalDisclaimerMsgObject {
...
+        <span id={id} className="adyen-checkout-disclaimer__label">
             <LabelOnlyDisclaimerMessage message={message} urls={urls} />
         </span>
     );
 }
 
-export function LabelOnlyDisclaimerMessage({ message, urls }: Readonly<InternalDisclaimerMsgObject>) {
+export function LabelOnlyDisclaimerMessage({ message, urls = [] }: Readonly<DisclaimerMessageProps>) {
     const messageIsStr = typeof message === 'string';
     const validUrls = urls.every(url => typeof url === 'string' && isValidHttpUrl(url));
```
ScottiBR, line 40: Do we need to worry about a merchant passing an `http` link?  Because `isValidHttpUrl` defaults to `allowHttp = false`

packages/lib/src/components/internal/UIElement/types.ts
```diff
@@ -76,6 +77,11 @@ export type UIElementProps = {
          */
         showPayButton?: boolean;
 
+        /**
+         * Disclaimer message displayed above the pay button
+         */
+        disclaimerMessage?: DisclaimerMsgObject;
```
ScottiBR, line 86: I was looking Paypal, ApplePay, Amazon and GooglePay all handle `showPayButton` themselves and never go through `PayButton`. Are we going to support wallets or call that out in the doc like this:

```
        /**
         * Disclaimer message displayed above the pay button.
         *
         * Not supported by express/wallet components (PayPal, Apple Pay, Google Pay, Amazon Pay, Cash App Pay),
         * which render their own branded button instead of the SDK pay button.
         */
        disclaimerMessage?: DisclaimerMsgObject;
```

ScottiBR (changes requested):
Just a small detail

packages/lib/src/components/internal/DisclaimerMessage/DisclaimerMessage.tsx
dimitra-adyen: I believe that we should keep this behavior, allowing https only, since we expect these links to usually contain legal consent info, that could otherwise be altered or redirect shoppers to a different page.

packages/lib/src/components/internal/UIElement/types.ts
dimitra-adyen: It's better if we keep them excluded since as you mentioned they render their own PayButton. I updated the documentation, to reflect that.

packages/lib/src/components/Dropin/stories/dropin/Dropin.stories.tsx
```diff
@@ -40,6 +40,13 @@ const meta: MetaConfiguration<DropinConfiguration> = {
                     buttonType: 'plain',
                     challengeWindowSize: '05'
                 }
+                // card: {
+                //     disclaimerMessage: {
+                //         message: 'By continuing you agree with the %{terms}',
+                //         linkText: 'terms and conditions',
+                //         link: 'https://www.adyen.com'
+                //     }
+                // },
```
ScottiBR, line 49: Should we remove the comments or add a new stories for that? maybe at component level would be better

ScottiBR (dismissed):
Well done, just a small comment but feel free to ignore

packages/lib/src/components/Dropin/stories/dropin/Dropin.stories.tsx
dimitra-adyen: Removed the comment and added a story under the card component

ScottiBR (dismissed):
LGTM!

ScottiBR (approved):

